PR #4, opened 2015-11-13
database: reduce InsertPackages transaction.

Quentin-M:
Could possibly create deadlock transactions on PostgreSQL.


database/package.go
```diff
@@ -115,6 +113,10 @@ func InsertPackages(packageParameters []*Package) error {
 		return nil
 	}
 
+	// Sort the packages that will be inserted to ensure that locks will be acquired in the same
+	// a deterministic order.
+	sort.Sort(ByVersion(packageParameters))
```
jakedt: This needs to sort by os and package name to ensure the locks are taken in order.

jakedt:
LGTM

PR #6, opened 2015-11-15
database: reduce pruneLocks/Unlock transaction.

Quentin-M:
pruneLocks could create deadlocked transactions on PostgreSQL if multiple locks expired and pruneLocks is called by multiple instances because Cayley does not guarantee the order.

[InsertNotifications](https://github.com/coreos/clair/blob/8bb6c50f802257228161f9f2d99edaee5b59d18e/database/notification.go#L312) and [InsertVulnerabilities](https://github.com/coreos/clair/blob/8bb6c50f802257228161f9f2d99edaee5b59d18e/database/vulnerability.go#L100) also insert multiple independent items in a single same transaction. Doing so in InsertNotifications helps because as the transaction only contains new quads, COPY can be used. For the same reason, it could not create a deadlocked transaction. In the other hand, InsertVulnerabilities may delete some fields. However, just like InsertNotifications, it is called from the updater - which run only once per cluster: it could not create a deadlocked transaction. Additionally, the updater relies on the fact that these two functions will fail and rollback entirely if any of the elements can't be inserted (for any reason) - in which case it will not mark the update as done and will redo it later. I believe that because of the reasons above, they should not be modified.

### DIFF
--- a/database/lock.go
+++ b/database/lock.go
@@ -68,20 +68,32 @@ func Lock(name string, duration time.Duration, owner string) (bool, time.Time) {
 func Unlock(name, owner string) {
 	pruneLocks()
 
-	t := cayley.NewTransaction()
+	unlocked := 0
 	it, _ := cayley.StartPath(store, name).Has("locked", "locked").Has("locked_by", owner).Save("locked_until", "locked_until").BuildIterator().Optimize()
 	defer it.Close()
-
 	for cayley.RawNext(it) {
 		tags := make(map[string]graph.Value)
 		it.TagResults(tags)
 
+		t := cayley.NewTransaction()
 		t.RemoveQuad(cayley.Quad(name, "locked", "locked", ""))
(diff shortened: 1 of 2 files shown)


database/lock.go
```diff
@@ -68,7 +68,6 @@ func Lock(name string, duration time.Duration, owner string) (bool, time.Time) {
 func Unlock(name, owner string) {
 	pruneLocks()
 
-	t := cayley.NewTransaction()
 	it, _ := cayley.StartPath(store, name).Has("locked", "locked").Has("locked_by", owner).Save("locked_until", "locked_until").BuildIterator().Optimize()
```
jakedt: If the assumption is that this will only return one thing from the iterator, please verify it and log an error if your assumption is violated. The change to the transaction scope should be fine.

jakedt:
LGTM

PR #21, opened 2015-11-20
stop reporting failures as successful updates

### LINKED ISSUE #20: Updater always set the latest successful date
Because Update does not return any status, the latest successful update date is always updated even if it failed.

Also, we want to make sure that we don't update the flag of a failed fetcher. Today, fetchers only returns a flag if they succeeded but eventually contributors may start writing fetchers and don't respect that.

jzelinskie:
Fixes #20.

### DIFF
--- a/updater/updater.go
+++ b/updater/updater.go
@@ -21,10 +21,10 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/coreos/pkg/capnslog"
 	"github.com/coreos/clair/database"
 	"github.com/coreos/clair/health"
 	"github.com/coreos/clair/utils"
+	"github.com/coreos/pkg/capnslog"
 	"github.com/pborman/uuid"
 )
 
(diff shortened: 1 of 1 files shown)


updater/updater.go
```diff
@@ -93,23 +93,16 @@ func Run(interval time.Duration, st *utils.Stopper) {
...
 						database.Lock(flagName, lockDuration, whoAmI)
 					}
 				}
 
-				// Write the last update time to the database and set the next update
-				// time.
-				now := time.Now().UTC()
-				database.UpdateFlag(flagName, strconv.FormatInt(now.Unix(), 10))
-				healthLatestSuccessfulUpdate = now
-				nextUpdate = now.Add(interval)
```
Quentin-M, line 111: If we remove that line, an updater that failed will just wait some random seconds and retry. I guess that's ok?

jzelinskie, line 111: Yep

Quentin-M:
LGTM

PR #25, opened 2015-11-23
Don't pass -z to tar

fatalbanana:
AFAICT `docker save` will produce an uncompressed tar archive.

This prevented successful run of `analyze-local-images`:

```
2015/11/23 13:59:30 - Could not save image: 
gzip: stdin: not in gzip format
tar: Child died with signal 13
tar: Error is not recoverable: exiting now
```

The patch fixed it. :)

If `docker save` will in fact produce compressed output?- & one's implementation of `tar` is not capable of self-identifying compression, then perhaps this would break things. :\

### DIFF
--- a/contrib/analyze-local-images/main.go
+++ b/contrib/analyze-local-images/main.go
@@ -126,7 +126,7 @@ func save(imageName string) (string, error) {
 	var stderr bytes.Buffer
 	save := exec.Command("docker", "save", imageName)
 	save.Stderr = &stderr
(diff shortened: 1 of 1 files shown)


Quentin-M:
You are right ! Good catch, my `tar` implementation is capable of skipping `-z` the archive is not actually gzip'd.

Can you just edit your commit message to make it more standard ? Something like `contrib: Don't pass -z to tar in analyze-local-images` ?

Thanks for your contribution.

PR #26, opened 2015-11-23
Fix installation command in README.md

stapelberg:
The package name was missing an “s”, and “go install” only works when you already have downloaded the package.


jonboulle:
LGTM, thanks.

PR #36, opened 2015-11-30
api/database: Add the ability to delete layers

Quentin-M:


database/layer.go
```diff
@@ -131,6 +131,59 @@ func InsertLayer(layer *Layer) error {
 	return nil
 }
 
+// DeleteLayer delete the specified layer and every layers that are based on that one, recursively.
```
jzelinskie: DeleteLayer deletes the specified layer and any child layers that are dependent on the specified layer.

Also, comments in Go are 80 chars per line

Quentin-M: Dang it. Such a pain in the butt to switch back and forth between 100C (Python) to 80C (Go).

jzelinskie: Use [go-lint](https://github.com/golang/lint)

Quentin-M: I do use it through https://atom.io/packages/go-plus, automatically on save but it does not wrap comments / enforce line length afaik.

jzelinskie:
LGTM

Issue #19, opened 2015-11-20
Generic vulnerability link & more metadata


Quentin-M:
Because otherwise it could link to the Debian vuln on an Ubuntu container and may be confusing.

We could also get more informations about CVEs (Access Vector, Access Complexity, Authentication ...)  with additional databases such as NVD.

Quentin-M:
#54 will fix part of it as the generalization will make Clair able to specify links, description and severity per namespace (operating system).

Quentin-M:
Merged (#71).

PR #37, opened 2015-12-01
updater: Refactor and merge fetcher responses

### LINKED ISSUE #17: Make the healthcheckers more precise
Concerns especially the updater's healthcheck in which health data is cached and can stay unchanged for a while. It would be better to actively retrieve these.

Quentin-M:
Fixes #17 and lays the groundwork for #19.

### DIFF
--- a/updater/fetchers.go
+++ b/updater/fetchers.go
@@ -14,10 +14,7 @@
 
 package updater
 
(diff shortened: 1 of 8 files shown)


jzelinskie:
LGTM

PR #38, opened 2015-12-01
api: Specify what packages cause the layer to have vulnerabilities.

Quentin-M:
Following up on a feedback !

@jzelinskie @josephschorr


barakmich:
[LGTM](http://lgtm.in/i/qjdMkyEnq)

PR #46, opened 2015-12-10
database: Fix incorrect behavior of toValue when reading empty strings.

Quentin-M:
If there are two strings for a path (ie. "", "yes"), always return "yes"
and never "".

Unlikely to happen but, still.


jzelinskie:
I'm not a fan of defensive programming for "what if" scenarios. If something is introducing an empty path, that's a bug, not this code. Is there ever a scenario when the path could be empty and it isn't a bug?

Quentin-M:
Updated. It basically makes toValues() and toValue() stop ignoring empty strings.

jzelinskie:
LGTM

PR #47, opened 2015-12-10
notifier: add ServerName configuration for TLS

jzelinskie:


Quentin-M:
You changed the generic utility function `loadTLSClientConfig` to make it specific to the HTTP `Notifier`. True it is not used by anything else at the moment, but it may be. Also, almost none of the functions in the utility package are re-used at that point (ie. `LoadTLSClientConfigForServer`, the sibling function is just used in the API as well) - but they are pretty generic imho and could be re-used at some point.

jzelinskie:
> True it is not used by anything else at the moment, but it may be.

This is a good reason why we should wait and see if it can be reused before refactoring it out. If it can, refactor it out; if it can't, keep it local to the thing using it. The reason I made it localized was because if we continue adding new options for the TLS config, the parameters to the function could get huge. To handle this problem you can do a number of things:
- Make the function specific to *config.NotifierConfig
- Make the function specific to a new *config.TLSConfig
- Make the function take a *tls.Config instead of allocating our own inside the function

I chose the least-invasive option.

Quentin-M:
Sounds good to me.
Should we do the same thing with `LoadTLSClientConfigForServer` then?

Quentin-M:
LGTM

Issue #48, opened 2015-12-11
Add support to ACI format


liangchenye:
Hi, 
Is there any plan to support ACI format?
https://github.com/coreos/rkt/blob/master/Documentation/app-container.md#ACI
(If we have such mechanize to support ACI, we could support OCI bundle easily in the future.)

jzelinskie:
@liangchenye We are interested in this behavior. Let's leave the issue open until we've merged code that supports scanning ACI.

liangchenye:
checked in, close.

Issue #8, opened 2015-11-16
Add support not only tar layouts


mrqwer88:
Hello!

How I see clair can add layouts only in tar archives now.
Are planned add layouts like directory and tar.gz?
As example with tar.gz support it be more comfortably to check OpenVZ templates - https://download.openvz.org/template/precreated/ ( if ungzip it - clair work correctly with this templates ).
And if clair can work with directory - it can check already created openvz containers and etc.

Quentin-M:
Hello,

Clair is actually able to accept .tar.gz archives already! However, it does not accept directories yet but it would really be a quick feature to add. Using Clair against OpenVZ containers is a pretty good idea. OpenVZ templates can be seen as one layer images. I am myself an user of OpenVZ and I never had that thought before.

However, be aware that Clair has initially been designed to analyze huge amount of images/templates, (which are immutable) and because of that, you can't just re-submit an image with the same ID and expect Clair to analyze it again, it will skip the analysis saying that it already knows about that image. Clair will only analyze images again if it's engine version is increased (meaning that Clair could know learn more about images). It makes it kinda unsuitable for analyzing mutable systems like running containers.

mrqwer88:
Thanks for answers.

Yes, with tar.gz clair work correctly  too - my bad.

Re-submit for containers not a problem - in any case, the addition would be to go through the script.
If be added support for add layouts via directory.

mrqwer88:
And little wrapper for add templates from openvz 6 mirror to clair - [check_openvz_mirror_with_clair](https://github.com/FastVPSEestiOu/check_openvz_mirror_with_clair)

Quentin-M:
That's really cool ! That could totally go to the `contrib/` folder under that repository. Make a pull request when you think your tool is ready/mature enough. I will be glad to merge it.

PR #49, opened 2015-12-11
Add DataDetector to support ACI and other layout format

liangchenye:
Now we have DetectOS and DetectPackage, I think it also makes sense to have a 'DetectData' to handle different image formats. The PR implements this feature.  It adds 'imageFormat' to the parameters for DataDetector to choose the right detector. For most use cases, user knows what format a image is. If not, the detector checks the suffix at present.

It solves my issue (https://github.com/coreos/clair/issues/48) and (https://github.com/coreos/clair/issues/8)
Other potential benefits:
1.  folder  (add a folder.go to detectors/data)
2. OCI   (add a oci.go to detectors/data)


Quentin-M:
The [aci spec](https://github.com/appc/spec/blob/master/spec/aci.md) specifies:

```
Image archives MAY be compressed with gzip, bzip2, or xz.
```

Only `tar/gzip` is supported at the moment.
Also, there is the encryption thing but I am unsure that it is necessary for Clair to handle it.

```
Image archives MAY be encrypted using PGP symmetric encryption with AES cipher, after optional compression.
```

liangchenye:
Thanks for collecting './'.

I hope 'Clair' could handle this feature, or the user have to untar the ACI and tar the rootfs to make it work.
Or is it suggested  to implement such feature in 'contrib' ?

Quentin-M:
It is definitely not difficult to do from what you've already done.
- You have to modify your PR to always trim `./` and then an optional prefix (`rootfs` in the case of ACI).
- If you want Clair to be able to handle bzip2/xz compressed images, modify `SelectivelyExtractArchive` to support it. However, that is not strictly **required** as stated by the spec. Same thing with PGP.

jzelinskie:
linking #48

liangchenye:
@Quentin-M @jzelinskie bzip2/xz are done.
Since there is no golang compress/xz, the current xz implementation is copied from github.com/appc/spec/aci/file.go. 
Add utils_test.tar.bz2 and utils_test.tar.xz at the same time.

The PGP support is not implemented yet, I don't have a concrete idea. It requires an extra signature file. Maybe we need to change the API?

Quentin-M:
Looks really cool ! I will take a look as soon as I get to the office !

worker/detectors/data/tar.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"io"
+	"strings"
+
+	"github.com/coreos/clair/utils"
+	"github.com/coreos/clair/worker/detectors"
+)
+
+// TarDataDetector implements DataDetector and detects layer data in 'tar' format
+type TarDataDetector struct{}
```
Quentin-M: It should probably not be called `TarDataDetector`. Both this/ACI are tar anyways. Should probably be `DockerDataDetector` instead.

worker/detectors/data/aci.go
```diff
@@ -0,0 +1,46 @@
...
+			return true
+		}
+	case "aci":
+		return true
+	}
+	return false
+}
+
+func (detector *ACIDataDetector) Detect(layerReader io.ReadCloser, toExtract []string, maxFileSize int64) (map[string][]byte, error) {
+	return utils.SelectivelyExtractArchive(layerReader, "./rootfs/", toExtract, maxFileSize)
```
Quentin-M: It looks for `./rootfs/` instead of `rootfs/`. Once it is fixed, ACI works!

worker/worker.go
```diff
@@ -64,7 +61,7 @@ func Process(ID, parentID, path string) error {
 		return cerrors.NewBadRequestError("could not process a layer which does not have a path")
 	}
 
-	log.Debugf("layer %s: processing (Location: %s, Engine version: %d, Parent: %s)", ID, utils.CleanURL(path), Version, parentID)
+	log.Debugf("layer %s: processing (Location: %s, Engine version: %d, Parent: %s, format: %s)", ID, utils.CleanURL(path), Version, parentID, imageFormat)
```
Quentin-M: Use `Format:` instead (uppercase).

utils/tar.go
```diff
@@ -32,19 +34,75 @@ var (
...
+	closech chan error
+}
+
+// NewXzReader shells out to a command line xz executable (if
+// available) to decompress the given io.Reader using the xz
+// compression format and returns an *XzReader.
+// It is the caller's responsibility to call Close on the XzReader when done.
+func NewXzReader(r io.Reader) (*XzReader, error) {
+	rpipe, wpipe := io.Pipe()
+	ex, err := exec.LookPath("xz")
```
Quentin-M, line 57: `xz` should be added as a dependency in the [Dockerfile](https://github.com/coreos/clair/blob/master/Dockerfile#L4)

api/logic/layers.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // POSTLayersParameters represents the expected parameters for POSTLayers.
 type POSTLayersParameters struct {
-	ID, Path, ParentID string
+	ID, Path, ParentID, ImageFormat string
```
Quentin-M, line 33: API doc about [layer insertion](https://github.com/coreos/clair/blob/master/docs/API.md#insert-a-new-layer) should be updated as a new parameter appeared

Quentin-M:
That looks cool. we are really looking forward to having ACI support !

I quickly tested it and did some comments.
As far as I can tell:
- ACI support works after a small fix (see comments)
- bzip2 / xz support work as well

Also, I discussed a bit with the team and we think that we should not try to do format detection automatically - it could become tricky at the end of the day if we support more than Docker/ACI. In other words, `ImageFormat` should be a mandatory parameter, with currently two choices: `Docker` and `ACI`. Not giving the format should fail (BadRequest) - asking the user to specify it. 

Thanks!

Quentin-M:
@alban Adding to the discussion about PGP. See https://github.com/coreos/clair/pull/49#issuecomment-164488516.

liangchenye:
@Quentin-M  Thanks for the review, the new commits solve problems you mentioned above.

One thing is missing: a ACI image for testing. I'm not sure if I should make a fake one based on the rootfs in utils_test.tar or use the official 'etcd' one in https://github.com/coreos/rkt.

Quentin-M:
@liangchenye Thanks for the fixes and your contribution! I am really excited by seeing you contributing to Clair. I will take a deeper look at it tomorrow (Almost 1 AM here) and ensure Clair still works fine for the general usage (well, I bet it does), but I think that's pretty great! @jzelinskie has the last words on the code-reviewing procedure. He may have some comments too. After that and a commit squashing with an [adequate](https://github.com/coreos/clair/blob/master/CONTRIBUTING.md#format-of-the-commit-message) commit message, I'll be happy to merge your work!

About testing, I think that, generally speaking, we should not include big files in the project as it slows down repository pulling and increase the size of the project. Instead, we should use minimal data samples - that are relevant to the tests we want to run. As you can notice in `worker/packages` for instance, we have an excerpt from a `/var/lib/dpkg/status` file (that I modified slightly to cover more test surface) and a shrunken (minimized) rpm database from `/var/lib/rpm/Packages`. Depending on what you want to do, couple files would be enough.

liangchenye:
my pleasure :)

PR #50, opened 2015-12-14
fix type ctrl^C not stop updating

coolljt0725:
when clair has network problem during updating vulnerability and failed
to update vulnerability, it will keep updating and even if type ctrl^C
can't stop the clair. 

<pre><code>2015-12-14 06:01:30.632114 E | updater: an error occured when fetching update 'Ubuntu': could not download requested ressource.
^C2015-12-14 06:01:42.915105 I | main: Received interruption, gracefully stopping ...
2015-12-14 06:01:42.915259 I | api: main API stopped
2015-12-14 06:01:42.915286 I | api: health API stopped
</code></pre>

As you can see I type ^C and expect the clair will stop,`main API` and `health API` do stop
but `updater` doesn't, it keeps running.
This patch make clair to stop updating if type
ctrl^C.

### DIFF
--- a/updater/updater.go
+++ b/updater/updater.go
@@ -61,6 +61,7 @@ func Run(config *config.UpdaterConfig, st *utils.Stopper) {
 		// is no last update time stored in database (first update) or if an error
 		// occurs.
 		var nextUpdate time.Time
+		var stop bool
 		if lastUpdate := getLastUpdate(); !lastUpdate.IsZero() {
 			nextUpdate = lastUpdate.Add(config.Interval)
 		} else {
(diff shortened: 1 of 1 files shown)


Quentin-M:
Thanks,

What happens is that when the interruption is sent, `bzr` is killed right away and it makes the fetcher fails right away. However, an update doesn't fail if a fetcher does - so the update just continues as planned and Clair does not stop until it's done.

Your commit will definitely fix, by interrupting the update completion wait. My only concern is that if Clair stops between L153 and L161, some notifications could be lost. We can either accept that & merge your PR or make the updater fail completely if any of the fetchers fail. Any thoughts ?

coolljt0725:
IMO, I don't think `make the updater fail completely if any of the fetchers fail` is a good idea, because sometimes the networking is not stable especially for those who use vpn (like me, thanks the GFW), updating fails this time but may success next time.

Quentin-M:
Exact. That's the reason that motivated my decision initially, and it should probably stay like this in my opinion.

Thanks for your contribution.

PR #51, opened 2015-12-15
Update README.md of analyze_local_image

coolljt0725:
The latest Clair use config file for configuration, the command line to
run Clair instance is outdate in the README.md


Quentin-M:
Nice catch. Thanks!

PR #53, opened 2015-12-16
updater: Fix Ubuntu's partial update bug.

Quentin-M:
Deferring file closing causes `too many open files` on some platforms!


jzelinskie:
Just put a comment above both closes, saying how you have to manually close and not use `defer` due to it operating on function-level scope and LGTM

PR #71, opened 2016-02-14
Clair 1.0 Mega PR

Quentin-M:
This is still a work in progress. Any tester are very welcomed.

PR #74, opened 2016-02-18
Add output for package causing vulnerability

mnuessler:
Include the name of package that caused the vulnerability in the
output.


Quentin-M:
Thanks.

The good news is that I'm going to spend some time today to make the output of this tool better for the new major Clair version (https://github.com/coreos/clair/pull/71). It's going to include the package name and version that introduces the vulnerability but also the version that would fix it and CVSS information. We could also add the layer that introduces the package.

mnuessler:
Sounds great, looking forward to it. :)

Issue #80, opened 2016-02-26
add a `ListVulnerabilities` API


liangchenye:
When I use Clair, sometimes I want to know whether my images are really secure or just because my CVE database is empty, so I try to read my CVE database. I found that Clair does not have such API yet. There is a similar API (list namespace), I think maybe we can add this.  I already work on it in my repo (liangchenye/clair:getvulns), now the status is 'in progress'. 

@Quentin-M  @jzelinskie what do you think about this API?

Quentin-M:
Hi,

That makes sense to me!
I looked a bit at your branch, I think that this API call should be paginated (see the notification code).

jzelinskie:
@liangchenye I can see the use for `ListVulnerabilities` but for the scenario you described, would a better solution be to have an "last updated" field on the namespaces to see the last time a fetcher ran for it?

liangchenye:
@jzelinskie yes, and that is another API:  get the metainfo for a given namespace.

Something like this:
GET http://localhost:6060/v1/namespaces/debian%3A8

{ "last_updated": ...
  "vulns_count": 1000
}

Quentin-M:
#82 Merged. I'm closing this issue :)

PR #82, opened 2016-02-27
New API: list vulnerabilities by namespace

liangchenye:
https://github.com/coreos/clair/issues/80

Signed-off-by: liangchenye liangchenye@huawei.com

### DIFF
--- a/api/v1/README.md
+++ b/api/v1/README.md
@@ -8,6 +8,7 @@
 - [Namespaces](#namespaces)
   - [GET](#get-namespaces)
 - [Vulnerabilities](#vulnerabilities)
+  - [List](#get-namespacesnsnamevulnerabilities)
   - [POST](#post-namespacesnamevulnerabilities)
   - [GET](#get-namespacesnsnamevulnerabilitiesvulnname)
   - [PUT](#put-namespacesnsnamevulnerabilitiesvulnname)
@@ -195,6 +196,60 @@ Server: clair
 ```
 
 ## Vulnerabilities
+
+#### GET /namespaces/`:nsName`/vulnerabilities
+
+###### Description
+
+The GET route for the Vulnerabilities resource displays the vulnerabilities data for a given namespace.
+
+###### Query Parameters
+
+| Name    | Type | Required | Description                                                |
+|---------|------|----------|------------------------------------------------------------|
+| limit   | int  | required | Limits the amount of the vunlerabilities data for a given namespace. |
+| page    | int  | required | Displays the specific page of the vunlerabilities data for a given namespace. |
+
+###### Example Request
+
+```json
+GET http://localhost:6060/v1/namespaces/debian%3A8/vulnerabilities?page=0&limit=2 HTTP/1.1
+```
+
+###### Example Response
+
+```json
+HTTP/1.1 200 OK
+Content-Type: application/json;charset=utf-8
+Server: clair
+
+{
+    "Vulnerabilities": [
+        {
+            "Name": "CVE-1999-1332",
+            "Namespace": "debian:8",
+            "Description": "gzexe in the gzip package on Red Hat Linux 5.0 and earlier allows local users to overwrite files of other users via a symlink attack on a temporary file.",
+            "Link": "https://security-tracker.debian.org/tracker/CVE-1999-1332",
+            "Severity": "Low"
+        },
+        {
+            "Name": "CVE-1999-1572",
+            "Namespace": "debian:8",
+            "Description": "cpio on FreeBSD 2.1.0, Debian GNU/Linux 3.0, and possibly other operating systems, uses a 0 umask when creating files using the -O (archive) or -F options, which creates the files with mode 0666 and allows local users to read or overwrite those files.",
+            "Link": "https://security-tracker.debian.org/tracker/CVE-1999-1572",
+            "Severity": "Low",
+            "Metadata": {
+                "NVD": {
+                    "CVSSv2": {
+                        "Score": 2.1,
+                        "Vectors": "AV:L/AC:L/Au:N/C:P/I:N"
+                    }
+                }
+            }
+        }
+    ]
+}
+```
 
 #### POST /namespaces/`:name`/vulnerabilities
 
--- a/api/v1/README.md
+++ b/api/v1/README.md
@@ -213,7 +213,7 @@ The GET route for the Vulnerabilities resource displays the vulnerabilities data
 ###### Example Request
 
 ```json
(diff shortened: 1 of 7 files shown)


database/pgsql/queries.go
```diff
@@ -141,6 +141,9 @@ const (
 	searchVulnerabilityForUpdate          = ` FOR UPDATE OF v`
 	searchVulnerabilityByNamespaceAndName = ` WHERE n.name = $1 AND v.name = $2 AND v.deleted_at IS NULL`
 	searchVulnerabilityByID               = ` WHERE v.id = $1`
+	searchVulnerabilityByNamespace        = ` WHERE n.name = $1 AND v.deleted_at IS NULL
+						  ORDER BY v.name
+						  LIMIT $2 offset $3`
```
Quentin-M: To ensure good performances at scale, we try to avoid using `OFFSET` but instead use `WHERE id >= $3 ORDER BY l.ID LIMIT $2`. And then to avoid leaking that internal identifier, we encrypt the id at the API level. See how notifications work.

database/pgsql/vulnerability.go
```diff
@@ -28,6 +28,50 @@ import (
 	"github.com/guregu/null/zero"
 )
 
+func (pgSQL *pgSQL) ListVulnerabilities(namespaceName string, limit int, page int) ([]database.Vulnerability, error) {
+	return listVulnerabilities(pgSQL, namespaceName, limit, page)
+}
+
+func listVulnerabilities(queryer Queryer, namespaceName string, limit int, page int) ([]database.Vulnerability, error) {
```
Quentin-M: This can be a single function. This pattern with the `Queryer` parameter is used in `scanVulnerability` because one of the caller needs to use a transaction, while another caller doesn't.

Quentin-M:
Thanks! I just wrote some comments ~

liangchenye:
Update by encrypting page, just like notificationPage.
I added numberToToken/tokenToNumber, and NextPage *string pointer to VulnerabilityEnvelope .

api/v1/models.go
```diff
@@ -316,3 +319,24 @@ func pageNumberToToken(page database.VulnerabilityNotificationPageNumber, key st
 
 	return string(tokenBytes)
 }
+
+func tokenToNumber(token, key string) (int, error) {
```
jzelinskie: We could definitely only have one function for marshaling and one function for unmarshaling for tokens. Just make it only return an error and you can pass a newly allocated type by reference similar to [normal JSON unmarshalling](https://godoc.org/encoding/json#Unmarshal).

liangchenye:
@jzelinskie  good idea! updated by using tokenMarshal/tokenUnmarshal

api/v1/models.go
```diff
@@ -215,7 +215,8 @@ func NotificationFromDatabaseModel(dbNotification database.VulnerabilityNotifica
 
 	var nextPageStr string
 	if nextPage != database.NoVulnerabilityNotificationPage {
-		nextPageStr = pageNumberToToken(nextPage, key)
+		nextPageBytes, _ := tokenMarshal(nextPage, key)
```
jzelinskie, line 218: `gofmt -s` would rewrite this to `nextPageBytes := pageNumberToToken(nextPage, key)`

api/v1/models.go
```diff
@@ -274,8 +275,10 @@ type NamespaceEnvelope struct {
 }
 
 type VulnerabilityEnvelope struct {
-	Vulnerability *Vulnerability `json:"Vulnerability,omitempty"`
-	Error         *Error         `json:"Error,omitempty"`
+	Vulnerability   *Vulnerability   `json:"Vulnerability,omitempty"`
+	Vulnerabilities *[]Vulnerability `json:"Vulnerabilities,omitempty"`
+	NextPage        *string          `json:"NextPage,omitempty"`
```
jzelinskie: I'm pretty sure this doesn't need to be a pointer. If the string is `""` it should also be excluded via `omitempty`. I'd double check, though.

api/v1/routes.go
```diff
@@ -184,6 +185,59 @@ func getNamespaces(w http.ResponseWriter, r *http.Request, p httprouter.Params,
...
+	pageStrs, pageExists := query["page"]
+	if pageExists {
+		err = tokenUnmarshal(pageStrs[0], ctx.Config.PaginationKey, &page)
+		if err != nil {
+			writeResponse(w, r, http.StatusBadRequest, VulnerabilityEnvelope{Error: &Error{"invalid page format: " + err.Error()}})
+			return getNotificationRoute, http.StatusBadRequest
+		}
+	}
+
+	dbVulns, nextPage, err := ctx.Store.ListVulnerabilities(p.ByName("namespaceName"), limit, page)
```
Quentin-M: Should check if the namespace has been given or 400 otherwise.

api/v1/models.go
liangchenye: You are right, just confirmed.

api/v1/routes.go
liangchenye: Updated.
1. taking the empty namespace as invalid, returns 400.
2. taking the non-exist namespace as valid, but just not found, returns 404.

api/v1/models.go
liangchenye, line 218: Sorry, I don't get it.

jzelinskie, line 218: run `gofmt -s` on this file

api/v1/README.md
```diff
@@ -8,6 +8,7 @@
 - [Namespaces](#namespaces)
   - [GET](#get-namespaces)
 - [Vulnerabilities](#vulnerabilities)
+  - [List](#get-namespacesnsnamevulnerabilities)
```
jzelinskie, line 11: I wish there was a better way to put this, because I was using HTTP verbs for all of these and technically this is a `GET`

Quentin-M, line 11: Yeah but how though?

api/v1/models.go
liangchenye, line 218: Yes, but nothing changes.
And pageNumberToToken is already replaced by tokenMarshal.

jzelinskie, line 218: Ah! I'm sorry for the confusion. Usually `gofmt -s` drops the ignored value if it's second.

Quentin-M:
I think we're pretty much ready to merge that PR @jzelinskie? Except if you can find some better wording for the API docs.

Again, thanks a lot @liangchenye for this great and much appreciated contribution! That's awesome, we're all very excited :+1:

jzelinskie:
LGTM

PR #84, opened 2016-03-01
add makefile for clair

keloyang:
add makefile in clair for easy build.

Signed-off-by: yangshukui yangshukui@huawei.com

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,10 @@
+FROM golang:1.5
+MAINTAINER Quentin Machu <quentin.machu@coreos.com>
+RUN apt-get update && \
+    apt-get install -y bzr rpm xz-utils && \
+    apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+WORKDIR /go/src/github.com/coreos/clair/
+ENV GO15VENDOREXPERIMENT 1
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,22 @@
+.PHONY: all binary build
+
+IMAGE := clair-dev
+CLAIR_IMAGE := clair
+DOCKERFILE := Dockerfile.dev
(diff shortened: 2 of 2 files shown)


coolljt0725:
@Quentin-M  I think this pr intend to make `make binary` to build clair binary and `make clair` to build a clair image

coolljt0725:
With this, it's more convenient to build clair binary and clair image, no need type a long docker cli, just `make binary` or `make image`

Quentin-M:
@coolljt0725 Thanks for the clarification and thanks again for your involvement in Clair. It is much appreciated. I now understand your desire. Having a single Clair binary, generated from a simple command, as well as a container image, sounds like a great idea to us. We could also then distribute the binary on a very mininal container. 

However,
- The modification made to the Dockerfile will definitely break Quay builds of Clair.
- I am afraid that building/distributing/using a single binary of Clair won't just work as expected at the moment because of the current use of `goose` for the database migrations. The library uses a number of `.sql` (and potentially `.go`) files for migrations, which are not embed in the binary and need to be there. I know that some software uses [go-bindata](https://github.com/jteeuwen/go-bindata) as a workaround but I'm not sure that it could work with `goose`. I am open to having another migration tool if it may solve that issue - as long as the migration tool still handles `.go` migrations (as opposed to only managing `.sql` migrations).

keloyang:
I can also keep the old Dockerfile and add a new  makefile for clair.WDYT?

jzelinskie:
> add makefile in clair for easy build.

Is this necessary? Building clair is done idiomatically and without any special flags for both Go and Docker. If the build process gets more complicated in the future, I would see the value in adding a Makefile. In general for Go code, Makefiles are not idiomatic and I'd like to avoid them if possible.

@Quentin-M Your second bullet point is tangentially related and should have its own GitHub issue.

jzelinskie:
I'm going to close this for now. If there is a good counter-argument, I'll re-open it and we can come to a conclusion. Thanks for the PR!

PR #85, opened 2016-03-02
Allow host

keloyang:
"r.Host == allowedHost" is not available for testing the func of clair when clair and analyze-local-images are not in the same host, and if the fileserver  want clair to access,itshould use RemoteAddr to determine which host can access

### DIFF
--- a/contrib/analyze-local-images/main.go
+++ b/contrib/analyze-local-images/main.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
@@ -257,7 +258,8 @@ func listenHTTP(path, allowedHost string) {
 
 	restrictedFileServer := func(path, allowedHost string) http.Handler {
 		fc := func(w http.ResponseWriter, r *http.Request) {
-			if r.Host == allowedHost {
+			host, _, err := net.SplitHostPort(r.RemoteAddr)
+			if err == nil && strings.EqualFold(host, allowedHost) {
 				http.FileServer(http.Dir(path)).ServeHTTP(w, r)
 				return
 			}


Quentin-M:
Hi,

What was the r.Host variable set to in your case? What was the difference with r.RemoteAddr? Is that because it does DNS resolution?

Could we use `net.SplitHostPort(r.RemoteAddr)` instead of `strings.HasPrefix(r.RemoteAddr, ...)`?

keloyang:
r.Host has the style of localip:port, allowedHost is the clair 'host ip, they are not match.

jzelinskie:
> r.Host has the style of localip:port, allowedHost is the clair 'host ip, they are not match.

I think that's why @Quentin-M is suggesting changing

``` go
if strings.HasPrefix(r.RemoteAddr, allowedHost) {
    http.FileServer(http.Dir(path)).ServeHTTP(w, r)
    return
```

to

``` go
host := net.SplitHostPort(r.RemoteAddr)
if strings.EqualFold(host, allowedHost) {
    http.FileServer(http.Dir(path)).ServeHTTP(w, r)
    return
```

keloyang:
yes, I think so,updated,thank you.

contrib/analyze-local-images/main.go
```diff
@@ -257,7 +258,8 @@ func listenHTTP(path, allowedHost string) {
 
 	restrictedFileServer := func(path, allowedHost string) http.Handler {
 		fc := func(w http.ResponseWriter, r *http.Request) {
-			if r.Host == allowedHost {
+			host, _, err := net.SplitHostPort(r.RemoteAddr)
+			if nil == err && strings.EqualFold(host, allowedHost) {
```
jzelinskie: Can you reorder this to `if err == nil && ...`? It's idiomatic to write the error check in this order in Go.

Quentin-M:
Thanks @keloyang! I'm going to try it as soon as possible and then merge it.

Quentin-M:
It works fine with docker-machine on OSX! LGTM.
Sorry for the delay and thanks for the contribution!

PR #87, opened 2016-03-03
database: Look for migrations relative to executable

fatalbanana:
Clair looks for migration scripts beneath _the directory where pgsql.go was at build time_. Somewhere relative to the location of the executable would probably make more sense and may have been what was intended.


fatalbanana:
Closing in favour of #93

PR #88, opened 2016-03-03
fetcher: Send BZR_LOG=/dev/null environment variable to bzr

fatalbanana:
IME (I'm testing on Debian 8), the Ubuntu updater fails on trying to open /nonexistent/.bzr.log - which causes clair to leak socket descriptors. Passing BZR_LOG=/dev/null in environment should solve it.

### DIFF
--- a/updater/fetchers/ubuntu/ubuntu.go
+++ b/updater/fetchers/ubuntu/ubuntu.go
@@ -42,6 +42,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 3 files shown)


jzelinskie:
This broke the build. Can you fix the build?

fatalbanana:
Perhaps this is not actually needed- sorry. I will try to check this later.

PR #89, opened 2016-03-03
database: Find the FeatureVersion we try to insert before doing any lock

Quentin-M:
This commit is issued in order to limit the bottleneck that the
exclusive database lock on Vulnerability_Affects_FeautreVersion
introduces, when we inserting FeatureVersions. This slowdowns a bit
the FeatureVersion insertion on a mostly empty database but should
increase a lot the throughput and parallelism on a populated database.


jakedt:
LGTM

PR #96, opened 2016-03-09
config: add top-level YAML namespace 'clair'

### LINKED ISSUE #95: Namespace YAML Configuration
Our YAML configuration should all be under one top-level key such that the file can contain configuration for other projects in addition to Clair.

jzelinskie:
This enables Clair to use YAML configuration files that contain more
than just a Clair configuration.

Fixes #95.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -13,55 +13,31 @@
 # limitations under the License.
 
 # The values specified here are the default values that Clair uses if no configuration file is specified or if the keys are not defined.
----
-database:
-  # PostgreSQL Connection string
-  # http://www.postgresql.org/docs/9.4/static/libpq-connect.html
-  source:
+clair:
+  database:
+    # PostgreSQL Connection string
+    # http://www.postgresql.org/docs/9.4/static/libpq-connect.html
+    source:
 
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -78,6 +78,12 @@ var DefaultConfig = Config{
 	},
 }
 
+// File represents a YAML configuration file that namespaces all Clair
+// configuration under the top-level "clair" key.
+type File struct {
```
Quentin-M: Should be the first struct declared, above Config.

config.example.yaml
```diff
@@ -13,61 +13,61 @@
...
+    # 32-bit URL-safe base64 key used to encrypt pagination tokens
+    # If one is not provided, it will be generated.
+    # Multiple clair instances in the same cluster need the same value.
+    paginationKey:
 
     # Optional PKI configuration
     # If you want to easily generate client certificates and CAs, try the following projects:
     # https://github.com/coreos/etcd-ca
     # https://github.com/cloudflare/cfssl
-    servername:
```
jzelinskie: oops

Quentin-M:
LGTM

PR #100, opened 2016-03-16
api/v1: create namespace type

### LINKED ISSUE #99: namespace API should return objects
In order to future-proof the API for namespaces, we should return a list of objects for the GET route rather than just a list of strings.

jzelinskie:
This change creates a struct type for namespaces rather than using a
string. This enables us to extend namespaces in the future to contain
metadata. This change also required renaming other field references of
namespaces to "NamespaceName".

Fixes #99


Quentin-M:
LGTM.

PR #105, opened 2016-03-17
contrib: Add a ability to force colored output in analyze-local-images

Quentin-M:
Just like `ls`, this commit introduces the ability to specify `-color always` or `-color never` to force colorization, regardless of whether the TTY is a terminal or not.


josephschorr:
LGTM

PR #110, opened 2016-03-18
misc config improvements

jzelinskie:


Quentin-M:
LGTM.

philips:
thanks, this was causing a crash otherwise, right?

jzelinskie:
Yep, no arguments to Clair used to cause a panic due to nil pointer.

Issue #61, opened 2016-01-12
Question: Does Clair support Mysql database?


coolljt0725:
In our scenario, other services use Mysql as backend database. Does 
Clair support Mysql? If so, there is no need to deploy another database
for us.

Quentin-M:
Hi,

MySQL hasn't been tested explicitly on Cayley (and thus on Clair).

The new database implementation on which I am working on uses PostgreSQL and will provide a huge performance boost and many features as well as caveats fixes compared to the current implementation.

jzelinskie:
Rather than _does_ clair support MySQL, allow me to answer the question _will_ clair eventually support MySQL. There is going to be a new database implementation that uses database/sql and postgres that will be far more performant. Once that is in place, we will be open to accepting PRs that add MySQL support, but it is not a priority for us to work on. However, this will likely be nontrivial because the postgres implementation takes advantage of a couple of features that do not exist in MySQL in order to get decent performance.

coolljt0725:
@jzelinskie  

> > There is going to be a new database implementation that uses database/sql and postgres that will be far more performant

when does this in place?  is the PR comming?

Quentin-M:
@coolljt0725 We are really far in the development already. We are currently writing the new API and starting to run benchmarks on the official Docker images. I will push a branch to this repository as soon as the API will be mostly written. I'd say that the ETA is few weeks.

For now, it sits at https://github.com/Quentin-M/clair/tree/sql.

coolljt0725:
@Quentin-M  Thank you :)

PR #114, opened 2016-03-22
[WIP] database: add support mysql

coolljt0725:
#61

This is work in progress, most code are the same with `pgsql` include the unit tests, all the tests can pass execpt `complex_test.go`  flaky, sometimes it pass and some times fail, I need fix this.  And the code is not clean.
I don't know if this is the right approach to implement `mysql`, I need some feedback and then move on.

### DIFF
--- a/clair.go
+++ b/clair.go
@@ -20,12 +20,15 @@ import (
 	"math/rand"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
 	"github.com/coreos/clair/api"
 	"github.com/coreos/clair/api/context"
 	"github.com/coreos/clair/config"
+	"github.com/coreos/clair/database"
+	"github.com/coreos/clair/database/mysql"
 	"github.com/coreos/clair/database/pgsql"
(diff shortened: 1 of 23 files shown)


Quentin-M:
Hi !

Thanks for the WIP.

How different is it from the `pgsql` driver? This looks like a lot of duplicated code. 
Couldn't we just abstract the pgsql driver a bit instead? Would a library like `gorm` fit since they support SQL Builder now (recursive queries...)?

coolljt0725:
@Quentin-M  Yes, there are a lot of duplicated code, I'll refactor in the next step

nmaas87:
Does Clair support mysql now :)?

jzelinskie:
This PR is quite old now and needs a lot of work to integrate into the current version of Clair.
Anyone still interested in implementing a MySQL backend for Clair should wait for the next major release (v3.0.0) in which we will have a new interface that should make implementing MySQL support significantly more simple.

mowangdk:
Does Clair support mysql now ？。。。

PR #118, opened 2016-03-23
contrib: Use `return` instead of `os.Exit(1)` in analyze-local-images

### LINKED ISSUE #117: contrib/analyze-local-image pollutes /tmp
As [os.Exit() in go skips all deferred instructions](https://gobyexample.com/exit), the statement `defer os.RemoveAll(path)` on [line 116](https://github.com/coreos/clair/blob/master/contrib/analyze-local-images/main.go#L116) gets executed _only_ when our image has visible vulnerabilities.

So almost every time this tool is used the temporary directory `/tmp/analyze-local-image-*` remains.

Quentin-M:
In order to cleanup properly the allocated resources (`defer`).
Note that we still use os.Exit(1) for command line errors.

Fixes #117

### DIFF
--- a/contrib/analyze-local-images/main.go
+++ b/contrib/analyze-local-images/main.go
@@ -45,10 +45,10 @@ const (
 )
 
 var (
-	endpoint        = flag.String("endpoint", "http://127.0.0.1:6060", "Address to Clair API")
-	myAddress       = flag.String("my-address", "127.0.0.1", "Address from the point of view of Clair")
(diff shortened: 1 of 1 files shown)


Quentin-M:
PTAL

jzelinskie:
Ideally the program should exit non-zero on failure. You should pull out everything that isn't flags into a function and let it return an error, we can get a proper exit code.

Quentin-M:
PTAL

jzelinskie:
LGTM

PR #131, opened 2016-04-04
Update README.md

candita:
Add two missing steps to get analyze-local-images built.


contrib/analyze-local-images/README.md
```diff
@@ -9,6 +9,8 @@ There are absolutely no guarantees and it only uses a minimal subset of Clair's
 To install the tool, simply run the following command, with a proper Go environment:
 
     go get -u github.com/coreos/clair/contrib/analyze-local-images
```
jzelinskie, line 11: We could also just change this line to `goinstall` rather than `go get`. It'll pull in the dependencies, right?

Quentin-M, line 11: deps are vendored anyways

jzelinskie:
This PR is now irrelevant because `analyze-local-images` is deprecated and lives in another repository.

PR #140, opened 2016-04-15
config: not properly loaded error

### LINKED ISSUE #134: Connection refused between docker container of clair and pgsql
When I try to run the docker container of clair, there is something wrong with the connection to my pgsql. The docker logs are quoted as follows.

```
2016-04-05 10:27:47.078096 I | pgsql: running database migrations
2016-04-05 10:27:47.079614 E | pgsql: dial tcp [::1]:5432: getsockopt: connection refused
2016-04-05 10:27:47.079655 C | main: database: could not open database
```

I have tried several ways and on different machines and the issue still exists.

jgsqware:
When a wrong config file was used, it result in a panic.
Adding some check condition to validate the Unmarshaled configuration
before using it

fixes #134

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -112,10 +112,12 @@ func Load(path string) (config *Config, err error) {
 		return
 	}
 	config = &cfgFile.Clair
-	if config.API == nil || config.Database == nil || config.Notifier == nil || config.Updater == nil {
-		err = cerrors.ErrConfigNotLoaded
+
+	if config.Database.Source == "" {
+		err = cerrors.ErrDatasourceNotLoaded
 		return
 	}
+
 	// Generate a pagination key if none is provided.
 	if config.API.PaginationKey == "" {
 		var key fernet.Key
--- a/config/config.go
+++ b/config/config.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"time"
 
+	cerrors "github.com/coreos/clair/utils/errors"
 	"github.com/fernet/fernet-go"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -111,7 +112,10 @@ func Load(path string) (config *Config, err error) {
 		return
 	}
 	config = &cfgFile.Clair
-
+	if config.API == nil || config.Database == nil || config.Notifier == nil || config.Updater == nil {
```
Quentin-M: This isn't going to work because the default configuration is loaded first and these fields will never be nil. Instead, we might simply add a check in the database code that tries to open the database.

Quentin-M: Or find a way to verify that there's a `clair` key somehow.

jgsqware:
it's what I'm thinking of but that mean unmarshaling yaml than merge 

I can check for the db connection is the only critical value in the config? 

You had a health check previously for the db connection  isn't it?

jgsqware:
From the configuration Load method,  it could be overhead to check for the database connection.

jgsqware:
Could I supposed that only the database source shouldn't be empty?

utils/errors/errors.go
```diff
@@ -30,8 +30,8 @@ var (
 	// ErrCouldNotParse is returned when a fetcher fails to parse the update data.
 	ErrCouldNotParse = errors.New("updater/fetchers: could not parse")
 
-	//ErrConfigNotLoaded is returned when the configuration file is not loaded properly
-	ErrConfigNotLoaded = errors.New("could not load configuration properly")
+	// ErrDatasourceNotLoaded is returned when the datasource variable in the configuration file is not loaded properly
+	ErrDatasourceNotLoaded = errors.New("could not load datasource configuration properly")
```
Quentin-M: We try to declare the errors are locally as possible - it should be in config/config.go as this error is very specific and not meant to be re-used.

Also, I am not sure that everybody will understand that error message. Maybe something like `could not load configuration: no database source specified` would be more appropriate?

jgsqware: OK don't know for the locality of errors. I will move it. 
Better message indeed. Thanks for the feedback

Quentin-M:
LGTM.

Thank you so much for your contribution!

PR #156, opened 2016-04-23

liangchenye:
Signed-off-by: liang chenye liangchenye@huawei.com


liangchenye:
close, since the sql operation did not pass the testing.

Issue #150, opened 2016-04-20
detect multiply namespaces in a layer


liangchenye:
I met this issue when I work on https://github.com/coreos/clair/issues/40.

The current concept of 'namespace' is inherited from 'operating system', so the `debian/redhat/ubuntu` data source which will never coexist with each other in a layer. But if we count npm/python in, we will have multiple namespaces in a layer. 
 I think we need to modify currently scanning process a bit like this:
1. detect multiple namespaces
     `namespace.go:DetectNamespace()` should return a list of `database.Namespace`
2. detect multiple data if we find multiple namespaces
3. scan the layer by multiple namespace database
   layer struct { `Namespace     *Namespace` } should be a list of `database.Namespace`

We can explain 'namespace' in this way: a self-consistent package system. Although `Namespace - a context around features and vulnerabilities` is a better explanation.

Quentin-M:
Well, we anticipated this issue - which is I believe a big leap we need to solve.

For the records, the reason we detect and store namespaces for each layer is because we might not find this information in a particular layer - but only in a previous layer. For example, a layer that simply does `apt-get install` won't modify the files that tell us that this command actually installs `debian:8` packages - and thus, we won't be able to detect it at that point.

If we allow detecting multiple namespaces, then we'll have to pass them to the feature detectors and let them decide which namespace to use for their packages. For instance, if we pass `{"debian:8", "python:2.7"}` to the `dpkg` detector, it should make the decision to use `debian:8` to tag the detected features. This is kind of related to this comment: [51e4086/worker/worker.go#L168](https://github.com/coreos/clair/blob/51e4086/worker/worker.go#L168). We might want to make the `Namespace` structure better. At least, we should make the name and the version two separates fields. In another discussion, I thought about having a third field, which would be filled by the vulnerability fetchers, telling if there's any vulnerability information for that namespace, and maybe a fourth field describing how we could fix/upgrade the features that belongs to that namespace.

Overall, I think this makes sense and the change has to happen one way or another. We just want to make sure that we do it properly, in a way that will give us the most flexibility and extensibility.

liangchenye:
```
NAMESPACE                                    
   |___ TO BE EXTENDED 
   |___ package system
                 |___  rpm
                 |           |___ centos
                 |           |___ redhat 
                 |                        |_____ 5
                 |                        |_____ 6
                 |                        |_____ 7
                 |___  dpkg
                 |           |___ debian
                 |            |___ ubuntu
[...]

Issue #93, opened 2016-03-08
Improve database migration strategy


Quentin-M:
We currently use [liamstask/goose](https://bitbucket.org/liamstask/goose), it allows doing both .go and .sql migrations. However, it [requires](https://bitbucket.org/liamstask/goose/src/8488cc47d90c8a502b1c41a462a6d9cc8ee0a895/lib/goose/migration_go.go?at=master&fileviewer=file-view-default#migration_go.go-84) a working Go environment and migration files aren't embed in the Clair binary, and thus, we can't distribute a single binary.

jzelinskie:
related project: https://github.com/jteeuwen/go-bindata

Quentin-M:
^ As discussed before, go-bindata is not applicable. It wouldn't help running .go migrations without a go env.

ejholmes:
@Quentin-M we extracted https://github.com/remind101/migrate from https://github.com/remind101/empire. Fairly simple, and designed for embedding.

Quentin-M:
@ejholmes That's awesome. Thank you for the tip!

PR #157, opened 2016-04-25

liangchenye:
Signed-off-by: liang chenye liangchenye@huawei.com

### DIFF
--- a/api/v1/models.go
+++ b/api/v1/models.go
@@ -34,13 +34,13 @@ type Error struct {
 }
 
 type Layer struct {
-	Name             string    `json:"Name,omitempty"`
-	NamespaceNames   []string  `json:"NamespaceNames,omitempty"`
-	Path             string    `json:"Path,omitempty"`
-	ParentName       string    `json:"ParentName,omitempty"`
-	Format           string    `json:"Format,omitempty"`
-	IndexedByVersion int       `json:"IndexedByVersion,omitempty"`
-	Features         []Feature `json:"Features,omitempty"`
+	Name             string      `json:"Name,omitempty"`
+	Namespaces       []Namespace `json:"NamespaceName,omitempty"`
+	Path             string      `json:"Path,omitempty"`
+	ParentName       string      `json:"ParentName,omitempty"`
+	Format           string      `json:"Format,omitempty"`
+	IndexedByVersion int         `json:"IndexedByVersion,omitempty"`
+	Features         []Feature   `json:"Features,omitempty"`
 }
 
 func LayerFromDatabaseModel(dbLayer database.Layer, withFeatures, withVulnerabilities bool) Layer {
@@ -54,26 +54,26 @@ func LayerFromDatabaseModel(dbLayer database.Layer, withFeatures, withVulnerabil
 	}
 
 	for _, namespace := range dbLayer.Namespaces {
-		layer.NamespaceNames = append(layer.NamespaceNames, namespace.Name)
+		layer.Namespaces = append(layer.Namespaces, Namespace{Name: namespace.Name, Version: namespace.Version.String()})
 	}
 
 	if withFeatures || withVulnerabilities && dbLayer.Features != nil {
 		for _, dbFeatureVersion := range dbLayer.Features {
 			feature := Feature{
-				Name:          dbFeatureVersion.Feature.Name,
-				NamespaceName: dbFeatureVersion.Feature.Namespace.Name,
-				Version:       dbFeatureVersion.Version.String(),
-				AddedBy:       dbFeatureVersion.AddedBy.Name,
+				Name:      dbFeatureVersion.Feature.Name,
+				Namespace: Namespace{Name: dbFeatureVersion.Feature.Namespace.Name, Version: dbFeatureVersion.Feature.Namespace.Version.String()},
+				Version:   dbFeatureVersion.Version.String(),
+				AddedBy:   dbFeatureVersion.AddedBy.Name,
 			}
 
 			for _, dbVuln := range dbFeatureVersion.AffectedBy {
 				vuln := Vulnerability{
-					Name:          dbVuln.Name,
-					NamespaceName: dbVuln.Namespace.Name,
-					Description:   dbVuln.Description,
-					Link:          dbVuln.Link,
-					Severity:      string(dbVuln.Severity),
-					Metadata:      dbVuln.Metadata,
+					Name:        dbVuln.Name,
+					Namespace:   Namespace{Name: dbVuln.Namespace.Name, Version: dbVuln.Namespace.Version.String()},
+					Description: dbVuln.Description,
+					Link:        dbVuln.Link,
+					Severity:    string(dbVuln.Severity),
+					Metadata:    dbVuln.Metadata,
 				}
 
 				if dbVuln.FixedBy != types.MaxVersion {
@@ -89,18 +89,19 @@ func LayerFromDatabaseModel(dbLayer database.Layer, withFeatures, withVulnerabil
 }
 
 type Namespace struct {
-	Name string `json:"Name,omitempty"`
+	Name    string `json:"Name,omitempty"`
+	Version string `json:"Version, omitempty"`
 }
 
 type Vulnerability struct {
-	Name          string                 `json:"Name,omitempty"`
-	NamespaceName string                 `json:"NamespaceName,omitempty"`
-	Description   string                 `json:"Description,omitempty"`
-	Link          string                 `json:"Link,omitempty"`
-	Severity      string                 `json:"Severity,omitempty"`
-	Metadata      map[string]interface{} `json:"Metadata,omitempty"`
-	FixedBy       string                 `json:"FixedBy,omitempty"`
-	FixedIn       []Feature              `json:"FixedIn,omitempty"`
+	Name        string                 `json:"Name,omitempty"`
+	Namespace   Namespace              `json:"Namespace,omitempty"`
+	Description string                 `json:"Description,omitempty"`
+	Link        string                 `json:"Link,omitempty"`
+	Severity    string                 `json:"Severity,omitempty"`
+	Metadata    map[string]interface{} `json:"Metadata,omitempty"`
+	FixedBy     string                 `json:"FixedBy,omitempty"`
+	FixedIn     []Feature              `json:"FixedIn,omitempty"`
 }
 
 func (v Vulnerability) DatabaseModel() (database.Vulnerability, error) {
@@ -121,7 +122,7 @@ func (v Vulnerability) DatabaseModel() (database.Vulnerability, error) {
 
 	return database.Vulnerability{
 		Name:        v.Name,
-		Namespace:   database.Namespace{Name: v.NamespaceName},
+		Namespace:   database.Namespace{Name: v.Namespace.Name, Version: types.NewVersionUnsafe(v.Namespace.Version)},
 		Description: v.Description,
 		Link:        v.Link,
 		Severity:    severity,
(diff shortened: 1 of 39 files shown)


liangchenye:
Hi @jzelinskie @Quentin-M , the PR follows the suggestion from you in https://github.com/coreos/clair/issues/150, PTAL.

It makes the following changes:
1.  change *namespace to []namespace in struct `Layer`  
2.  return 0~N namespaces while detecting a layer and let the underling feature detector to decide which namespace does it support 
3.  add a new table - LayerNamespace
   remove `namespace_id` from talbe `Layer` and add the related soi/search/delete SQL accordingly.
   When insert a layer into the database, merge its own namespace with its parent's namespace 
   and insert them into LayerNamespace.
4. modify the related test cases

database/pgsql/layer.go
```diff
@@ -238,28 +250,51 @@ func (pgSQL *pgSQL) InsertLayer(layer database.Layer) error {
...
 
 		parentID = zero.IntFrom(int64(layer.Parent.ID))
-	}
 
-	// Find or insert namespace if provided.
-	var namespaceID zero.Int
-	if layer.Namespace != nil {
-		n, err := pgSQL.insertNamespace(*layer.Namespace)
+		// Find its parent's namespaces
+		rows, err := pgSQL.Query(searchLayerNamespace, parentID)
```
Quentin-M: Why aren't layer.Parent.Namespaces already loaded? We already make the assumption that the provided Parent layer has been retrieved from the database (see L255).

database/pgsql/layer.go
```diff
@@ -238,28 +250,51 @@ func (pgSQL *pgSQL) InsertLayer(layer database.Layer) error {
...
+		id, err := pgSQL.insertNamespace(n)
+		if err != nil {
+			return err
 		}
+
+		// Layer's namespaces has high priority than its parent.
+		// Once a layer has a 'same' namespace (the content before ':' is the same)
+		// with its parent, it will only keep its namespace.
+		name := strings.Split(n.Name, ":")
+		mapNamespaceIDs[name[0]] = id
```
Quentin-M: I understand what you are doing here and this is absolutely right. I think that instead of doing this split around `:`, we should probably extend `Namespace`. This is quite a big design decision and we need to discuss more with @jzelinskie about it.

database/pgsql/layer.go
liangchenye: The layer.Parent.Namespaces are loaded in work.go.
But in layer_test, it is not. In FindLayer (around L55), the parent's ID and name are loaded, but parent's namespace is not.

As we already make the assumption that the provided Parent layer has been retrieved.
So either we modify the related test cases or load parent's namespace in FindLayer.
My second commits did the later choice, it is easier for developers who use database interface directly but it will be time consuming.

database/pgsql/layer.go
Quentin-M: Basically we don't want to assume that every `Namespace` are formatted with a colon because it's up to the `NamespaceDetectors` but we need a way to replace some in the list in case of an upgrade for instance (like `debian:7` to `debian:8`). I wonder if we should let the NamespaceDetector modify the list of Namespaces and thus, let them do the necessary replacements. This is definitely tough.

liangchenye: We can extend 'Namespace' by following the os-release spec. (https://www.freedesktop.org/software/systemd/man/os-release.html)
The `name` is sufficient to the namespace replacement.
The 'version' is used to identify the namespace difference.

api/v1/models.go
```diff
@@ -35,7 +35,7 @@ type Error struct {
 
 type Layer struct {
 	Name             string    `json:"Name,omitempty"`
-	NamespaceName    string    `json:"NamespaceName,omitempty"`
+	NamespacesName   []string  `json:"NamespacesName,omitempty"`
```
jzelinskie: NamespaceNames

worker/worker.go
```diff
@@ -150,43 +154,42 @@ func detectContent(name, path, imageFormat string, parent *database.Layer) (name
...
 
 	return
 }
 
-func detectFeatures(name string, data map[string][]byte, namespace *database.Namespace) (features []database.FeatureVersion, err error) {
+func detectFeatures(name string, data map[string][]byte, namespaces []database.Namespace) (features []database.FeatureVersion, err error) {
 	// TODO(Quentin-M): We need to pass the parent image DetectFeatures because it's possible that
 	// some detectors would need it in order to produce the entire feature list (if they can only
 	// detect a diff). Also, we should probably pass the detected namespace so detectors could
 	// make their own decision.
```
jzelinskie: This comment should be updated now that we pass the namespaces

worker/worker.go
```diff
@@ -150,43 +154,42 @@ func detectContent(name, path, imageFormat string, parent *database.Layer) (name
...
-func detectNamespace(data map[string][]byte, parent *database.Layer) (namespace *database.Namespace, err error) {
-	namespace = detectors.DetectNamespace(data)
+func detectNamespaces(data map[string][]byte, parent *database.Layer) (namespaces []database.Namespace, err error) {
+	namespaces = detectors.DetectNamespaces(data)
+
+	// Inherit the non-detected namespace from its parent
+	if parent != nil {
+		mapNamespaces := make(map[string]database.Namespace)
+		for _, pn := range parent.Namespaces {
+			name := strings.Split(pn.Name, ":")
```
jzelinskie, line 165: These splits end up being pretty ugly.

Down the road, we should probably have a version property for namespaces. Introducing it now would break the v1 API, though.

worker/detectors/features.go
```diff
@@ -54,15 +56,24 @@ func RegisterFeaturesDetector(name string, f FeaturesDetector) {
...
-		pkgs, err := detector.Detect(data)
-		if err != nil {
-			return []database.FeatureVersion{}, err
+		for _, namespace := range namespaces {
+			if detector.Supported(namespace) {
+				pkgs, err := detector.Detect(data)
+				if err != nil {
+					return []database.FeatureVersion{}, err
+				}
+				// Ensure that every feature has a Namespace associated
```
jzelinskie, line 69: Is this necessary? In what scenario is this value not set?

worker/detectors/features.go
```diff
@@ -28,6 +28,8 @@ type FeaturesDetector interface {
 	// GetRequiredFiles returns the list of files required for Detect, without
 	// leading /.
 	GetRequiredFiles() []string
+	//Supported check if the input Namespace is supported by the underling detector
```
jzelinskie:
``` go
  // Supported checks if the input Namespace is supported by the underlying detector.
```

database/pgsql/layer.go
```diff
@@ -55,12 +53,46 @@ func (pgSQL *pgSQL) FindLayer(name string, withFeatures, withVulnerabilities boo
...
+			return layer, handleError("searchLayerNamespace.Rows()", err)
+		}
 	}
-	if !namespaceID.IsZero() {
-		layer.Namespace = &database.Namespace{
-			Model: database.Model{ID: int(namespaceID.Int64)},
-			Name:  namespaceName.String,
+
+	// Find its namespaces
+	rows, err := pgSQL.Query(searchLayerNamespace, layer.ID)
```
jzelinskie, line 81: This query also probably needs the `observeQueryTime` executed afterwards.

database/pgsql/queries.go
```diff
@@ -112,12 +111,17 @@ const (
 						AND v.namespace_id = vn.id
 						AND v.deleted_at IS NULL`
 
+	searchLayerNamespace = `
+		SELECT n.id, n.name
+		From LayerNamespace ln, Namespace n
```
jzelinskie: FROM

worker/detectors/feature/dpkg/dpkg.go
```diff
@@ -113,3 +113,16 @@ func (detector *DpkgFeaturesDetector) Detect(data map[string][]byte) ([]database
 func (detector *DpkgFeaturesDetector) GetRequiredFiles() []string {
 	return []string{"var/lib/dpkg/status"}
 }
+
+//Supported check if the input Namespace is supported by the underling detector
```
jzelinskie:
``` go
  // Supported checks if the input Namespace is supported by the underlying detector.
```

worker/detectors/feature/rpm/rpm.go
```diff
@@ -118,3 +118,16 @@ func (detector *RpmFeaturesDetector) Detect(data map[string][]byte) ([]database.
 func (detector *RpmFeaturesDetector) GetRequiredFiles() []string {
 	return []string{"var/lib/rpm/Packages"}
 }
+
+//Supported check if the input Namespace is supported by the underling detector
```
jzelinskie:
``` go
  // Supported checks if the input Namespace is supported by the underlying detector.
```

jzelinskie:
When complete, this change will also need a database migration to move all current installations to the new schema.

worker/worker.go
liangchenye: So, should I add a `//TODO` to note it and add a version property for namespaces in v2 ?

worker/detectors/features.go
liangchenye, line 69: Do you mean the 'Namespace' value? It is not set in featureDetectors. 
In the previous implementation, it was filled by work.go/detectFeatures.
Since we support multiple namespaces, I move that codes here.

liangchenye:
Add  'version' field to `Namespace`. (commit: add 'version' to namespace')

I have two questions:
1. Since it breaks V1, should there be a V2 branch?
    (commit: testing commit: modify V1 by add Namespace.Version)
2. about the database migration, is that mean should we keep the origial 2015*.sql and anther `sql` file? Also, should we add a .go file to migrate v1 data to this new version?

jzelinskie:
This should be rebased now that it cannot be merged anymore and then reviewed again.

Master is going to be the branch for "v2" work, we're managing the branches similar to etcd where we have branches for each minor release.

The idea behind migrations is to always move forward by adding new files. This includes writing code (not just SQL) to upgrade from the old data schemas.

liangchenye:
Rebased in my branch https://github.com/liangchenye/clair/tree/multiple-detector.

I don't know why it becomes an 'unknown' repository. Once the upgrading code was done, I'll add a new PR and close this one.

liangchenye:
Thanks @jzelinskie , some create/alter table work could be done by SQL, but the sequence between SQL and code matters. So now I'm working on adding a pure migration code.

liangchenye:
Hi @jzelinskie , I don't know if you have ever meet the same [problem](https://bitbucket.org/liamstask/goose/issues/63/failed-to-migrate-go-file-when-goose-lib) in goose.
Seems goose does not work well with 'vendor' ?

And another thing is, a user has to have goose lib under his/her GOPATH directory.
The goose golang migration is quite different with goose sql migration, it genereates `.go` files under a `/tmp/goose*` directory and then 'go run' them. So Clair will depend on a go-dev environment , is it acceptable?

liangchenye:
As comment above, .go migration is not well supported by goose. (to my knowledge).
I change to use 'pure' sql way.
It works good and pass tests  :)  https://github.com/liangchenye/clair/tree/multiple-detector.
I'll review that thoroughly and submit again next week.

database/pgsql/layer.go
liangchenye: in new PR, I'll update test cases. So will not reload parent's namespaces.

liangchenye:
close and work on https://github.com/coreos/clair/pull/193

Quentin-M:
Hi,

Thank you again for your work on this.

_liamstask/goose_ is clearly problematic. Issue #93 tracks it. There is a terrible lack of solid solution for database migration in Go. As far as I know, none of them except _liamstask/goose_ support doing migrations using go code and none can do it without having a go env. That is a huge drawback for any Go app that use databases and that can pretend to be ran in production.

PR #158, opened 2016-04-25
contrib: Catch signals to delete tmp folder in local-analyze-images

Quentin-M:
- Create the temporary folder earlier in the app (main instead of save)
- Catch SIGINT/SIGKILL signals to delete the temporary folder and exit

### DIFF
--- a/contrib/analyze-local-images/main.go
+++ b/contrib/analyze-local-images/main.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"os/signal"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


contrib/analyze-local-images/main.go
```diff
@@ -110,24 +111,40 @@ func main() {
...
+	go func() {
+		<-interrupt
+		os.RemoveAll(tmpPath)
+		os.Exit(130)
+	}()
+
+	// Analyze the image.
+	err = AnalyzeLocalImage(imageName, minSeverity, *flagEndpoint, *flagMyAddress, tmpPath)
+	if err != nil {
+		log.Print(err)
```
jzelinskie: this case is going to error, but exit 0

jzelinskie:
Let's wrap main so that we can return custom errors and get custom exit codes.

``` go
package main

import "fmt"
import "os"

func doTheStuff() int {
    defer fmt.Println("!")

    return 3
}

func main() {
    os.Exit(doTheStuff())
}
```

Let's also block on a select at the bottom of the function waiting for analyzelocalimage or the interrupt

Quentin-M:
Meh ^

jzelinskie:
LGTM

PR #166, opened 2016-05-04
api/worker: introduce optional authorization

jzelinskie:
This allows clients to specify the contents of the HTTP Authorization
header so that Clair can access protected resources.


Quentin-M:
LGTM

PR #167, opened 2016-05-05
api/worker: s/Authorization/Headers

jzelinskie:
This allows clients to specify any HTTP headers that need to be used in
order to allow Clair to download a layer, rather than just the
Authorization header.


Quentin-M:
LGTM

PR #174, opened 2016-05-13
Plug FD leak in Debian updater

fatalbanana:
I've found Clair to leak FDs (sockets). This fixed it for me.

### DIFF
--- a/updater/fetchers/debian/debian.go
+++ b/updater/fetchers/debian/debian.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
 
@@ -67,8 +68,11 @@ func (fetcher *DebianFetcher) FetchUpdate(datastore database.Datastore) (resp up
 	r, err := http.Get(url)
 	if err != nil {
 		log.Errorf("could not download Debian's update: %s", err)
+		io.Copy(ioutil.Discard, r.Body)
 		return resp, cerrors.ErrCouldNotDownload
 	}
+	defer r.Body.Close()
+	r.Close = true
 
 	// Get the SHA-1 of the latest update's JSON data
 	latestHash, err := datastore.GetKeyValue(updaterFlag)


jzelinskie:
I think I've read a discussion about this corner case somewhere -- maybe the Go mailing list. I'm going to find said discussion before I merge to ensure this is the best approach.

Afterwards, we should audit all of our calls to `http.Get` or `http.Do` to ensure we clean up properly.

Thanks for the PR!

Quentin-M:
Hi,

Thanks for your contribution. It would be interesting to see if we didn't miss that somewhere else.

I have a question though:
- Discarding the r.Body before calling r.Close() allows the TCP connection to be re-used.
- However, setting r.Close to true prevents the client from reusing the TCP Connection.

What am I missing?

fatalbanana:
So this patch is actually the product of some general speculation as to what leaks may exist, which I have belatedly dusted off and tested _successfully_ - now that I'm no longer making rotten builds ...

The error case is not what I'm seeing: so this is untested: possibly connection should need to be closed there as well. I am not even sure if body needs to be discarded but suspect so.

I had intended to look into this further & refine the patch but I didn't get round to it- it is possible that I could still though.

updater/fetchers/debian/debian.go
```diff
@@ -67,8 +68,11 @@ func (fetcher *DebianFetcher) FetchUpdate(datastore database.Datastore) (resp up
 	r, err := http.Get(url)
 	if err != nil {
 		log.Errorf("could not download Debian's update: %s", err)
+		io.Copy(ioutil.Discard, r.Body)
```
Quentin-M, line 71: That line would generates an `invalid memory address or nil pointer dereference`.

updater/fetchers/debian/debian.go
```diff
@@ -67,8 +68,11 @@ func (fetcher *DebianFetcher) FetchUpdate(datastore database.Datastore) (resp up
...
 		return resp, cerrors.ErrCouldNotDownload
 	}
+	defer r.Body.Close()
+	r.Close = true
```
Quentin-M, line 75: That's useless, it has to be set before the request is done. `http.NewRequest()` must be used, then `r.Close = true` and then `http.DefaultClient.Do(request)`. Plus it prevents TCP connections from being re-used and I don't we really need that.

updater/fetchers/debian/debian.go
Quentin-M, line 71: That kind of copy is only useful when `r.Body` isn't read completely so the TCP connection can be re-used. As long as the body is read and then closed, the TCP connection can be re-used.

Quentin-M:
Really I think we only need the missing `defer r.Body.Close()` there.

Quentin-M:
It appears that the `json.Decoder` may not read entirely the body (i.e. parser error). It wouldn't generate any leaks but will prevent Go from re-using connections. Thus, it might be valuable for us to create an helper function that discards and closes the body - and then use that function on defer everywhere in the code.

fatalbanana:
Had persistent problems with this for the longest time but I really can't replicate it now, _shrugs_\- sorry about this ...

PR #179, opened 2016-05-18
pgsql: remove unnecessary join

jzelinskie:


Quentin-M:
Can you mention in the commit message that it is about `GetNotification`? Other than that, LGTM.
This should be included in the next point-release.

PR #180, opened 2016-05-19
Updated API documentation to reflect changes to authentication scheme.

sarnowski:
With #167, the API was changed to provide a more generic way of how Clair can retrieve the layers from a registry. This change reflects it in the documentation.


jzelinskie:
Thanks! I totally forgot to update that doc!

PR #182, opened 2016-05-19
pgsql: use subquery to plan GetNotification query

jzelinskie:
This change enables the query planner to wait and sort the result set of
our query rather than attempting to re-use the layer table's index for
the ORDER BY clause. Because the result set is always small, this makes
queries that were previous tens of seconds, now tens of milliseconds.


Quentin-M:
LGTM

PR #184, opened 2016-05-24
Added json output option.

mkulke:
This adds a `-json` flag to the `analyze-local-images` tool to be able to further process the data.


jzelinskie:
If you're looking for machine readable output, why not use the [API](https://github.com/coreos/clair/blob/master/api/v1/README.md) directly rather than going through the CLI tool?

mkulke:
I considered doing this, but there's some logic in the cli tool to extract
the image, resolve the layers from the image tag, etc, which is not present
in the API yet.

An endpoint which accepts an image tag and provides the vulnerabilities of
its layers would be great, but I guess this would be quite an effort, no?

Jimmy Zelinskie notifications@github.com schrieb am Di., 24. Mai 2016
20:43:

jgsqware:
you could use github.com/wemanity-belgium/hyperclair

the command expose a 'hyperclair report - -format json <image>/<tag>'

it will replace analyze image tool soon

mkulke:
@jgsqware i looked at this yesterday, seems the docker-machine support was merged just now. I was wondering whether it would happen at all, since the PR was old. I'm fine with closing the PR then.

PR #193, opened 2016-05-31

liangchenye:
This implementation mainly adds a 'version' field to database.namespace, this change affects the following things:

1) API
     Since namespace.Name could not represent a namespace, I use the encoded namespace.ID in [Vulnerabilities API ](https://github.com/liangchenye/clair/blob/multiple-detector/api/v1/README.md#vulnerabilities) and [Fixes API](https://github.com/liangchenye/clair/blob/multiple-detector/api/v1/README.md#fixes)
2) vulnerability & fixes sql operation
   Use namespaceID to add/remove/change vulnerabilities.
3) database migration 
   using [sql migration file](https://github.com/liangchenye/clair/blob/multiple-detector/database/pgsql/migrations/20160524162211_multiple_namespace.sql)
detectors
4) test cases
   for example change debian:8 to {Name: "debian", Version: "8"} 
5) debian 'unstable'
   'unstable' is illegal as a version content, change it to '10'.

Close the previous [PR](https://github.com/coreos/clair/pull/157) since that 'head fork' became 'unknown'.

### DIFF
--- a/api/v1/README.md
+++ b/api/v1/README.md
@@ -129,18 +129,18 @@ Server: clair
 {
   "Layer": {
     "Name": "17675ec01494d651e1ccf81dc9cf63959ebfeed4f978fddb1666b6ead008ed52",
-    "NamespaceName": "debian:8",
+    "Namespaces": [{"Name": "debian", "Version": "8"}],
     "ParentName": "140f9bdfeb9784cf8730e9dab5dd12fbd704151cf555ac8cae650451794e5ac2",
     "IndexedByVersion": 1,
     "Features": [
       {
         "Name": "coreutils",
-        "NamespaceName": "debian:8",
+        "Namespace": {"Name": "debian", "Version": "8"},
         "Version": "8.23-4",
         "Vulnerabilities": [
           {
             "Name": "CVE-2014-9471",
-            "NamespaceName": "debian:8",
+            "Namespace": {"Name": "debian", "Version": "8"},
             "Description": "The parse_datetime function in GNU coreutils allows remote attackers to cause a denial of service (crash) or possibly execute arbitrary code via a crafted date string, as demonstrated by the \"--date=TZ=\"123\"345\" @1\" string to the touch or date command.",
             "Link": "https://security-tracker.debian.org/tracker/CVE-2014-9471",
             "Severity": "Low",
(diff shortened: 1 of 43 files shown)


Quentin-M:
Hi,

Thanks for your work here!

About (1), I understand why you encrypt the namespace IDs in the API as it is preferred not to leak database IDs. However, I believe that usually, the list of Namespaces is pretty much stable as it depends mainly on the registered detectors/updaters (and not on the number of images we analyzed). I am not sure if we should make this trade-off at this level between confidentiality and readability/usability. We shall make a decision and stick to it for any later development. I'd ping @jzelinskie as he wrote the API.

About (5), is there anything that led you to use the `Version` type instead of a more general string? I think it might limit extensibility. For instance, you chose to replace _unstable_ by 10 here, however, Debian 10 will eventually exist in parallel to Debian Unstable.

api/v1/README.md
```diff
@@ -220,7 +220,7 @@ The GET route for the Vulnerabilities resource displays the vulnerabilities data
 ###### Example Request
 
 ```json
-GET http://localhost:6060/v1/namespaces/debian%3A8/vulnerabilities?limit=2 HTTP/1.1
+GET http://localhost:6060/v1/namespaces/gAAAAABXTQKgma_TLKq0wr1D6wVB507N3fi9wsWMUypYOSXTDVxQ8OR5L5S6PqZ9Wh0IzWojnVmlspyTL4cyjytyra7U9vAHMA==/vulnerabilities?limit=2 HTTP/1.1
```
jzelinskie, line 223: I'd rather this be something along the lines of `/v1/namespaces/debian/versions/9/vulnerabilities?limit=2` rather than using the encrypted token. Obviously, it'll be slower on the database if we don't already have the ID, but I'd rather make that sacrifice for a sane API.

Quentin-M, line 223: +1 for that

liangchenye, line 223: OK, so the other APIs could also be implement in this way.

liangchenye:
> > About (1),

I'll use ':nsname/version/:nsversion' in all the related APIs.

> > About (5), is there anything that led you to use the Version type instead of a more general string? I think it might limit extensibility. For instance, you chose to replace unstable by 10 here, however, Debian 10 will eventually exist in parallel to Debian Unstable.

You are right. My original idea is we need to compare versions between a layer and its parent. But since parent's namespace version will be covered so a 'string' is sufficient.

jzelinskie:
@liangchenye is this ready for review? Did you change the API?

liangchenye:
@jzelinskie  no, I planed to rebase this after https://github.com/coreos/clair/pull/175 been merged.

vbatts:
@liangchenye looking forward to this getting merged! :smile:

jzelinskie:
This PR is now extremely out of date, but did inspire the design decisions which does implement this behavior in #432.

PR #194, opened 2016-05-31
Initial services refactoring of database.Datastore

mattmoor:
This is an initial strawman PR of a proposal I made verbally to Quentin and Brandon last week.

This breaks out the following service interfaces:
- locks.Service
- keyvalue.Service
- notifications.Service
- vulnerabilities.Service
- layers.Service
- namespaces.Service

Roughly, each of the directories under `services/` exposes a similar model to what `database/` exposes currently.
- An interface to be implemented `Service`
- A method to `Register` a named implementation 
- A method to `Open` a named implementation

I have not yet started to pick apart the `pgsql` implementation as implementations of each of these, that would be a subsequent change assuming this change seems reasonable to folks.

I also tweaked `.travis.yml` to work on my fork by rsyncing the build dir to `coreos/clair` and switching to that directory.


mattmoor:
@Quentin-M FYI

jzelinskie:
As per our earlier discussion. I have concerns over maintaining consistency when vulnerabilities, layers, and namespaces are decoupled into potentially separate databases. I think after we're on the same page and you've taken a look out our Postgres schema/queries we should talk about adjusting this strawman or proposing a new one.

mattmoor:
@jzelinskie Let's start a bit smaller, and skirt the consistency issue for now because I think our storage layer is going to run into similar problems.  Also, the way `Namespace` is factored as its own table means that to embed it in `List` operations via `JOIN` will always outperform lookups against another `Service` in code.

What if we started with:
- `locks.Service`
- `keyvalue.Service`
- `notifications.Service`
- `tbd.Service` => `{layers.Service, vulnerabilities.Service, namespaces.Service}`

This last bullet is where virtually all of the `JOIN`s occur at the `SQL` level.

I still think it makes sense for the fetchers to take a more limited interface reflecting their more limited requirements, but that can be a follow up PR.

jzelinskie:
This sounds like a far more reasonable initial change.

As for the name of tbd.Service:
¯_(ツ)_/¯

Quentin-M:
That looks way more reasonable to me too. Performance is a major concern for Clair.

mattmoor:
See: https://github.com/coreos/clair/pull/198

PR #195, opened 2016-06-03
Jkroll cf/no migration

jkroll-cf:
This should make clair more packageable.


Quentin-M:
Hi. Can you explain a little further?

jkroll-cf:
Oh, I meant to make this PR to my fork! Sorry.

I was trying to work around the issue in #93 a different way, by putting the migrations elsewhere in a container build script. But my end goal was to just distribute a binary, albeit one that would depend on the migrations happening separately.

Quentin-M:
I understand. We currently use [liamstask/goose](https://bitbucket.org/liamstask/goose) which [requires](https://bitbucket.org/liamstask/goose/src/8488cc47d90c8a502b1c41a462a6d9cc8ee0a895/lib/goose/migration_go.go?at=master&fileviewer=file-view-default#migration_go.go-84) a working Go environment. We would really like to ship a single binary (and a minimal container) too.

jkroll-cf:
Yes. I was trying to move to a model where I can ship a binary that expects the migrations to be done elsewhere, so that I can factor them into my build scripts and ship a binary package to build a smaller container without Go in it.

You might consider putting migrations behind a flag or allowing the Goose directory to be made an option in the Clair config instead of linking in the migration script as bindata.

Quentin-M:
That makes a lot of sense, of course!

We were thinking about creating a small migration library. It would enable registering migrations, similar to the way [database/sql](https://golang.org/pkg/database/sql/#Register) or Clair does. Thus, we wouldn't have to ship a Go environment to run either .go or .sql defined migrations. Here is a really quick gist that we wrote a while ago: https://gist.github.com/Quentin-M/5436f1d13bebe5b24a7cb8a0e6ada818.

Quentin-M:
@jkroll-cf Given the fact that you were working on this, would you be interested in contributing in order to solve this issue? There is a terrible lack of solid solution for database migration in Go. As far as I know, none of them except _liamstask/goose_ support doing migrations using go code and none can do it without having a go env. That is a huge drawback for any Go app that use databases and that can pretend to be ran in production.

PR #197, opened 2016-06-07
Modify URL for libpq documentation

lbausch:
When opening the URL in the browser the section about "Connection Strings" is directly displayed. This way there's no need to scroll to it. (Also use HTTPS)

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -19,7 +19,7 @@ clair:
     type: pgsql
     options:
       # PostgreSQL Connection string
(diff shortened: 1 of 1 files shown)


jzelinskie:
Let's remove the version from the URL and use "current", while we're at it.
If you change the URL to `https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING`, I'll merge this PR.

Thanks!

lbausch:
Thanks for your feedback, I changed the URL as you said. :)

jzelinskie:
LGTM

PR #202, opened 2016-06-13
travis: Remove 'tip' until 'gimme' fixed

jgsqware:
`gimme`script of Travis has incorrect behavior for the `tip`command. It return _1.4.1_ instead of _master_.

I propose to remove `tip`until travis-ci/gimme#38 is fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: go
 go:
   - 1.6
   - 1.5
-  # - tip
+  - tip
 
 sudo: false
 
@@ -30,3 +30,7 @@ addons:
 
 notifications:
   email: false
+
+matrix:
+  allow_failures:
+    - go: tip
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: go
 go:
   - 1.6
   - 1.5
-  - tip
+  # - tip
 
 sudo: false
 


jzelinskie:
Even better, we can leave it there, but set the failure matrix to ignore its failures.
[This is what Torus does](https://github.com/coreos/torus/blob/b2359cf8bf7cf5ecf267be774646bbf8df3a15d6/.travis.yml#L18-L20).

jgsqware:
What a great id,  I don't even know that's possible. I'll WILL update the
PR

Le lun. 13 juin 2016 16:24, Jimmy Zelinskie notifications@github.com a
écrit :

jgsqware:
@jzelinskie I've updated `.travis.yml` as you say, and it's passing but again, gimme version doesn't change but retrieve the master branch now. I don't get why sometimes the behavior of gimme change.

jzelinskie:
@jgsqware That's fine. I don't think we should enforce `tip` passing even if `gimme` worked perfectly. However, we should be able to see if something will eventually be introduced into the language that will break us and that's what enabling this does.

LGTM

jgsqware:
👍 thanks

Le mar. 14 juin 2016 07:55, Jimmy Zelinskie notifications@github.com a
écrit :

PR #203, opened 2016-06-14
config: fix wrong location of database config in config.example.yaml

cynipe:
This fixes configuration error `main: failed to load configuration: could not load configuration: no database source specified`
when booting clair with following Hello Heartbleed instruction.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -17,14 +17,14 @@ clair:
   database:
     # Database driver
     type: pgsql
-    options:
-      # PostgreSQL Connection string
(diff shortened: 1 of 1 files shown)


jgsqware:
The master branch contains the latest development and the config. yml is related to it. 
Your change will work with latest release but not later. 

I'm not agree with this PR.  @jzelinskie?

cynipe:
Oh, I see. So I'd better to refer this file https://github.com/coreos/clair/blob/release-1.2/config.example.yaml.
However, IMHO, it's should be documented specially like a quick start guide.

Issue #204, opened 2016-06-14
config.example.yaml has data source in the wrong keyspace.


zfletcher:
It fails to run unless it is in the database keyspace:

```
$ docker run --rm -ti -p 6060-6061:6060-6061 -v $PWD/config:/config -v /tmp:/tmp --link clair_postgres --net clair_default quay.io/coreos/clair -config /config/config.yaml
2016-06-14 18:10:02.017211 C | main: failed to load configuration: could not load configuration: no database source specified
```

My patch fixes clair config and now it runs as expected.

https://github.com/zfletcher/clair/commit/c9a2272a94994e7f2bd5b3d16c6f7ff0f9ea3384

```
$ docker logs -f clair_clair
2016-06-14 18:18:17.216011 I | pgsql: running database migrations
goose: no migrations to run. current version: 20151222113213
2016-06-14 18:18:17.294032 I | pgsql: database migration ran successfully
2016-06-14 18:18:17.294260 I | notifier: notifier service is disabled
2016-06-14 18:18:17.294345 I | api: starting main API on port 6060.
2016-06-14 18:18:17.294577 I | api: starting health API on port 6061.
2016-06-14 18:18:17.294685 I | updater: updater service started. lock identifier: 8ac5f141-e6bc-425c-b3ec-592e3f139fb3
2016-06-14 18:18:17.300814 I | updater: updating vulnerabilities
2016-06-14 18:18:17.300918 I | updater: fetching vulnerability updates
2016-06-14 18:18:17.300971 I | updater/fetchers/ubuntu: fetching Ubuntu vulnerabilities
2016-06-14 18:18:17.310008 I | updater/fetchers/debian: fetching Debian vulnerabilities
2016-06-14 18:18:17.310521 I | updater/fetchers/rhel: fetching Red Hat vulnerabilities
2016-06-14 18:28:07.318672 I | updater: adding metadata to vulnerabilities
2016-06-14 18:45:27.597493 I | updater: update finished
```

Please advise if this is a real issue, and if so, do you need me to PR it... ??

jgsqware:
As said in #203, the config.example.yml refer to the master branch, where the configuration is correct. Your patch refer to the release.

@jzelinskie @Quentin-M, we should address this point, we receive a lot of question around that. Any idea?

zfletcher:
Sorry!!  I am using the official image on quay.io/coreos, but was following the Master Readme.. My apologies for the erroneous question.

jgsqware:
[...]

PR #210, opened 2016-06-22
Fix source location in hierarchy

Ulexus:
The `source` field appears to be at `clair.database.source`, not `clair.database.options.source`.  The latter results in clair complaining that no source has been specified, while the former works as expected.


jgsqware:
It's a confusion between Release and Master. Please refer to #204

Ulexus:
Should have investigated.  Thanks.

PR #212, opened 2016-06-27
Add informations for analyze-local-images with Docker for Mac

solidnerd:
Hey,
thanks for clair. It tried it to use on OSX with Docker for Mac and I got  to get it working. I have found the issue so I have found a way how to get it working. I want to document this for others.


jzelinskie:
Thanks for the PR!

I think we might be better off just prefixing L26 with `TMPDIR=/tmp` and maybe extending the paragraph below that, rather than introducing a new section. Technically Docker4Mac is still a remote daemon.

jzelinskie:
`analyze-local-images` is now deprecated and located in another repository.
If this PR is still relevant, it can be made to that repository (or forked and personally maintained).

PR #216, opened 2016-07-15
README: add reference to Klar tool

hashmap:
This update adds reference to Klar tool. Klar is a simple integration
tool designed to be used in scripts and CI.


Quentin-M:
LGTM. Thanks!

Issue #78, opened 2016-02-25
Clair support for Oracle Linux


pipseed:
Are there any plans to expand the OS support for Oracle Linux. Scanning an OL image flags as unrecognized OS, as this is derived from a RedHat OS with an rpm based package manager would this be an easy fix?

Quentin-M:
Hi.

Sure. If the packages are the same, it's just about updating https://github.com/coreos/clair/tree/master/worker/detectors/namespace/redhatrelease to make Oracle Linux detected as CentOS (ie. `centos:6`). Shouldn't be more than 10 lines of code.

Actually, to be more accurate, we should also:
- rename `centos` to `rhel` there,
- rename `centos` to `rhel` in `updater/fetchers/rhel/rhel.go`,
- rename every existing Namespace `centos:*` to `rhel:*` in databases.

philips:
@pipseed would you be interested in helping out with this?

Djelibeybi:
@Quentin-M Oracle Linux has it's own OVAL database and prefixes, so I'm not sure if the existing `rhel` fetcher would work. 

I've been meaning to take a look at Clair for a while, so I might do that and take a look at this at the same time.

(FYI: I'm the Product Manager for Oracle Linux)

PR #220, opened 2016-07-21
Initial support for Oracle Linux

Djelibeybi:
(Please be gentle when reviewing, this is my first PR to a Go project).

This addresses issue #78 and adds support for Oracle Linux by creating a specific fetcher to grab OVAL content from Oracle and extends the `osrelease` and `redhatrelease` detectors to detect Oracle Linux.

I couldn't find the order in which the detectors are run, so I've adjusted both. Oracle Linux 5 triggers `redhatrelease` while OL6 and OL7 will be detected by `osrelease` because we ship an `/etc/os-release` on both.

Signed-off-by: Avi Miller avi.miller@oracle.com

### DIFF
--- a/cmd/clair/main.go
+++ b/cmd/clair/main.go
@@ -30,6 +30,7 @@ import (
 
 	_ "github.com/coreos/clair/updater/fetchers/debian"
 	_ "github.com/coreos/clair/updater/fetchers/rhel"
+	_ "github.com/coreos/clair/updater/fetchers/oracle"
 	_ "github.com/coreos/clair/updater/fetchers/ubuntu"
 	_ "github.com/coreos/clair/updater/metadata_fetchers/nvd"
 
--- a/updater/fetchers/oracle/oracle.go
+++ b/updater/fetchers/oracle/oracle.go
@@ -40,9 +40,7 @@ const (
 var (
 	ignoredCriterions = []string{
 		" is signed with the Oracle Linux",
-		" Client is installed",
-		" Workstation is installed",
-		" ComputeNode is installed",
+		".ksplice1.",
 	}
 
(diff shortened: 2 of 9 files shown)


Djelibeybi:
Note that Clair will output warnings for some OVAL files because we ignore the Ksplice Userspace packages (versions with .ksplice1. in the value). This results in a warning because the package list becomes empty.

Not sure what the best way to suppress these warnings is.

Quentin-M:
Thank you so much for your contribution!
- The decision you made about the detectors looks fine to me.
- Could you please quote one of the warnings here? I am not absolutely sure which ones you are facing.
- I just skimmed very quickly at your PR, as a first note, I'd say that we should extract the OVAL parsing code to the utility package as it is used by both RHEL and Oracle. Even better, if the two fetchers look very similar, we could simply try to use the same code for both.

Djelibeybi:
Actually, we've made a change to our OVAL files so that the ksplice warnings shouldn't appear any more. 

I agree that we should extract the OVAL parsing and possibly the fetchers as well, but as I mentioned, my Go experience is limited to this PR so far. I'm traveling this week, but may have some time next week to try my hand at this.

jzelinskie:
> I'd say that we should extract the OVAL parsing code to the utility package as it is used by both RHEL and Oracle.

OpenSUSE also uses OVAL and this has been done in the OpenSUSE PR.
@Djelibeybi that PR is likely to be merged first, then you'll have to rebase your work to use that new OVAL utility. Is that okay?

Djelibeybi:
@jzelinskie that suits me fine. I'm on leave for a couple of weeks, so I can take a look at it again once your PR is merged.

Djelibeybi:
I am going to close this PR and submit a new one based on the latest code.

jzelinskie:
sounds good!

PR #229, opened 2016-08-12
Redhatrelease detector

vbatts:
Due to the detector registration and fact that their in a non-ordered
map, it is random whether the osrelease or redhatrelease detector would
hit. And likely resulted in alternately formatted namespace strings.

This change causes the osrelease to not detect when data has
centos-release or redhat-release, which is not _great_ because if the
redhatrelease detector is not compiled in, then that would not be a
fallback that the osrelease detector could rely on. :-\

Until https://github.com/coreos/clair/pull/193 is merged, having
vulnerabilities that are tagged both rhel and centos would duplicate in
the database or use a change that requires a migration.

But presently due to the fetcher logic, the rhel provided
vulnerabilities are labelled for centos, and then the namespace does not
match and therefore not tested against.

So until such a day that a vulnerability could have both rhel and centos
label, then hack this in. It'll accomplish the same during this interim.

### DIFF
--- a/updater/fetchers/rhel/rhel.go
+++ b/updater/fetchers/rhel/rhel.go
@@ -291,6 +291,7 @@ func toFeatureVersions(criteria criteria) []database.FeatureVersion {
 		}
 
 		if osVersion >= firstConsideredRHEL {
+			// TODO(vbatts) this is where features need multiple labels ('centos' and 'rhel')
 			featureVersion.Feature.Namespace.Name = "centos" + ":" + strconv.Itoa(osVersion)
 		} else {
 			continue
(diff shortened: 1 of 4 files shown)


vbatts:
bump

jzelinskie:
LGTM. Is this causing enough of an issue that it should be backported into stable?

vbatts:
Oh that is an idea. Let me check

brianwcook:
@jzelinskie Any idea when the next release will occcur that would include this if it isn't backported?

jzelinskie:
@brianwcook realistically on the order of months. We need to spend some time to making sure we get the multiple-namespace support in properly and make a few other breaking changes to clean things up. Doing so requires a large amount of thought around the future of project's architecture and our maintainers currently have a lot on their plates with other projects.

Backporting should be a simple `git cherry-pick`, so I'm totally fine with getting useful changes like this into our current stable version.

brianwcook:
@jzelinskie in that case, then yes it would be great to have it backported to stable!

PR #233, opened 2016-08-22
this just wasted a decent chunk of my afternoon

heph:
@Quentin-M please review

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -17,11 +17,12 @@ clair:
   database:
     # Database driver
     type: pgsql
-    options:
-      # PostgreSQL Connection string
-      # https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
-      source:
 
+    # PostgreSQL Connection string
+    # https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
+    source:
+
+    options:
       # Number of elements kept in the cache
       # Values unlikely to change (e.g. namespaces) are cached in order to save prevent needless roundtrips to the database.
       cachesize: 16384


Quentin-M:
Hi,

Sorry to hear that.

You used the `config.example.yaml` present on `master` while using Clair `1.2.X`. Instead, you have to use the configuration example from the `release-1.2` branch or even better, from the exact tag of your image.

After due verification of the images, the proper format for `1.2.X` releases is:

```
clair:
  database:
    # http://www.postgresql.org/docs/9.4/static/libpq-connect.html
    source:

    # Number of elements kept in the cache
    # Values unlikely to change (e.g. namespaces) are cached in order to save prevent needless roundtrips to the database.
    cacheSize: 16384
```

While the configuration for builds of `master` looks like:

```
clair:
  database:
    # Database driver
    type: pgsql
    options:
      source:

      # Number of elements kept in the cache
      # Values unlikely to change (e.g. namespaces) are cached in order to save prevent needless roundtrips to the database.
      cachesize: 16384
```

The README.md file appropriately warns about the `master` branch being a development branch and also suggests to use the a v1.2.2 configuration file with the v1.2.2 image. Note however that the latest version is v1.2.3. If you'd like to use an image built on master, you may use `quay.io/repository/coreos/clair-git:master`.

jzelinskie:
There is definitely a need for more clear documentation. For now, the only sane setup documentation for Clair is a part of the [Quay Enterprise documentation](https://tectonic.com/quay-enterprise/docs/1.16.5/clair.html) which does not apply to all users of Clair.

We already have a warning at the top of the readme, that master can contain breaking changes. I'm going to make an issue for this topic.

Sorry for the confusion.

PR #247, opened 2016-09-22
Added tempdir parameter for analyze-local-images

zuBux:
While using `analyze-local-images` in Mac OSX i encountered a problem with the temporary folder which is defined by the $TEMPDIR environmental variable. I added a parameter so the user can define the temporary folder where the image will be stored. It defaults to `/tmp`


jzelinskie:
`analyze-local-images` is now deprecated and located in another directory.

PR #252, opened 2016-09-29
Verify database state with retry

ScorpioCPH:
As mentioned in README

`Docker Compose may start Clair before Postgres which will raise an error. If this error is raised, manually execute docker start clair_clair.` 

so I add retry in verify database to avoid exiting.

### DIFF
--- a/database/pgsql/pgsql.go
+++ b/database/pgsql/pgsql.go
@@ -156,10 +156,26 @@ func openDatabase(registrableComponentConfig config.RegistrableComponentConfig) 
 		return nil, fmt.Errorf("pgsql: could not open database: %v", err)
 	}
 
-	// Verify database state.
-	if err := pg.DB.Ping(); err != nil {
+	// Verify database state with retry.
+	ch := make(chan int, 1)
+	go func() {
+		for {
+			err := pg.DB.Ping()
+			if err == nil {
+				ch <- 1
+			} else {
+				log.Error("failed to verify database state, retry after 2 seconds")
+				time.Sleep(2 * time.Second)
+			}
+		}
+	}()
+
+	select {
+	case <-ch:
+		// verify database state ok, break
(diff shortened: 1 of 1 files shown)


ScorpioCPH:
@Quentin-M, please review my PR, thanks!

database/pgsql/pgsql.go
```diff
@@ -156,10 +156,26 @@ func openDatabase(registrableComponentConfig config.RegistrableComponentConfig)
...
+	// Verify database state with retry.
+	ch := make(chan int, 1)
+	go func() {
+		for {
+			err := pg.DB.Ping()
+			if err == nil {
+				ch <- 1
+			} else {
+				log.Error("failed to verify database state, retry after 2 seconds")
+				time.Sleep(2 * time.Second)
```
jzelinskie, line 168: this should be exponential backoff

database/pgsql/pgsql.go
```diff
@@ -156,10 +156,26 @@ func openDatabase(registrableComponentConfig config.RegistrableComponentConfig)
 		return nil, fmt.Errorf("pgsql: could not open database: %v", err)
 	}
 
-	// Verify database state.
-	if err := pg.DB.Ping(); err != nil {
+	// Verify database state with retry.
+	ch := make(chan int, 1)
```
jzelinskie, line 160: `make(chan struct{})`

jzelinskie (changes requested):
I think this is a fair idea, but I think that 60s is probably way too long and that we should be using exponential backoff. You can use `github.com/coreos/pkg/timeutil`'s ExpBackOff function for this.

ajgreenb:
@jzelinskie I think this PR can now be closed?

ScorpioCPH:
@ajgreenb Sure! Sorry for the late response.

PR #253, opened 2016-09-30
Add flag to notify when the db is done updating

AndreeaDora:
This uses a flag that will notify when the updating of the db is done so an image could be tested only after all the vulnerabilities are added to the db.

### DIFF
--- a/api/v1/routes.go
+++ b/api/v1/routes.go
@@ -25,6 +25,7 @@ import (
 	"github.com/julienschmidt/httprouter"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/coreos/clair/updater"
 	"github.com/coreos/clair/api/context"
 	"github.com/coreos/clair/database"
 	"github.com/coreos/clair/utils"
@@ -138,6 +139,15 @@ func postLayer(w http.ResponseWriter, r *http.Request, p httprouter.Params, ctx 
 }
 
 func getLayer(w http.ResponseWriter, r *http.Request, p httprouter.Params, ctx *context.RouteContext) (string, int) {
+
+	log.Info(updater.DbFinishUpdate)
+
+	if updater.DbFinishUpdate {
+		writeResponse(w, r, http.StatusNotFound, LayerEnvelope{Error: & Error{"DB is being updated so no query to it can be done"}})
+		log.Errorf("DB is being updated so no query to it can be done")
+		return getLayerRoute, http.StatusNotFound
+	}
+
 	_, withFeatures := r.URL.Query()["features"]
 	_, withVulnerabilities := r.URL.Query()["vulnerabilities"]
 
--- a/api/v1/routes.go
+++ b/api/v1/routes.go
@@ -143,9 +143,9 @@ func getLayer(w http.ResponseWriter, r *http.Request, p httprouter.Params, ctx *
 	log.Info(updater.DbFinishUpdate)
 
 	if updater.DbFinishUpdate {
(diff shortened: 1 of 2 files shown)


soninob:
👍 
Can this be merged?
It is important to be notified when the update is done.
Looking forward seeing it merge :-)

updater/updater.go
```diff
@@ -133,6 +135,9 @@ func Run(config *config.UpdaterConfig, datastore database.Datastore, st *utils.S
 					nextUpdate = lockExpiration
 				}
 			}
+		} else {
+			DbFinishUpdate = false
```
Quentin-M: This is unfortunately far from enough to determine whether the initial update succeeded or not. First of all, your else block will be executed when it is not yet time to update and set the update status as unfinished, this does not really make sense. Also,
- Consider the case in which you have multiple Clair instances running and only one of them realizes the updates.
- Consider the case in which Clair is restarted.

Quentin-M, line 139: This is unfortunately far from enough to determine whether the initial update succeeded or not. First of all, your else block will be executed when it is not yet time to update and set the update status as unfinished, this does not really make sense. Also,
- Consider the case in which you have multiple Clair instances running and only one of them realizes the updates.
- Consider the case in which Clair is restarted.

Quentin-M (changes requested):

updater/updater.go
AndreeaDora: @Quentin-M Thank you for your feedback. This code determines if the update finished or not. As you can see in the logs the variable is set to false when the DB update is completed. If Clair is restarted same checks will be done. I will consider the case where we have multiple Clair instances running and only one of them  realizes the updates.

api/v1/routes.go
```diff
@@ -138,6 +139,15 @@ func postLayer(w http.ResponseWriter, r *http.Request, p httprouter.Params, ctx
 }
 
 func getLayer(w http.ResponseWriter, r *http.Request, p httprouter.Params, ctx *context.RouteContext) (string, int) {
+
+	log.Info(updater.DbFinishUpdate)
+
+	if updater.DbFinishUpdate {
+		writeResponse(w, r, http.StatusNotFound, LayerEnvelope{Error: & Error{"DB is being updated so no query to it can be done"}})
```
Quentin-M: The message could be clearer. Something like: `The initial vulnerability update has not finished yet, therefore this API call is currently unavailable. Please try again later`. Also, the error code should probably be 503.

updater/updater.go
Quentin-M: @AndreeaDora You may make use of the available `firstUpdate` variable. While not entirely safe (the updater doesn't keep its distributed state in a specialized service such as etcd), it is better already.

api/v1/routes.go
AndreeaDora: @Quentin-M I added the changes that you requested on the error code and message.

updater/updater.go
AndreeaDora: @Quentin-M The `firstUpdate` variable can not be used to determine when the update is done, that is the reason I used another variable that will change the value depending on the stage of the update.

tobilg:
What's the status on this? Thanks!

yixingjia:
This is a very valuable feature that greatly improve the usability.

It will be great if this can be include in Clair earlier. 

Our project(https://github.com/vmware/harbor) also integrate Clair  and we need this information when Vulnerability db has been updated.

soninob:
any updates on this?

jzelinskie:
The v3 API now contains a timestamp for the last update in its responses. I don't think this PR is necessary anymore.

Thanks for all the work.

PR #256, opened 2016-10-24
skip TLS check for detecting layer data

### LINKED ISSUE #251: x509 certificate error when detect content through docker registry blob api
I use docker registry blob api to let Clair to get layer content from the docker registry, but it has x509 certificate error. The docker registry is configured with TLS, docker client can skip TLS check by supporting "--insecure-registry" option. The TLS check of detecting data can be skipped?

**Error log**
2016-09-29 03:22:09.276447 I | notifier: notifier service is disabled
2016-09-29 03:23:22.556815 W | detectors: could not download layer: Get https://myregistry.vclound.com/v2/centos/blobs/sha256:a4de6e508363cd7e97c85ca458a2f61e949825793b1cfd3f5b72dde575eafd56: x509: certificate signed by unknown authority
2016-09-29 03:23:22.556864 E | worker: layer a4de6e508363cd7e97c85ca458a2f61e949825793b1cfd3f5b72dde575eafd56: failed to extract data from https://myregistry.vclound.com/v2/centos/blobs/sha256:a4de6e508363cd7e97c85ca458a2f61e949825793b1cfd3f5b72dde575eafd56: could not find layer
2016-09-29 03:23:22.557631 I | api: 10.101.136.149:62770 "POST /v1/layers" 400 (86.073213ms)

supereagle:
This skips TLS check for detecting layer data from Server 
configured with TLS. 

Fixes #251

### DIFF
--- a/cmd/clair/main.go
+++ b/cmd/clair/main.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/coreos/clair"
 	"github.com/coreos/clair/config"
+	"github.com/coreos/clair/worker/detectors"
 
 	// Register components
 	_ "github.com/coreos/clair/notifier/notifiers"
@@ -57,6 +58,7 @@ func main() {
 	flagConfigPath := flag.String("config", "/etc/clair/config.yaml", "Load configuration from the specified file.")
 	flagCPUProfilePath := flag.String("cpu-profile", "", "Write a CPU profile to the specified file before exiting.")
 	flagLogLevel := flag.String("log-level", "info", "Define the logging level.")
+	flagInsecureTLS := flag.Bool("insecure-tls", false, "Disable TLS check when detect the data of layers.")
 	flag.Parse()
 	// Load configuration
 	config, err := config.Load(*flagConfigPath)
@@ -72,6 +74,11 @@ func main() {
 	// Enable CPU Profiling if specified
 	if *flagCPUProfilePath != "" {
 		defer stopCPUProfiling(startCPUProfiling(*flagCPUProfilePath))
+	}
+
+	// Disable TLS check when detect data if specified
+	if *flagInsecureTLS {
+		detectors.SetInsecureTLS(*flagInsecureTLS)
 	}
 
 	clair.Boot(config)
--- a/worker/detectors/data.go
+++ b/worker/detectors/data.go
@@ -17,6 +17,7 @@
 package detectors
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io"
 	"math"
@@ -87,7 +88,11 @@ func DetectData(format, path string, headers map[string]string, toExtract []stri
 		}
 
 		// Send the request and handle the response.
-		r, err := http.DefaultClient.Do(request)
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		client := &http.Client{Transport: tr}
+		r, err := client.Do(request)
 		if err != nil {
 			log.Warningf("could not download layer: %s", err)
 			return nil, ErrCouldNotFindLayer
--- a/worker/detectors/data.go
+++ b/worker/detectors/data.go
@@ -46,6 +46,10 @@ var (
 
 	// ErrCouldNotFindLayer is returned when we could not download or open the layer file.
 	ErrCouldNotFindLayer = cerrors.NewBadRequestError("could not find layer")
+
+	insecureTLSLock sync.Mutex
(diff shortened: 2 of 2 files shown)


supereagle:
@Quentin-M  PTAL. Thanks!

jzelinskie:
Disabling these checks should be behind a flag if we're going to support it because defaulting to this behavior is dangerous. You _should_ be adding your certificates to the system root (and mounting that into a container).

supereagle:
@jzelinskie   Thanks for your review and advice.

Yes. Users can avoid this error by adding the certificates. But if the layer data is from several different servers, this way maybe is not so convenient. At the same time, it will restrict the container to run anywhere.

Adding a flag to control TLS check is a good way. I think we can add a flag `Insecure` in the body of `POST Layer` API, as TLS check is only used for detecting layer data in this REST API. Make the users can dynamically control the TLS check by themselves when call this API, and the default is to enable TLS check.

supereagle:
@jzelinskie  @Quentin-M  Whether we need to add a flag to support disabling this check?

jzelinskie:
@supereagle I would not be opposed to adding a flag to the binary for testing called `--insecure-tls` that disabled this across the board.

supereagle:
@jzelinskie  Is the flag `--insecure-tls` just for testing? Why not for production?

jzelinskie:
@supereagle For TLS to be secure, it needs to use PKI rather than blindly trusting any certificate it sees. People should be able to run Clair without verifying certificates, but if they want run it in that mode, they should be forced to acknowledge that it's insecure.

supereagle:
@jzelinskie   Thanks.  I will add this flag.

supereagle:
@jzelinskie  Have added the flag `--insecure-tls`.  PTAL, thanks.

supereagle:
@jzelinskie  Any comments about the new work in this PR?

jzelinskie:
I'm not 100% convinced that this API needs enable dynamic toggling of functionality. thoughts @Quentin-M ?

jzelinskie:
This needs a rebase.
Also follow-up ping @Quentin-M for thoughts

cmd/clair/main.go
```diff
@@ -57,6 +58,7 @@ func main() {
 	flagConfigPath := flag.String("config", "/etc/clair/config.yaml", "Load configuration from the specified file.")
 	flagCPUProfilePath := flag.String("cpu-profile", "", "Write a CPU profile to the specified file before exiting.")
 	flagLogLevel := flag.String("log-level", "info", "Define the logging level.")
+	flagInsecureTLS := flag.Bool("insecure-tls", false, "Disable TLS check when detect the data of layers.")
```
Quentin-M, line 61: > Disable TLS server's certificate chain and host name verification when pulling layers

worker/detectors/data.go
```diff
@@ -45,6 +46,10 @@ var (
 
 	// ErrCouldNotFindLayer is returned when we could not download or open the layer file.
 	ErrCouldNotFindLayer = cerrors.NewBadRequestError("could not find layer")
+
+	insecureTLSLock sync.Mutex
```
Quentin-M, line 50: I feel like the mutex is overkill.

worker/detectors/data.go
```diff
@@ -45,6 +46,10 @@ var (
...
+	// insecureTLS controls the TLS check when detect the data of layers, enabled in default.
```
Quentin-M, line 51: > insecureTLS controls whether TLS server's certificate chain and host name are verified when pulling layers

worker/detectors/data.go
```diff
@@ -120,3 +129,10 @@ func DetectData(format, path string, headers map[string]string, toExtract []stri
 
 	return nil, cerrors.NewBadRequestError(fmt.Sprintf("unsupported image format '%s'", format))
 }
+
+// SetInsecureTLS sets the insecureTLS to control the TLS check when detect the data of layers.
+func SetInsecureTLS(insecure bool) {
```
Quentin-M, line 134: I would not bother having a function if there is no mutex.

Quentin-M (changes requested):

Quentin-M:
I understand the use-case and would be willing to merge this.
I don't feel like having an API-level toggle is worth at all though, flag is fine, I even thought for a second that a const overwritten at build time would be enough - but other users might actually want to use this.

worker/detectors/data.go
supereagle: OK, I will remove it.

supereagle:
Too many conflicts while rebasing.  Create another PR #331.  This PR can be closed.

PR #257, opened 2016-10-25
Add Ubuntu yakkety to the namespace mapping.

mattmoor:
Unknown Ubuntu mappings cause the revision number to not get [updated](https://github.com/coreos/clair/blob/051564facd852ef5bbb62db9625631a21f2199d5/updater/fetchers/ubuntu/ubuntu.go#L142), and yakkety has started appearing in the feed.  Add the appropriate mapping to restore incrementality.


Quentin-M:
Thank you! LGTM.

Quentin-M:
@jzelinskie

PR #261, opened 2016-11-11
Replace liamstask/goose by remind101/migrate

### LINKED ISSUE #93: Improve database migration strategy

Quentin-M:
This PR replaces our current migrator tool for PostgreSQL (liamstack/goose), which is an abandoned project, by a simpler one.

- Migrations can now be embedded, 
- An [advisory lock](https://www.postgresql.org/docs/9.4/static/explicit-locking.html) is maintained on the database during the migration process,
- The initial database migration is now the second migration, instead of the first one. The first one verifies if there's a goose's table present in the database, if that is the case,it is removed the initial database migration is marked as completed for the new migrator.

Fixes #93

> +997 −2,224

What could go wrong?


jzelinskie:
> What could go wrong?

The CI could fail XD

Quentin-M:
Calculated!

jzelinskie (approved):
yay

PR #271, opened 2016-11-18
updater: Set vulns' Severity from NVD metadata fetcher if unknown 

Quentin-M:
If a Vulnerability that goes through the NVD metadata fetcher has an
empty or Unknown Severity, then use the CVSS score to set one. This
will help to get a more consistent database when a vulnerability source
does not provide this information.

### DIFF
--- a/updater/metadata_fetchers/nvd/nvd.go
+++ b/updater/metadata_fetchers/nvd/nvd.go
@@ -105,7 +105,6 @@ func (fetcher *NVDMetadataFetcher) AddMetadata(vulnerability *updater.Vulnerabil
 
 	if nvdMetadata, ok := fetcher.metadata[vulnerability.Name]; ok {
 		vulnerability.Lock.Lock()
-		defer vulnerability.Lock.Unlock()
 
 		// Create Metadata map if necessary.
 		if vulnerability.Metadata == nil {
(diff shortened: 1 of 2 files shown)


updater/metadata_fetchers/nvd/nvd.go
```diff
@@ -224,3 +230,22 @@ func getHashFromMetaURL(metaURL string) (string, error) {
 
 	return "", errors.New("invalid .meta file format")
 }
+
+// scoreToPriority converts the CVSS Score (0.0 - 10.0) into the user-friendy
+// types.Priority using the usually reference correlation table, extended
+// for the Negligible level.
+func scoreToPriority(score float64) types.Priority {
```
mattmoor, line 239: Worth mentioning that this follows the specification?
  https://www.first.org/cvss/specification-document

See table 14

Quentin-M: Yeah, thanks. That's what I used indirectly but didn't specify any link. Let me edit the comment. I still believe that `< 1.0` is better than `= 0.0` for the _Negligible_ Severity as 0.0 is virtually non-existent. If you prefer keeping `= 0.0`, I'm open to that.

jzelinskie (approved):

PR #272, opened 2016-11-18
[WIP] Alpine support via Alpine-SecDB 

jzelinskie:
I'm making this PR now just to let people know that it's finally happening. Data is sourced from [Alpine SecDB](http://git.alpinelinux.org/cgit/alpine-secdb/) (MIT licensed). This fetcher adds `git` as a dependency on the system path.

- [x] Namespace Detector based on `/etc/alpine-release` 
- [x] Feature detector based on `/lib/apk/db/installed`
- [x] Fetcher using alpine-secdb via git
- [x] Unit tests for all
- [x] Ensure everything actually works 😜


Quentin-M:
- [x] README (source, license)
- [x] Dockerfile (git dep)
- [ ] GIF centering (swag)

barakmich:
> GIF centering

:trolleybus:

betawaffle:
Umm, the GIF **doesn't even load!**

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,177 @@
...
+	// Pull the master branch.
+	var commit string
+	commit, err = f.pullRepository()
+	if err != nil {
+		return
+	}
+
+	// Ask the database for the latest commit we successfully applied.
+	var dbCommit string
+	dbCommit, err = db.GetKeyValue("alpineUpdater")
```
Quentin-M: const

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return
+	}
+
+	// Append any changed vulnerabilities to the response.
+	if commit == dbCommit {
+		log.Debug("no alpine update")
+	} else {
```
Quentin-M: set FlagName/FlagValue before; then no need for else, just return; saves an indentation level.

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,177 @@
...
+	err = yaml.Unmarshal(rBytes, &file)
+	if err != nil {
+		return
+	}
+	for _, pack := range file.Packages {
+		pkg := pack.Pkg
+		for _, fix := range pkg.Fixes {
+			version, err := types.NewVersion(pkg.Version)
+			if err != nil {
+				continue
```
Quentin-M, line 225:
```
log.Warningf("could not parse package version '%s': %s. skipping", pkg.version, err.Error())
```

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+
+			var vuln database.Vulnerability
```
Quentin-M: should we use the `database.Vulnerability{}` notation?

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,177 @@
...
+			vuln.Name = fix
+			vuln.Link = nvdURLPrefix + fix
+			vuln.FixedIn = append(vuln.FixedIn, database.FeatureVersion{
```
Quentin-M: why `append()` if you just created the object?

worker/detectors/feature/apk/apk.go
```diff
@@ -0,0 +1,84 @@
...
+				if err != nil {
+					log.Warningf("could not parse package version '%s': %s. skipping", line[2:], err.Error())
+				}
+			}
+
+			// If we have a whole feature, store it in the set and try to parse a new
+			// one.
+			if ipkg.Feature.Name != "" && ipkg.Version.String() != "" {
+				pkgSet[ipkg.Feature.Name+"#"+ipkg.Version.String()] = ipkg
+				ipkg = database.FeatureVersion{}
```
Quentin-M: TODO?

worker/detectors/feature/apk/apk.go
```diff
@@ -0,0 +1,84 @@
...
+
+func init() {
+	detectors.RegisterFeaturesDetector("apk", &detector{})
+}
+
+type detector struct{}
+
+func (d *detector) Detect(data map[string][]byte) ([]database.FeatureVersion, error) {
+	file, exists := data["lib/apk/db/installed"]
+	if exists {
```
Quentin-M:
```
    f, hasFile := data["lib/apk/db/installed"]
    if !hasFile {
        return []database.FeatureVersion{}, nil
    }
```

simpler, avoids one level of indentation

mattmoor:
Instead of adding a dependency on Git, did you consider simply fetching the tarball at HEAD?

```
http://git.alpinelinux.org/cgit/alpine-secdb/snapshot/alpine-secdb-master.tar.bz2
```

Quentin-M:
It makes a lot of sense now because their database is extremely small. As the database gets bigger (hopefully.. but maybe never big enough to make sense), `git pull` will become faster and download less data after few update loops. However using the tarball would surely make adapting the fetcher easier when we'll want to adapt Clair to be used without external network connectivity.

jzelinskie:
> However using the tarball would surely make adapting the fetcher easier when we'll want to adapt Clair to be used without external network connectivity.

I'm absolutely open to changing to the tarball download if this a goal for the future. Do we have a plan for migrating existing fetchers to systems that could support this?

mattmoor:
FWIW, we run in contexts where shelling out to tools like `bzr`, `git`, `rpm` simply isn't practical.  In our pseudo air-gapped environment we can reach outside by swapping out `http.DefaultTransport` in-process for certain sites, but tool invocations create trouble (harder to do the same, especially without carrying patches).

worker/detectors/feature/apk/apk.go
jzelinskie: ipkg = intermediary package, once you have both the name and version, we reset it so that we can parse a new one.

jzelinskie:
@mattmoor Have you been unable to use Ubuntu fetchers for this same reason? They depend on bzr.

mattmoor:
@jzelinskie Yes, we have a variant (that we were supposed to upstream, which I'm running down) that basically makes the mechanism pluggable.

The tl;dr is that we define an interface for the core methods that shell out to `bzr` and implement it one way with `bzr` and another with the `bzr` tarball url.

mwarkentin:
Does anyone have a feel for how maintained [alpine-secdb](http://git.alpinelinux.org/cgit/alpine-secdb/) is? It looks like it was last updated ~7 weeks ago?

Really hoping for a good solution here for monitoring for vulnerable packages in Alpine, just wondering if there are any plans for how the upstream database will be kept up to date?

Quentin-M:
@mwarkentin That's a very good question. We have contacted the main committer to ask about the license and whether we should expect this database to be a first-time citizen / updated regularly or not. We now know that the database is licensed under MIT but as far as I know, we have had no update on that second question yet. We hope that our implementation will increase demand on this important topic at Alpine.

mwarkentin:
@Quentin-M Thanks! I agree that having an implementation like this could be a good driver for keeping the db updated.

Please update if you get any further information about this. :)

jzelinskie:
@mattmoor any timeline on upstreaming this work? If we could settle on a streamlined a process for backchanneling data sources for air-gapped Clair instances, that'd be great!

mattmoor:
FWIW, I wouldn't characterize it as a general air-gap solution so much as a pattern that enables it.  No timeframe yet, so don't block this for it.

Quentin-M:
@jzelinskie Have any e2e tested it on an image that actually contains one of the managed vulnerability? Also, I am aware that an existing namespace detector catched Alpine (probably wrongly) on some images, that'd be good to make sure they don't collide and that only the actual new Alpine detector gets it rather than the other one.

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,247 @@
...
+
+	"github.com/coreos/clair/database"
+	"github.com/coreos/clair/updater"
+	"github.com/coreos/clair/utils"
+	cerrors "github.com/coreos/clair/utils/errors"
+	"github.com/coreos/clair/utils/types"
+)
+
+const (
+	secdbGitURL  = "http://git.alpinelinux.org/cgit/alpine-secdb"
```
Quentin-M, line 40: Can we add a note for 'httpss://' when supported?

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,247 @@
...
+	resp.FlagValue = commit
+
+	// Short-circuit if there have been no updates.
+	if commit == dbCommit {
+		log.Debug("no alpine update")
+		return
+	}
+
+	// Append any changed vulnerabilities to the response.
+	for _, namespace := range []string{"v3.3", "v3.4"} {
```
Quentin-M: nit: It'd be neat to extract the version->parser mapping outside of this function, at the top.

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+
+		if out, err := utils.Exec(f.repositoryLocalPath, "git", "clone", secdbGitURL, "."); err != nil {
+			f.Clean()
+			log.Errorf("could not pull alpine-secdb repository: %s. output: %s", err, out)
+			return "", cerrors.ErrCouldNotDownload
+		}
+	}
+
+	out, err := utils.Exec(f.repositoryLocalPath, "git", "rev-parse", "HEAD")
```
Quentin-M, line 181: So the first time, we `clone`. The next times, we just `rev-parse HEAD`. Shouldn't we pull before? Otherwise even if `HEAD` moved, we'd still be reading the old files when opening them.

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,247 @@
...
+	for _, pack := range file.Packages {
+		pkg := pack.Pkg
+		for _, fix := range pkg.Fixes {
+			version, err := types.NewVersion(pkg.Version)
+			if err != nil {
+				log.Warningf("could not parse package version '%s': %s. skipping", pkg.Version, err.Error())
+				continue
+			}
+
+			var vuln database.Vulnerability
```
Quentin-M: Is not using `database.Vulnerability{Name: "", Link: ""}` style a conscious decision?

Quentin-M:
Besides my comment regarding to potential Alpine detections with the existing detectors above, this looks fine.

I think my idea around `range []string{"v3.3", "v3.4"}` was to have a map `Version->Parsing function` at the top and avoid having these hardcoded in a function (in which we range and switch over). So we could simply modify the map and add a new parsing function whenever we need to. Now they are hardcoded in two functions (`parseVulnsFromNamespace` and `FetchUpdate`). It doesn't matter too much though, just style.

Quentin-M (approved):

updater/fetchers/alpine/alpine.go
```diff
@@ -107,13 +143,12 @@ func parseVulnsFromNamespace(repositoryPath, namespace string) (vulns []database
...
 	defer file.Close()
 
-	switch namespace {
-	case "v3.3":
-		vulns, err = parse33YAML(file)
-	case "v3.4":
-		vulns, err = parse34YAML(file)
+	parseFunc, exists := parsers[namespace]
+	if !exists {
+		return
```
Quentin-M: All these are much better. Please just add a Warning here as well as an updater Note for that so we can be alerted when we need to take action. See https://github.com/coreos/clair/blob/master/updater/fetchers/ubuntu/ubuntu.go#L139

updater/fetchers/alpine/alpine.go
```diff
@@ -0,0 +1,293 @@
...
+	var file io.ReadCloser
+	file, err = os.Open(repositoryPath + "/" + namespace + "/main.yaml")
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	parseFunc, exists := parsers[namespace]
+	if !exists {
+		return
```
Quentin-M, line 154: All these are much better. Please just add a Warning here as well as an updater Note for that so we can be alerted when we need to take action. See https://github.com/coreos/clair/blob/master/updater/fetchers/ubuntu/ubuntu.go#L139

Quentin-M:
LGTM. Users will then be able to figure out action is needed by reading Prometheus.

adityacs:
@jzelinskie Could you please provide me a alpine docker image which has some CVEs? 
I scanned many images with alpine but always getting 0 vulnerabilities.

boydgreenfield:
@adityacs Very late, but `node:6.9.4-alpine` has a bunch of vulnerabilities (first one I googled, was also worried as the latest few all returned 0 and I wasn't sure if any real checks were being executed):

```
2018/01/25 18:44:33 [WARN] ▶ Image [node:6.9.4-alpine] contains 7 total vulnerabilities
2018/01/25 18:44:33 [ERRO] ▶ Image [node:6.9.4-alpine] contains 7 unapproved vulnerabilities
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| STATUS     | CVE SEVERITY          | PACKAGE NAME | PACKAGE VERSION | CVE DESCRIPTION                                               |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| Unapproved | High CVE-2016-9843    | zlib         | 1.2.8-r2        |                                                               |
|            |                       |              |                 | https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9843  |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| Unapproved | High CVE-2016-9841    | zlib         | 1.2.8-r2        |                                                               |
|            |                       |              |                 | https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9841  |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| Unapproved | Medium CVE-2017-16544 | busybox      | 1.24.2-r12      |                                                               |
|            |                       |              |                 | https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-16544 |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| Unapproved | Medium CVE-2017-15650 | musl         | 1.1.14-r14      |                                                               |
|            |                       |              |                 | https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-15650 |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| Unapproved | Medium CVE-2016-9842  | zlib         | 1.2.8-r2        |                                                               |
|            |                       |              |                 | https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9842  |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| Unapproved | Medium CVE-2016-9840  | zlib         | 1.2.8-r2        |                                                               |
|            |                       |              |                 | https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9840  |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
| Unapproved | Medium CVE-2017-15873 | busybox      | 1.24.2-r12      |                                                               |
|            |                       |              |                 | https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-15873 |
+------------+-----------------------+--------------+-----------------+---------------------------------------------------------------+
```

PR #277, opened 2016-12-04
travis: add matrix for postgres

jzelinskie:
This updates our CI to test Postgres 9.4, 9.5, and 9.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,34 @@
 language: go
-
-go:
-  - 1.7
-  - tip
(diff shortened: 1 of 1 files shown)


Quentin-M:
Removed Go versions?

jzelinskie:
Testing against tip of Go isn't super useful unless we were maybe benchmarking things. It also makes the matrix very large 2*3

Quentin-M (approved):

PR #278, opened 2016-12-04
pgsql/migrations: add ldfv compound index

jzelinskie:
This speeds up the SearchNotificationLayerIntroducingVulnerability query
by an order magnitude.

After adding this index and vacuuming the table, the following query changes from `570578.327 ms` to `1239.984 ms` execution time:

```sql
EXPLAIN ANALYZE WITH subquery AS (
  SELECT l.ID, l.name
  FROM Vulnerability_Affects_FeatureVersion vafv, FeatureVersion fv, Layer_diff_FeatureVersion ldfv, Layer l
  WHERE l.id >= $2
    AND vafv.vulnerability_id = $1
    AND vafv.featureversion_id = fv.id
    AND ldfv.featureversion_id = fv.id
    AND ldfv.modification = 'add'
    AND ldfv.layer_id = l.id
  ORDER BY l.ID
)
SELECT *
FROM subquery
LIMIT $3;
```

### DIFF
--- a/database/pgsql/migrations/00004_ldfv_index.go
+++ b/database/pgsql/migrations/00004_ldfv_index.go
@@ -1,0 +1,29 @@
+// Copyright 2016 clair authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


Quentin-M:
LGTM. 

Actually, I noticed yesterday that the query planner was taking bad decisions such as doing seq scans  rather than index scans at few places where the index was present. It must have been due to the table not being properly analyzed, with PostgreSQL considering that the query will return > 5-10% of the total table.

Verified performance gain with the index and analyzed table.

Quentin-M (approved):

PR #287, opened 2016-12-19
worker: bump engine version

jzelinskie:
Now that we support OpenSUSE and Alpine Linux the engine version should
be increased.


Quentin-M (approved):

Quentin-M:
GL HF

PR #288, opened 2016-12-19
worker: clarify maxFileSize purpose

### LINKED ISSUE #237: Failure to scan layers over 200MB
Due to [this line](https://github.com/coreos/clair/blob/28fbfc72e7a77f15fffc3848ecbb44943dd15a70/worker/worker.go#L34), we cannot scan layers larger than 200MB. This value should probably be configurable and not hard-coded.

jzelinskie:
Fixes #237.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -30,7 +30,9 @@ const (
 	// Increased each time the engine changes.
 	Version = 3
 
-	// maxFileSize is the maximum size of a single file we should extract.
+	// maxFileSize enforces a maximum size of a single file within a tarball that
+	// will be extracted. This protects against malicious layers that may contain
+	// extremely large package database files.
 	maxFileSize = 200 * 1024 * 1024 // 200 MiB
 )
 


worker/worker.go
```diff
@@ -30,7 +30,9 @@ const (
 	// Increased each time the engine changes.
 	Version = 3
 
-	// maxFileSize is the maximum size of a single file we should extract.
+	// maxFileSize enforces a maximum size of a single file within a tarball that
+	// will be extracted. This protects against against malicious layers that may
```
Quentin-M: against against

Quentin-M: ^

Quentin-M (approved):

alekssaul:
@jzelinskie : any chance we can make it a variable in config.yaml? I believe at least 2 QE customers are interested in scanning large layers.

Quentin-M:
@alekssaul As the comment in the code and as my [comment](https://github.com/coreos/clair/issues/237#issuecomment-242172914) in the initial issue state, this value has nothing to do with the size of the layer Clair can analyze but with the size of the targeted files we extract to parse informations from (e.g. /etc/os-release, /var/lib/dpkg, etc). These files today have literally no reason to exceed 200 MiB except clear abuse attempt. Clair can analyze layers of any size.

PR #299, opened 2016-12-29
fix error when scanning folders for feteched update for Alpine vulnerabilities

alexei-led:
Fix error `updater: an error occured when fetching update 'alpine': open /private/tmp/alpine-secdb808866024/Makefile/main.yaml: not a directory`.

This error happens on MacOS with `go version go1.7.4 darwin/amd64`.

For some strange reason `File.Readdirnames` returns names of directories and **files**. I've changed this code to use `File.Readdir` that returns `[]Fileinfo` and thus adding additional validation that skips non-directories, using `IsDir()` method.

### DIFF
--- a/updater/fetchers/alpine/alpine.go
+++ b/updater/fetchers/alpine/alpine.go
@@ -115,19 +115,22 @@ func detectNamespaces(path string) ([]string, error) {
 	defer dir.Close()
 
 	// Get a list of the namspaces from the directory names.
-	names, err := dir.Readdirnames(0)
+	finfos, err := dir.Readdir(0)
 	if err != nil {
 		return nil, err
 	}
 
 	var namespaces []string
-	for _, name := range names {
+	for _, info := range finfos {
+		if !info.IsDir() {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


jzelinskie:
Thanks for the PR! Strange corner-case indeed.

Your commit seems to also make unrelated changes to the vendor directory, though.

alexei-led:
changes to vendor folder are not related indeed. I just run "glide install" to build and test clair and glide updated couple files in vendor folder. Ignore these, sorry.

jzelinskie:
If these changes are indeed pulled in by `glide install`, but are not currently apart of the vendor directory, we should commit them. However, they should be in a separate commit. Likewise if you'd like to add a .gitignore file, please put it in a separate commit. Having individual commits like this helps in the future when you need to use tools like `git blame` to remember why a specific change was made.

alexei-led:
OK, fixed this PR to include only one change, that fixes described issue

jzelinskie (approved):
LGTM. Thanks for adjusting the PR!

PR #307, opened 2017-01-19
pgsql: searchNotificationLayerIntroducingVulnerability order by layer ID

jzelinskie:
This fixes a bug where the API was returning Notification pages ordered
by LDFV.ID instead of by Layer ID.


Quentin-M (approved):
LGTM.

I can't recall what the ordering was based on ldfv.ID rather than ldfv.layer_id. Maybe there was an actual reason for it, or maybe it represents an actual bug. The change doesn't seem wrong to me anyways.

Let's see how it behaves both in term of performance and consistency.

PR #308, opened 2017-01-20
pgsql: only select distinct layers for LayersIntroducingVulnerability

jzelinskie:
This removes duplicate Layers from the pagination over GetNotification.

Previously we would show the same Layer multiple times if it had multiple FeatureVersions introducing the the same vulnerability. This was unintuitive to the consumer of the API because the API doesn't return the FeatureVersion, so it just appeared to randomly be returning duplicates.

This should also make the query slightly faster in the cases I tested.

### DIFF
--- a/database/pgsql/queries.go
+++ b/database/pgsql/queries.go
@@ -212,7 +212,7 @@ const (
 
 	searchNotificationLayerIntroducingVulnerability = `
 		WITH LDFV AS (
-		  SELECT ldfv.layer_id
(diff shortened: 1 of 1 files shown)


Quentin-M:
FWIW The duplication was desired behavior to notify the users that even though they are already affected by a vulnerability, there is now a new way the vulnerability is introduced in their layers (and vice-versa). Your point that the output might be confusing is legitimate and I think that this is a corner case anyways.

Quentin-M (approved):

PR #309, opened 2017-01-20
pgsql/migrations: fix dpkg default versionfmt

jzelinskie:


Quentin-M (approved):

Quentin-M:
This is fine because it's the master branch, expected to be broken but eventually we'll have to create a new migration for these if any got released.

PR #312, opened 2017-01-27
Relax the version's leading digit check.

mattmoor:
This change relaxes a check in `types.NewVersion` that requires the middle segment of a version string to start with a leading digit.

From the relevant `upstream_version` documentation (link above `NewVersion`, **emphasis** mine): "and **should** start with a digit".

You can see that Alpine deviates from this [here](http://git.alpinelinux.org/cgit/alpine-secdb/tree/v3.3/main.yaml#n72).  This isn't breaking the Clair Alpine ingester (at least yet) because the v3.3 format changed from its own to what the v3.4 file is using, which means Clair currently isn't seeing new vulnerabilities in the v3.3 feed.  While investigating this, I also noticed that there are two additional changes that Clair is missing support for: v3.5 and community.yaml

### DIFF
--- a/utils/types/version.go
+++ b/utils/types/version.go
@@ -94,10 +94,6 @@ func NewVersion(str string) (Version, error) {
 		return Version{}, errors.New("No version")
 	}
 
-	if !unicode.IsDigit(rune(version.version[0])) {
-		return Version{}, errors.New("version does not start with digit")
-	}
-
 	for i := 0; i < len(version.version); i = i + 1 {
 		r := rune(version.version[i])
 		if !unicode.IsDigit(r) && !unicode.IsLetter(r) && !containsRune(versionAllowedSymbols, r) {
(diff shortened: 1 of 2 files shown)


mattmoor:
@jzelinskie FYI

mattmoor:
Now with (hopefully) passing tests!

jzelinskie:
Thanks for the PR. However, `types.Version` isn't being used anymore. Rather, `ext/versionfmt/(dpkg|rpm)` is being used because rpm had version comparison rules that could not be reconciled with the dpkg algorithm. We actually just forgot to delete these files, but I did delete them in #305. Is there a formal specification for apk version strings? If so we can create an versionfmt for it that is independent from the other package types.

jzelinskie:
Hey, I fixed this in #316 as a part of adding support for actually parsing the new v3.5 stuff.

Thanks for the heads up.

PR #316, opened 2017-02-07
Alpine v3.5 support

jzelinskie:


config.example.yaml
```diff
@@ -20,7 +20,7 @@ clair:
     options:
       # PostgreSQL Connection string
       # https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
-      source:
+      source: host=localhost port=5432 user=postgres sslmode=disable
```
Quentin-M: We generally have `statement_timeout=60000` too

ext/vulnsrc/alpine/alpine.go
```diff
@@ -79,8 +79,13 @@ func (u *updater) Update(db database.Datastore) (resp vulnsrc.UpdateResponse, er
 		return
 	}
 
+	// Get the list of namespaces from the repository.
 	var namespaces []string
-	namespaces, err = detectNamespaces(u.repositoryLocalPath)
+	namespaces, err = ls(u.repositoryLocalPath, directoriesOnly)
```
Quentin-M, line 84: is `detectNamespaces` still used anywhere?

ext/vulnsrc/alpine/alpine.go
```diff
@@ -134,28 +139,65 @@ func detectNamespaces(path string) ([]string, error) {
...
-	parseFunc, exists := parsers[namespace]
-	if !exists {
-		note = fmt.Sprintf("The file %s is not mapped to any Alpine version number", namespace)
+	for _, filename := range dbFilenames {
+		var file io.ReadCloser
+		file, err = os.Open(filepath.Join(nsDir, filename))
+		if err != nil {
+			return
+		}
+		defer file.Close()
```
Quentin-M: please don't use `defer`, it will clog the fds, which won't be closed until the function ends. not like it matters too much considering the limited number of files in the Alpine's database but..

ext/vulnsrc/alpine/alpine.go
```diff
@@ -134,28 +139,65 @@ func detectNamespaces(path string) ([]string, error) {
...
 		return
 	}
-	defer file.Close()
 
-	parseFunc, exists := parsers[namespace]
-	if !exists {
-		note = fmt.Sprintf("The file %s is not mapped to any Alpine version number", namespace)
+	for _, filename := range dbFilenames {
+		var file io.ReadCloser
+		file, err = os.Open(filepath.Join(nsDir, filename))
```
Quentin-M: Is there a reason you don't use `file, err := os.Open(...` ?

ext/vulnsrc/alpine/alpine.go
```diff
@@ -134,28 +139,65 @@ func detectNamespaces(path string) ([]string, error) {
...
 	if err != nil {
 		return
 	}
-	defer file.Close()
 
-	parseFunc, exists := parsers[namespace]
-	if !exists {
-		note = fmt.Sprintf("The file %s is not mapped to any Alpine version number", namespace)
+	for _, filename := range dbFilenames {
+		var file io.ReadCloser
```
Quentin-M: kinda annoying that you have to declare every variables to avoid hiding err.. oops.

ext/vulnsrc/alpine/alpine.go
jzelinskie: nope

ext/vulnsrc/alpine/alpine.go
jzelinskie: It's this or `:=` everywhere with no named return values. Would you prefer that?

ext/vulnsrc/alpine/alpine.go
jzelinskie: Nice catch. You actually lead me to discover we were returning early here, too.

ecordell (approved):
LGTM

Quentin-M (approved):

PR #318, opened 2017-02-12
fix example config

### LINKED ISSUE #317: main: pgsql: no database connection string specified
Hi! I'm working with Docker version 1.13.0, quay.io/coreos/clair:v2.0.0-pre.3 (because I want to be able to scan images and Alpine as the base), and postgres:latest.  I followed all the instructions and set the clair's database to `source: postgresql://postgres:password@postgres:5432?sslmode=disable`. When use docker-compose to start postgres and clair, clair won't start up right and the database doesn't have any relations.  

clair_clair log:
`docker logs clair_clair`
`2017-02-09 21:07:32.638558 C | main: pgsql: no database connection string specified`  

When I try to list the tables, there's nothing:
`postgres=# \dt`
`No relations found.` 

What am I missing? Thanks!

jessfraz:
fixes #317

options is not a thing it seems..

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -17,14 +17,13 @@ clair:
   database:
     # Database driver
     type: pgsql
-    options:
-      # PostgreSQL Connection string
-      # https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
-      source: host=localhost port=5432 user=postgres sslmode=disable statement_timeout=60000
+    # PostgreSQL Connection string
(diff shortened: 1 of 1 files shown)


Quentin-M:
Hi,

I am afraid you tried to run Clair 1.2.x using the example configuration file checked on _master_, which doesn't support the same format. Could you verify using the example from the _release-1.2_ branch or use the latest Clair build (recommended)?

jessfraz:
ah gotcha

PR #327, opened 2017-02-23
Fix areas only copying namespace names

jzelinskie:
When performing the first vulnerability update, the namespaces being inserted for all the vulnerabilities are having their version format stripped.

This address this issue by ensuring that the whole namespace is copied when being reassigned.


Quentin-M (approved):
assuming that's the only place we do that

Quentin-M:
Thanks.

PR #328, opened 2017-02-23
Move Hyperclair link to Clairctl

jgsqware:
github.com/wemanity-belgium/hyperclair is deprecated.
the new tool is located there: github.com/jgsqware/clairctl and the development continue there too

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,6 @@ To expose the new behavior, unqualified imports to the package must be added in 
 
 - [Quay](https://quay.io): the first container registry to integrate with Clair
 - [Dockyard](https://github.com/containerops/dockyard): an open source container registry with Clair integration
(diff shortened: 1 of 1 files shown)


jzelinskie:
Are we still going to make an effort to merge repositories at some point or are we going to keep them separate? I'd really like to remove basically everything in the `contrib` directory because they aren't all complete solutions and vendor old versions of Clair inside Clair's own git repository.

jgsqware:
I would prefer finally to integrate it in Clair repo, like we have for
kubernetes and kubectl.
I had a lot of difficulties to reintegrate with Clair due to cross
dependencies and we probably need to discuss how to address that
especially.
The current version on clairctl works better so I can take time to try to
reintagrate it

Le jeu. 23 févr. 2017 19:23, Jimmy Zelinskie <notifications@github.com> a
écrit :

jzelinskie (approved):
Sounds good, I'll close the other PR

Issue #313, opened 2017-01-27
Add support for Alpine Linux 3.5


jzelinskie:
These vulnerabilities are now available in the database, but need to explicitly be added to the Clair codebase.

mattmoor:
It is notable that the v3.5/main.yaml is currently malformed (it has a few tabs).  I have reached out to the maintainer about fixing this.

jzelinskie:
Fixed by #316

PR #341, opened 2017-03-06
add Alpine 3.5 Data Source Readme

pizzarabe:
Support for Alpine 3.5 was fixed with https://github.com/coreos/clair/issues/313, but the Readme does not contain Alpine 3.5


jzelinskie (approved):
Looks good! Thanks!

PR #342, opened 2017-03-07
fix an error produced by analyze-local-images remotely

### LINKED ISSUE #336: analyze-local-images remotely produces an error
I have a clair running on a server and I'm trying to scan an image that is local on a different server but I'm getting an error. As an example, I've simply pulled the latest mysql and these are the logs:

2017-03-01 18:06:33.293256 I | Saving mysql:latest to local disk (this may take some time)
2017-03-01 18:06:38.454858 I | Retrieving image history
2017-03-01 18:06:38.455120 I | Setting up HTTP server (allowing: **********)
2017-03-01 18:06:38.555234 I | Analyzing 11 layers... 
2017-03-01 18:06:38.555255 I | Analyzing 5e080092760e09c12b8ef50fdce5affaee2c854ba09fc7257f795d61b6dac3f2
2017-03-01 18:06:39.327722 I | Could not analyze layer: Got response 400 with message {"Error":{"Message":"could not find layer"}}
exit status 1

ruokai-lai:
scripts: fix an error produced by analyze-local-images remotely

We are getting the error, Could not analyze layer: Got response 400 with message {"Error":{"Message":"could not find layer"}}, because the script only allow endpoint (from -endpoint) to access our local data. However, the logic only match ip address, which mean if you use anlayze-local-images -endpoint http://<dns-name>:6060 xxx, it will fails. The fix is to remove the restriction. 

Fixes #336

### DIFF
--- a/contrib/analyze-local-images/main.go
+++ b/contrib/analyze-local-images/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
@@ -172,7 +173,7 @@ func AnalyzeLocalImage(imageName string, minSeverity types.Priority, endpoint, m
 		log.Printf("Setting up HTTP server (allowing: %s)\n", allowedHost)
 
 		ch := make(chan error)
-		go listenHTTP(tmpPath, ch)
+		go listenHTTP(tmpPath, allowedHost, ch)
 		select {
 		case err := <-ch:
 			return fmt.Errorf("An error occured when starting HTTP server: %s", err)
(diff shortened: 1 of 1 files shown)


Quentin-M:
Hi,

Thanks for the contribution. I understand the purpose of this PR. `analyze-local-images` is a limited tool aimed at helping new comers. I am a bit worried that removing the restriction altogether represents a security issue (even though this tool is definitely not secure already, nor not meant to be) and encourage users to use that tool more broadly / remotely.

Maybe we could come with a better policy / parsing instead?

ruokai-lai:
hi @Quentin-M ,

Instead of removing the security check, I change the match logic.

jzelinskie:
This tool was removed from our repository, so this PR is no longer necessary.

PR #343, opened 2017-03-07
add in a sum for total errors for analyze-local-images. `echo $?` return 1 when there are any errors, …

ruokai-lai:
add in a sum for total errors for analyze-local-images. `echo $?` return 1 when there are any errors, so that it can break our CI pipeline when needed

### DIFF
--- a/contrib/analyze-local-images/main.go
+++ b/contrib/analyze-local-images/main.go
@@ -270,7 +270,10 @@ func AnalyzeLocalImage(imageName string, minSeverity types.Priority, endpoint, m
 		fmt.Printf("%s No vulnerabilities were detected in your image\n", color.GreenString("Success!"))
 	} else if !hasVisibleVulnerabilities {
 		fmt.Printf("%s No vulnerabilities matching the minimum severity level were detected in your image\n", color.YellowString("NOTE:"))
-	}
+	} else {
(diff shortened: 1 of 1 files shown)


contrib/analyze-local-images/main.go
```diff
@@ -270,7 +270,10 @@ func AnalyzeLocalImage(imageName string, minSeverity types.Priority, endpoint, m
 		fmt.Printf("%s No vulnerabilities were detected in your image\n", color.GreenString("Success!"))
 	} else if !hasVisibleVulnerabilities {
 		fmt.Printf("%s No vulnerabilities matching the minimum severity level were detected in your image\n", color.YellowString("NOTE:"))
-	}
+	} else {
+                errorTotal := errors.New("total Errors: "+strconv.Itoa(len(vulnerabilities)))
```
Quentin-M: Let's make it a sentence and then LGTM: 

```
return fmt.Errorf("A total of %d vulnerabilities have been detected in your image", len(vulnerabilities))
```

Quentin-M (changes requested):

Quentin-M:
Thanks for your contribution!

contrib/analyze-local-images/main.go
ruokai-lai: Like your suggestion. :+1:  have update my code.

Quentin-M (approved):

PR #348, opened 2017-03-16
update docker image spec url

supereagle:
Update docker image spec url as image spec v1.2.0 has been adopted starting in Docker 1.12.

@Quentin-M  @jzelinskie  PTAL.


jzelinskie (approved):
Looks good! Thanks for the fix!

Issue #221, opened 2016-07-21
Disable fetcher(s)?


Djelibeybi:
Is there a way to disable one/more fetchers when Clair runs? 

I'd like to disable the Debian/Ubuntu fetchers as they take a very long time to run and I'm not intending on scanning any images based on Debian or Ubuntu.

jzelinskie:
You cannot at runtime, but you can at compile time. If you strip the imports from main.go (i.e. remove [this line](https://github.com/coreos/clair/blob/master/cmd/clair/main.go#L31))

Djelibeybi:
True. I guess I was hoping for something I wouldn't have to keep reverting to make sure I don't check it into my git repo by mistake. :)

Can I turn this into a config enhancement request?

jzelinskie:
Addressed in #432.

PR #351, opened 2017-03-21
Enable runtime fetcher disable/enable

Djelibeybi:
Implements #221 

This adds a configuration section to config.yaml which allows users
to enable/disable each fetcher at runtime.

I used the same mechanism being used by the notifications so that future fetchers can add additional configuration items without too much work.

Signed-off-by: Avi Miller <avi.miller@oracle.com>

### DIFF
--- a/cmd/clair/config.go
+++ b/cmd/clair/config.go
@@ -25,6 +25,7 @@ import (
 	"github.com/coreos/clair"
 	"github.com/coreos/clair/api"
 	"github.com/coreos/clair/database"
+	"github.com/coreos/clair/ext/vulnsrc"
 	"github.com/coreos/clair/ext/notification"
 	"github.com/fernet/fernet-go"
 )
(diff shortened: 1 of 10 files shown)


Quentin-M:
Thanks,

I really like the idea, thanks for your contribution!

jzelinskie:
The code looks really good. @Quentin-M should we ship this after 2.0 for as a part of it? We're already breaking the config.

Djelibeybi:
The existing config (unmodified) should allow Clair to run without modification. The config changes can be added to disable fetchers, but without the config, all fetchers should work (as before).

jzelinskie:
This sounds good to me. Let's have them all default to on, but can be disabled.

jzelinskie:
#432 implements this behavior, but with a much safer design at the database level.
I'm going to close this PR and prioritize work towards getting #432 merged.

PR #352, opened 2017-04-03
api: fix 404 error logging

kevinburke:
The first value is an integer and the second value is a string and the
format string has these backwards, leading to %!s(int=404) appearing
in the logs instead of "404".


jzelinskie (approved):
Thanks! We've actually noticed this, but ignored it for far too long!

PR #366, opened 2017-04-19
ext/vulnsrc/oracle: fix ELSA version comparison

jzelinskie:
Previously we naively compared integers. However, not all versions have
the same length.

### DIFF
--- a/ext/vulnsrc/oracle/oracle.go
+++ b/ext/vulnsrc/oracle/oracle.go
@@ -85,6 +85,36 @@ func init() {
 	vulnsrc.RegisterUpdater("oracle", &updater{})
 }
 
+func compareELSA(left, right int) int {
+	// Fast path equals.
+	if right == left {
+		return 0
+	}
+
+	lstr := strconv.Itoa(left)
+	rstr := strconv.Itoa(right)
+
+	for i := range lstr {
+		// If right is too short to be indexed, left is greater.
+		if i >= len(rstr) {
+			return 1
+		}
+
+		ldigit, _ := strconv.Atoi(string(lstr[i]))
+		rdigit, _ := strconv.Atoi(string(rstr[i]))
+
+		if ldigit > rdigit {
+			return 1
+		} else if ldigit < rdigit {
+			return -1
+		}
+		continue
+	}
+
+	// Everything the length of left is the same.
+	return len(lstr) - len(rstr)
+}
+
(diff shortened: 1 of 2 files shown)


ext/vulnsrc/oracle/oracle.go
```diff
@@ -115,7 +150,7 @@ func (u *updater) Update(datastore database.Datastore) (resp vulnsrc.UpdateRespo
 		r := elsaRegexp.FindStringSubmatch(line)
 		if len(r) == 2 {
 			elsaNo, _ := strconv.Atoi(r[1])
-			if elsaNo > firstELSA {
+			if compareELSA(elsaNo, firstELSA) == 1 {
```
Quentin-M: Usually you'd test for `> 0` because this kind of comparison functions typically only offers the sign and 0 guarantees (C / Java) and return a distance/difference between the two (at least for the first part that is different).

i.e. ```return len(lstr) - len(rstr)```

ext/vulnsrc/oracle/oracle.go
```diff
@@ -85,6 +85,41 @@ func init() {
...
+		continue
+	}
+
+	// Everything the length of left is the same.
+	if len(lstr) == len(rstr) {
+		return 0
+	} else if len(lstr) > len(rstr) {
+		return 1
+	}
+	return -1
```
Quentin-M: the whole block can be replaced by `return len(lstr) - len(rstr)` - typically what's used.

ext/vulnsrc/oracle/oracle.go
```diff
@@ -85,6 +85,41 @@ func init() {
...
+		}
+
+		ldigit, _ := strconv.Atoi(string(lstr[i]))
+		rdigit, _ := strconv.Atoi(string(rstr[i]))
+
+		if ldigit > rdigit {
+			return 1
+		} else if ldigit < rdigit {
+			return -1
+		}
```
Quentin-M, line 110: can test for eq and return the diff too.

jzelinskie: nope, the equal case here should be `continue` which it already falls back to.
Go linters always want you to drop the last case like this /shrug

Quentin-M: i meant, you can do:
```
if ldigit != rdigit {
  return ldigit - rdigit
}
```

Quentin-M (approved):

PR #367, opened 2017-04-19
contrib: only extract layers from history

jzelinskie:
This tool was written before v2 existed and I'm not sure if our assumptions are still true.


Quentin-M (approved):

Quentin-M:
I need an auto-approver script.

PR #371, opened 2017-04-25
Add logging

caipre:
These changes add some logging information that was useful to me when working with clair.

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -210,7 +210,7 @@ func detectFeatureVersions(name string, files tarutil.FilesMap, namespace *datab
 			continue
 		}
 
-		log.Warningf("layer %s: Layer's namespace is unknown but non-namespaced features have been detected", name)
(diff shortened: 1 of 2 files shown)


jzelinskie (approved):
LGTM. Thanks!

worker.go
```diff
@@ -210,7 +210,7 @@ func detectFeatureVersions(name string, files tarutil.FilesMap, namespace *datab
 			continue
 		}
 
-		log.Warningf("layer %s: Layer's namespace is unknown but non-namespaced features have been detected", name)
+		log.Warningf("Namespace unknown for feature %s %s, in layer %s", feature.Feature.Name, feature.Version, name)
```
Quentin-M, line 213: why removing the prefix? the format is used everywhere else.

Quentin-M: ideally, we'd like to switch to logrus and use structured messages though!

jzelinskie: I'm going to make a follow up PR doing that.

Issue #69, opened 2016-02-11
analyze-local-image could not analyze layer: Got response 404 with message


jeremie0:
I've got clair running in a docker container:

```
-sh-4.2$  docker run --env-file ./proxy-env  -p 6060:6060 -p 6061:6061  -v /tmp:/tmp  -v /home/jeremie/quay-config:/config:ro quay.io/coreos/clair:latest --config=/config/config.yaml
2016-02-11 17:27:33.823292 I | database: database at /db/bolt does not exist yet, creating it
2016-02-11 17:27:33.835770 I | api: starting main API on port 6060.
2016-02-11 17:27:33.836521 I | notifier: notifier service is disabled
2016-02-11 17:27:33.836549 I | api: starting health API on port 6061.
2016-02-11 17:27:33.836819 I | updater: updater service started. lock identifier: f3a44bee-bf15-4242-8b0a-60a0c3f70bbe
2016-02-11 17:27:33.839272 I | updater: updating vulnerabilities
2016-02-11 17:27:33.839360 I | updater/fetchers: fetching Debian vulneratibilities
2016-02-11 17:27:33.839437 I | updater/fetchers: fetching Red Hat vulneratibilities
2016-02-11 17:27:33.839863 I | updater/fetchers: fetching Ubuntu vulneratibilities
2016-02-11 17:56:24.140684 N | database: Ignoring 17374 notifications
2016-02-11 17:56:24.492351 I | updater: update finished
2016-02-11 17:58:37.701983 E | worker: layer <missing>: failed to extract data from /tmp/analyze-local-image-257595031/<missing>/layer.tar: the resource cannot be found
2016-02-11 17:58:37.704748 I | api: 404 POST /v1/layers (96.482403ms)
```

 But when I try to analyze-local-images debian:jessie, it is able to save the image, but errors during analyis:

```
-sh-4.2$ ./work/bin/analyze-local-images debian:jessie
Saving debian:jessie
Getting image's history
Analyzing 2 layers
- Analyzing <missing>
2016/02/11 11:58:37 - Could not analyze layer: Got response 404 with message {"Message":"the resource cannot be found"}
```

If I look inside the container that's running clair, I do see the images in /tmp/

```
-sh-4.2$ docker exec insane_bohr ls -l /tmp/analyze-local-image-257595031/
total 20
[...]

PR #373, opened 2017-04-26
contrib: Fix analyze_local_images when image has missing layers

josuesdiaz:
This issue was originally addressed in #69, but looks to have been
reverted since then. As of right now, certain images that have missing
layers or layers that are not in the docker save output will cause an error
with the analyze_local_images command. Switching back to reading layers
from manifest.json file from docker save of the image.

### DIFF
--- a/contrib/analyze-local-images/main.go
+++ b/contrib/analyze-local-images/main.go
@@ -154,7 +154,10 @@ func AnalyzeLocalImage(imageName string, minSeverity types.Priority, endpoint, m
 
 	// Retrieve history.
 	log.Println("Retrieving image history")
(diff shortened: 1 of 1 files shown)


jzelinskie:
Hey there,

It looks like this is (sadly) the best way to do this. I had reproduced a bug locally using this method, but it seems far less likely to occur than to have the command have missing layers.

Can you make your commit simply a `git revert` of `e639367a3082c5a2067803e9c66f1258d4e2b919`?

josuesdiaz:
Ya sure, sorry should have thought to do that in the first place. Updated the branch, and you should see the revert only in PR now.

jzelinskie (approved):
LGTM.

PR #377, opened 2017-04-27
Clair only support 1.9.1 docker

wcc526:
Clair only support 1.9.1 docker

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Clair currently supports and tests against:
 * [Postgres] 9.4
 * [Postgres] 9.5
 * [Postgres] 9.6
+* [Docker] 1.9.1
 
 [Postgres]: https://www.postgresql.org
 


jzelinskie:
What makes you think that?

Quentin-M:
We have been using Clair on Quay.io for almost two years now. I doubt it'd be working only with Docker 1.9.1 Technically speaking, Clair is pretty-much runtime agnostic as it only expects to receive tarballs of filesystem diffs and reads the parent relationship from the API directly.

wcc526:
https://github.com/coreos/clair/issues/69

@jzelinskie @Quentin-M 

Please see this issue 

Clair is broken because of the content addressability inconsistencies in docker 1.10 (clair works fine till docker version 1.9.1); "docker history" command gives SHA256 layer IDs, while "docker save" command saves the layers.tar files in plain layer-UID directories.
Not sure if clair is going to fix the issue or will it be taken care of when docker/distribution#727 feature is supported by Docker.

When I run docker 1.12.1,I got Got response 404 with message error

PR #378, opened 2017-04-27
ext/vulnsrc/oracle: ensure flag is largest elsa

jzelinskie:
If the Oracle Linux directory is ever in the wrong order, this should
ensure that the updaterFlag is always set the latest ELSA value.


Quentin-M (approved):
the meme is real

Issue #372, opened 2017-04-26
{"Error":{"Message":"could not find layer"}}


wcc526:
/root/projects/bin/analyze-local-images quay.io/coreos/clair-git:latest
2017-04-26 15:42:40.385232 I | Saving quay.io/coreos/clair-git:latest to local disk (this may take some time)
2017-04-26 15:42:48.722948 I | Retrieving image history
2017-04-26 15:42:48.735735 I | Analyzing 18 layers...
2017-04-26 15:42:48.735755 I | Analyzing <missing>
2017-04-26 15:42:48.738308 I | Could not analyze layer: Got response 400 with message {"Error":{"Message":"could not find layer"}}

 docker attach 3bb0b52dfece
2017-04-26 07:37:23.915746 E | clair: layer <missing>: failed to extract data from /tmp/analyze-local-image-217567589/<missing>/layer.tar: could not find layer
2017-04-26 07:37:23.915991 I | v1: 172.18.0.1:37518 "POST /v1/layers" 400 (1.223681ms)
2017-04-26 07:39:31.313341 I | v1: 10.2.4.191:50398 "GET /v1/namespaces" 200 (801.95µs)
2017-04-26 07:42:05.768879 E | clair: layer <missing>: failed to extract data from /tmp/analyze-local-image-278772881/<missing>/layer.tar: could not find layer
2017-04-26 07:42:05.769150 I | v1: 172.18.0.1:37538 "POST /v1/layers" 400 (1.375868ms)
2017-04-26 07:42:20.720419 E | clair: layer <missing>: failed to extract data from /tmp/analyze-local-image-463395154/<missing>/layer.tar: could not find layer
2017-04-26 07:42:20.720817 I | v1: 172.18.0.1:37544 "POST /v1/layers" 400 (1.533479ms)
2017-04-26 07:42:48.737657 E | clair: layer <missing>: failed to extract data from /tmp/analyze-local-image-217560860/<missing>/layer.tar: could not find layer
2017-04-26 07:42:48.738027 I | v1: 172.18.0.1:37550 "POST /v1/layers" 400 (1.346907ms)
2017-04-26 07:43:08.665162 E | clair: layer <missing>: failed to extract data from /tmp/analyze-local-image-047019759/<missing>/layer.tar: could not find layer
2017-04-26 07:43:08.665431 I | v1: 172.18.0.1:37554 "POST /v1/layers" 400 (1.735923ms)

docker version
Client:
 Version:         1.12.0
 API version:     1.24
[...]

PR #379, opened 2017-04-28
Clair only support docker 1.9.1

wcc526:
Clair only support docker 1.9.1

It will go wrong if docker run on 1.13.1 or 1.12

Clair is broken because of the content addressability inconsistencies in docker 1.10 (clair works fine till docker version 1.9.1); "docker history" command gives SHA256 layer IDs, while "docker save" command saves the layers.tar files in plain layer-UID directories.
Not sure if clair is going to fix the issue or will it be taken care of when docker/distribution#727 feature is supported by Docker.

https://github.com/coreos/clair/issues/372

https://github.com/coreos/clair/issues/69

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Clair currently supports and tests against:
 * [Postgres] 9.4
 * [Postgres] 9.5
 * [Postgres] 9.6
+* [Docker] 1.9.1
 
 [Postgres]: https://www.postgresql.org
 


jzelinskie:
Hi there! Thanks for re-opening this issue!

>Clair only support docker 1.9.1

This is not true. Only the `analyze-local-images` tool is affected by this, but Clair is not.
I recently wasted a lot due to this behavior in `analyze-local-images` which ultimately led to the regression being fixed in [this PR](https://github.com/coreos/clair/pull/373). `analyze-local-images` is really only good for testing locally and should not be used in any kind of production environment. You should strive to consume the Clair API directly.

Clair uses whatever name for layers is provided from the API. It does not require content addressability, but rather any unique name for a layer in a given ancestry. If you were to just provide only the SHA256 as the names for layers uploaded via the Clair API, you would eventually have a collision and Clair's internal representation of the ancestry would be broken. The registries that integrate with Clair use a combination of a unique identifier for an image in addition to an individual layer's SHA256 (e.g. `manifestSHA-layerSHA`).

Have you tried using an alternative client such as [clairctl](https://github.com/jgsqware/clairctl)?

adityacs:
I am using clairctl and latest clair. I get the below error.  Unable to push image. I am using docker 17.05-ce

image

**clair logs** 
image

jgsqware:
Hello you could open an issue in clairctl repository, I will help you there
;)

adityacs:
@jgsqware Created https://github.com/jgsqware/clairctl/issues/43

Issue #319, opened 2017-02-17
False positives due to a layer having different parents


MrTrustor:
Hello,

First of all, thank you for your work, Clair is an impressive software.

We are using Clair internally for our private repositories and we ran into the issue described in #68. To sum up: Clair is not aware of the whole chain of layers that make up a Docker image. As a layer is identified by its content, you can have the same layer in different images, with different parents, leading to false positives (or false negatives).

Here is how we ran into the problem:

1. We built an image ``FROM grafana/grafana:4.1.1``. Clair informed us that there were some vulnerabilities in this image (because it was not built from the latest ``debian:jessie`` image). The Dockerfile looks like this:
```
FROM grafana/grafana:4.1.1
COPY file1 /foo
COPY file2.sh /bar
ENTRYPOINT ['/bar/file2.sh']
```

2. To fix the problem, we took the Dockerfile from ``grafana/grafana:4.1.1`` and sourced our image directly from the latest ``debian:jessie``:
```
FROM debian:jessie
ARG DOWNLOAD_URL
RUN apt-get update && \
    apt-get -y --no-install-recommends install libfontconfig curl ca-certificates && \
    apt-get clean && \
    curl ${DOWNLOAD_URL} > /tmp/grafana.deb && \
    dpkg -i /tmp/grafana.deb && \
    rm /tmp/grafana.deb && \
    curl -L https://github.com/tianon/gosu/releases/download/1.7/gosu-amd64 > /usr/sbin/gosu && \
    chmod +x /usr/sbin/gosu && \
    apt-get remove -y curl && \
    apt-get autoremove -y && \
    rm -rf /var/lib/apt/lists/*
EXPOSE 3000
COPY file1 /foo
COPY file2.sh /bar
ENTRYPOINT ['/bar/file2.sh']
```

The last few layers are the same in the 2 resulting images, but the parents are not. In the first case, the parents have some vulnerabilities while they have not in the second (they have only vulnerabilities for which there is no fix).

When POSTing the layers of the second image to Clair, we got a 201 status code, but the parent of the layers were not updated in the database. So, when fetching the vulnerabilities for the last layer of the second image, we actually got the vulnerabilities of the first image.

In #68, it is said that the issue has been identified and put into the roadmap, but the current roadmap makes no mention of it. Is this problem still identified somewhere?

MrTrustor:
Any answer to this?

jzelinskie:
[...]

PR #382, opened 2017-05-02
pgsql: Change layer name column data type

caipre:
Presently the `layer.name` column has type `varchar(128)`. This works
fine enough when using the sha256 layer digests provided by docker.
However, if one wishes to encode the image name into the layer (eg, to
avoid collisions like in https://github.com/coreos/clair/issues/319),
the limit of 128 bytes becomes a bit cramped.

I've configured my client to name layers using the docker layer digest

    sha256:21a547c3a96a390f700a68bc506a26e5c6e8f129fce41e05485d907d7fc130c1

concatenated with sha256(image.name, image.tag, layer.parent.digest):

    69429e10ce617a247feca82cf50ffc9bd071c1184e6910872f26b5c3d37325d3

This change allows the name column to support such a format.

### DIFF
--- a/database/pgsql/migrations/00007_expand_column_width.go
+++ b/database/pgsql/migrations/00007_expand_column_width.go
@@ -1,0 +1,31 @@
+// Copyright 2017 clair authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migrations
+
+import "github.com/remind101/migrate"
+
+func init() {
+	RegisterMigration(migrate.Migration{
+		ID: 7,
+		Up: migrate.Queries([]string{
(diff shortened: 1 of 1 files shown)


caipre:
If this change is acceptable, I'm happy to write the migration code as well.

Quentin-M:
Hey,

We do actually use a combination of a few things as well in production. I'd be happy to merge this thing, but wondering if there is any performance hit involved here at scale?

Also, I believe you'd need to make a new migration, do not modify old ones, it might break existing deployments. We have a few casts here and there, I'd verify if there is any applying to that field.

Thanks!

caipre:
@Quentin-M : Thanks for the review; I've updated the commit to do a migration rather than modify the initial schema. I looked into the [Postgres docs](https://www.postgresql.org/docs/current/static/datatype-character.html) and they claim there is no performance penalty: `text` and `varchar` share the same implementation underneath. I also didn't see any casts on these columns, but I'm not sure where in the codebase they might be.

caipre:
@Quentin-M Mind taking another look at this?

jzelinskie:
This change looks good from the database-side; however, I do think there should be _some_ limit via the API to prevent DoS. I think we need to discuss a fair length and then add another commit to enforce this at the API before merging.

caipre:
@jzelinskie That's a valid concern. How would you like to proceed? For my purposes, the column need not be unbounded. Perhaps just bumping the width up to `varchar(256)` would be more agreeable?

jzelinskie:
@caipre This sounds fair. If you're quick, I'll cut this in v2.0.0.

caipre:
Looks like I missed the release, but I've updated the commit.

jzelinskie (approved):
LGTM! Thanks for the PR. Sorry about the rush for v2.0.0.

PR #389, opened 2017-05-05
Regenerate vendor redirectory

jzelinskie:
The previous configuration had dependencies from contrib projects that are now removed.


Quentin-M:
Giphy

Quentin-M (approved):

PR #392, opened 2017-05-11
ext/vulnsrc/alpine: use HTTPS

jzelinskie:


Quentin-M (approved):
Do we trust PKI? :1st_place_medal:

PR #394, opened 2017-05-12
added support for detecting multiple namespaces in a layer

### LINKED ISSUE #150: detect multiply namespaces in a layer

KeyboardNerd:
created table layer_namespace to store the many to many unique mapping of layers and namespaces
changed v1 api to provide a list of namespaces for each layer
changed namespace detector to use all registered detectors to detect namespaces
updated tests for multiple namespaces

Fixes #150

### DIFF
--- a/database/models.go
+++ b/database/models.go
@@ -31,7 +31,7 @@ type Layer struct {
 	Name          string
 	EngineVersion int
 	Parent        *Layer
-	Namespace     *Namespace
+	Namespaces    []Namespace
 	Features      []FeatureVersion
 }
 
--- a/database/pgsql/layer.go
+++ b/database/pgsql/layer.go
@@ -52,9 +52,6 @@ func (pgSQL *pgSQL) FindLayer(name string, withFeatures, withVulnerabilities boo
 		&layer.EngineVersion,
 		&parentID,
 		&parentName,
-		&nsID,
-		&nsName,
-		&nsVersionFormat,
 	)
 	observeQueryTime("FindLayer", "searchLayer", t)
 
@@ -68,11 +65,23 @@ func (pgSQL *pgSQL) FindLayer(name string, withFeatures, withVulnerabilities boo
 			Name:  parentName.String,
 		}
 	}
-	if !nsID.IsZero() {
-		layer.Namespace = &database.Namespace{
-			Model:         database.Model{ID: int(nsID.Int64)},
-			Name:          nsName.String,
-			VersionFormat: nsVersionFormat.String,
+
+	rows, err := pgSQL.Query(searchLayerNamespace, layer.ID)
+	defer rows.Close()
+	if err != nil {
+		return layer, handleError("searchLayerNamespace", err)
+	}
+	for rows.Next() {
+		err = rows.Scan(&nsID, &nsName, &nsVersionFormat)
+		if err != nil {
+			return layer, handleError("searchLayerNamespace", err)
+		}
+		if !nsID.IsZero() {
+			layer.Namespaces = append(layer.Namespaces, database.Namespace{
+				Model:         database.Model{ID: int(nsID.Int64)},
+				Name:          nsName.String,
+				VersionFormat: nsVersionFormat.String,
+			})
 		}
 	}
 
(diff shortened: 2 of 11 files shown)


database/models.go
```diff
@@ -30,9 +30,14 @@ type Layer struct {
...
-	Namespace     *Namespace
-	Features      []FeatureVersion
+
+
+	Parent *Layer
+
+	Namespaces []Namespace
+
+	Features []FeatureVersion
+
```
jzelinskie: Any reason why you spaced these out like this? Did this pass gofmt?

database/pgsql/layer.go
```diff
@@ -313,14 +326,37 @@ func (pgSQL *pgSQL) InsertLayer(layer database.Layer) error {
...
 		}
 
+		// replace the old namespace in the database
+		_, err := tx.Exec(removeLayerNamespace, layer.ID)
+		if err != nil {
+			tx.Rollback()
+			return handleError("removeLayerNamespace", err)
+		}
+		for nsid := range namespaceIDs {
+			_, err := tx.Exec(insertLayerNamespace, layer.ID, nsid)
```
jzelinskie: This could all be done with one insert rather one query per namespace

database/pgsql/queries.go
```diff
@@ -77,12 +77,17 @@ const (
...
-			LEFT JOIN Namespace n ON l.namespace_id = n.id
+			SELECT l.id, l.name, l.engineversion, p.id, p.name
+			FROM Layer l
+				LEFT JOIN Layer p ON l.parent_id = p.id
 		WHERE l.name = $1;`
 
+	searchLayerNamespace = `SELECT n.id, n.name, n.version_format 
+							FROM Namespace n
+								JOIN Layer_Namespace lns ON lns.namespace_id = n.id 
+							WHERE lns.layer_id = $1
```
jzelinskie: indentation is a little funky here

jzelinskie (changes requested):
awesome work! mostly just nitpicks

database/models.go
KeyboardNerd: ah, the broken gofmt in emacs caused this

database/pgsql/migrations/00007_add_multiplens.go
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migrations
+
+import "github.com/remind101/migrate"
+
+func init() {
+	RegisterMigration(migrate.Migration{
+		ID: 7,
```
jzelinskie: This needs to change to 8.

database/pgsql/layer.go
```diff
@@ -329,6 +348,21 @@ func (pgSQL *pgSQL) InsertLayer(layer database.Layer) error {
...
+	stmt, err := tx.Prepare(insertLayerNamespace)
+	for nsid := range namespaceIDs {
+		_, err := stmt.Exec(layer.ID, nsid)
+		if err != nil {
+			tx.Rollback()
+			return handleError("insertLayerNamespace", err)
+		}
+	}
+
+	err = stmt.Close()
```
jzelinskie: If anything fails while iterating over the namespaceIDs, the statement never gets closed. I think it's best to structure it like this instead:
```go
stmt, err := tx.Prepare(insertLayerNamespace)
defer func() {
  err = stmt.Close()
  if err != nil {
    tx.Rollback()
    log.WithFields(log.Fields{"error": err}).Error("failed to close prepared statement")
}()

for nsid := range namespaceIDs {
  ...
}
```

database/pgsql/queries.go
```diff
@@ -77,12 +77,18 @@ const (
...
+			FROM Layer l
+				LEFT JOIN Layer p ON l.parent_id = p.id
 		WHERE l.name = $1;`
 
+	searchLayerNamespace = `
+			SELECT n.id, n.name, n.version_format 
+			FROM Namespace n
+				JOIN Layer_Namespace lns ON lns.namespace_id = n.id 
+			WHERE lns.layer_id = $1
+							`
```
jzelinskie: let's move this ` to right after the $1

database/pgsql/testdata/data.sql
```diff
@@ -28,12 +28,19 @@ INSERT INTO featureversion (id, feature_id, version) VALUES
...
-  (5, 'layer-3b', 1, 3, 2);
+INSERT INTO layer (id, name, engineversion, parent_id) VALUES
+  (1, 'layer-0', 1, NULL),
+  (2, 'layer-1', 1, 1),
+  (3, 'layer-2', 1, 2),
+  (4, 'layer-3a', 1, 3),
+  (5, 'layer-3b', 1, 3);
+
+INSERT INTO layer_namespace (id, layer_id, namespace_id) VALUES
+(1, 2, 1),
```
jzelinskie: indent

ext/featurens/driver_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestMultipleNamespaceDetector(t *testing.T) {
+	testData := []MultipleNamespaceTestData{
+		{
+			ExpectedNamespaces: []database.Namespace{
+				database.Namespace{Name: "debian:8", VersionFormat: "dpkg"},
+				database.Namespace{Name: "alpine:v3.3", VersionFormat: "dpkg"},
+			},
+			Files: tarutil.FilesMap{
+				"etc/os-release": []byte(
+					`PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
```
jzelinskie: put the ` on the line above and start PRETTYNAME on a newline right above NAME

jzelinskie (changes requested):

database/pgsql/layer.go
```diff
@@ -329,6 +348,30 @@ func (pgSQL *pgSQL) InsertLayer(layer database.Layer) error {
...
+
+	defer func() {
+		err = stmt.Close()
+		if err != nil {
+			tx.Rollback()
+			log.WithError(err).Error("failed to close prepared statement")
+		}
+	}()
+
+	if err != nil {
```
jzelinskie: move this above the defer or else the defer could possibly cause a nil ptr deference

jzelinskie (changes requested):

jzelinskie (approved):

PR #395, opened 2017-05-14
versionfmt/rpm: handle a tilde correctly

knqyf263:
I found a bug that the handling of tilde is incorrect.
So, the following tests will fail in `TestParseAndCompare`.

```
{"~", GREATER, "~~"},
{"2~", GREATER, "1"},
```

A tilde must sort before everything else (including the blank).
However, it is necessary to compare numbers or strings before the tilde.

### DIFF
--- a/ext/versionfmt/rpm/parser.go
+++ b/ext/versionfmt/rpm/parser.go
@@ -212,23 +212,9 @@ func rpmvercmp(strA, strB string) int {
 			} else if len(b) > len(a) {
 				return -1
 			}
-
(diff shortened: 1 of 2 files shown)


jzelinskie (approved):
LGTM! Thanks a bunch. The RPM spec for versioning is very strange!

knqyf263:
@jzelinskie Thanks! 
I'm also struggling with RPM versioning at [Vuls](https://github.com/future-architect/vuls) development.
Let's both do our best!

PR #407, opened 2017-06-01
Updated Kubernetes config with new db connection string config

swestcott:
Updates the Kubernetes sample config for the Clair v2 format to fix the following error I was seeing,

```
{"Event":"pgsql: no database connection string specified","Level":"fatal","Location":"main.go:96","Time":"2017-06-01 22:32:23.645913"}
```


jzelinskie (approved):
LGTM! Thanks! We should really replace this stuff with a Helm chart.

PR #410, opened 2017-06-06
api: fix remote addr shows reverse proxy addr problem

KeyboardNerd:
Uses the first ip addr in X-forwarded-for as the client's remote addr if it exists
otherwise, fall back to use default http.Request.RemoteAddr

### DIFF
--- a/api/httputil/httputil.go
+++ b/api/httputil/httputil.go
@@ -1,0 +1,24 @@
+package httputil
+
+import (
+	"net"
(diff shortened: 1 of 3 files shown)


jzelinskie:
Let's move this code so that it's a subdirectory of `api`. Maybe we can call it `httputil`, so that we can put other stuff besides IP-related stuff in the package.

jzelinskie (approved):

Issue #397, opened 2017-05-19
Ubuntu precise/esm is not mapped


rtoma:
Hi,

Found this in my logs - thought you might like to know.

```
{"Event":"fetcher note","Level":"warning","Location":"updater.go:189",
  "Time":"2017-05-19 09:05:14.403832",
  "note":"Ubuntu precise/esm is not mapped to any version number (eg. trusty-\u003e14.04). Please update me."}
```

transcedentalia:
I'm getting the same error.

This might be correlated to the fact that running analyze-local-images with ubuntu_trusty and ubuntu_xenial docker images is providing the following output:

```
analyze-local-images c6a90c461866
2017-06-06 09:38:35.869374 I | Saving c6a90c461866 to local disk (this may take some time)
2017-06-06 09:38:47.548710 I | Retrieving image history
2017-06-06 09:38:47.548942 I | Analyzing 9 layers...
2017-06-06 09:38:47.548959 I | Analyzing 768d4f50f65f00831244703e57f64134771289e3de919a576441c9140e037ea2
2017-06-06 09:38:47.560230 I | Analyzing 4b02c0b8bb67a5515b901638a31d40128cf182cc19c23370a0546c8a0bd0e38a
2017-06-06 09:38:47.564423 I | Analyzing 1b0189593ef7b3cd68a0c72a1ef10d026f3576988fcf340ce053e8ca0a64d4ef
2017-06-06 09:38:47.567705 I | Analyzing e99fb703135a078c9ecc06692b864eafc0005df12295c4f1ae693ff2fa0cf10a
2017-06-06 09:38:47.571155 I | Analyzing f1c27873830f2847e52a9fee3fbf0bce6c293ff900aaf8dafdc382930b8906cd
2017-06-06 09:38:47.577937 I | Analyzing e5fe1023d344a4f9ad8bc6ea9149411d726d2fd43318cc9610e3df34b95aae82
2017-06-06 09:38:47.587933 I | Analyzing 9b1be316758d5d85d67977114f2e4fac642d866fc6ab68466ffac7b8c02a9913
2017-06-06 09:38:47.590944 I | Analyzing 7e030f432297d4563d044dbb35ed1ceef0aa0f8e99bf9c560365f3dada757e86
2017-06-06 09:38:47.598218 I | Analyzing 43486342132c2daf5066a0dcd977158d4259ebe2f3423653c3cee267c241b581
[...]

PR #413, opened 2017-06-07
Fixing the always database revision 0 for Ubuntu

transcedentalia:
This is the fix for https://github.com/coreos/clair/issues/397


jzelinskie (approved):
LGTM. Thanks!

PR #418, opened 2017-06-22
use namespace's versionfmt to specify listers scanning features

KeyboardNerd:


ext/featurefmt/driver.go
```diff
@@ -49,7 +52,7 @@ type Lister interface {
 //
 // If called twice with the same name, the name is blank, or if the provided
 // Lister is nil, this function panics.
-func RegisterLister(name string, l Lister) {
+func RegisterLister(name string, versionfmt string, l Lister) {
```
jzelinskie, line 55: `name, versionfmt string`
this will bother me forever, but I'm not going to prevent the merge for it.

jzelinskie (approved):

PR #424, opened 2017-06-29
Option to provide namespace'name when posting layer

yebinama:
This adds an optional field "NamespaceName" to the json provided when posting a Layer.
If during the analyse, Clair can't find the namespace associated to the Layer, we use the namespace provided by the user (only if it exists in the database).
I use this option to resolve the http 422 error I had when uploading some of my base images.

I made the change on the release-2.0 branch because it's the one I use at work and I know you are working on supporting multiple namespaces on the master branch. If it is in stable state, I can adapt the changes to that branch.

### DIFF
--- a/Documentation/api_v1.md
+++ b/Documentation/api_v1.md
@@ -66,6 +66,7 @@ In order to analyze a container image, this route has to be called for each laye
 This request blocks for the entire duration of the downloading and indexing of the layer and displays the provided Layer with an updated `IndexByVersion` property.
 The Name field must be unique globally. Consequently, using the Blob digest describing the Layer content is not sufficient as Clair won't be able to differentiate two empty filesystem diffs that belong to two different image trees.
 The Authorization field is an optional value whose contents will fill the Authorization HTTP Header when requesting the layer via HTTP.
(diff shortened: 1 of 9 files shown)


jzelinskie:
I appreciate the foresight to target the stable branch with this change! Thanks again for that.

Is there a particular reason that the namespace detectors aren't able to determine the namespace without you explicitly declaring them in the API? I think being able to provide namespaces is valuable, but I just want to make sure we aren't overlooking any fixable issues with our namespace detectors.

yebinama:
The Http 422 error occured with layers containing "/var/lib/dpkg/status" (features detected) but no "/etc/apt/sources.list" (namespace can't be detected).
In our base image's Dockerfile we delete sources.list and replace it with one containing our repositories.

There was in fact a bug on my code which sends the layers.
The problem is that I read manifests from multiples registries and after some extended research, it appears some have schemaVersion 1 and others have schemaVersion 2. Therefore the layers are not displayed in the same order for each image.
I had the error only  because the layers were uploaded in the wrong order.

Don't know if the pull request is still relevant.

PR #432, opened 2017-07-12
gRPC API & internals redesign

KeyboardNerd:
This is a large Pull Request that will contain a number of chunks. For easier code review, the Clair V3 PR will contains these commits separately.
1. API changes
2. datastore interface changes
3. mock datastore, extensions update, worker, updater, notifier changes with tests
4. datastore pgsql implementation with tests
5. potential vendor clean up


api/v2/clairpb/clair.proto
```diff
@@ -26,57 +27,63 @@ message Vulnerability {
 	string link = 4;
 	string severity = 5;
 	string metadata = 6;
+	// fixed_by only exists when querying features
```
jzelinskie: Can we treat comments like Go ones? Full sentences 80 char lines.

api/v2/clairpb/clair.proto
```diff
@@ -26,57 +27,63 @@ message Vulnerability {
...
 	string fixed_by = 7;
-	repeated Feature fixed_in_features = 8;
+	// affected_versions only exists when querying notifications
+	repeated Feature affected_versions = 8;
 }
 
-message Feature {
+message ClairStatus {
+	// all listers and detectors implemented in this clair version
```
jzelinskie: Is this what Clair is built with or configured with?

api/v2/clairpb/clair.proto
```diff
@@ -26,57 +27,63 @@ message Vulnerability {
...
-message Page {
+message IndexedAncestryName {
+	// index should be unique to name and in increasing order based on the page
+	int32 index = 1;
+	string name = 2;
+}
+
+message PagedVulnerableAncestries {
 	string this_token = 1;
 	string next_token = 2;
```
jzelinskie: maybe rename these 'current_page' and 'next_page'

api/v2/clairpb/clair.proto
```diff
@@ -117,8 +125,6 @@ message DeleteNotificationRequest {
 	string name = 1;
 }
 
-
-
 service AncestryService{
```
jzelinskie: I cannot comment on the line, but we should rename "DeleteNotification" to "MarkNotificationAsRead" and make HTTP DELETE be an implementation detail of the rest gateway

api/v2/rpc.go
```diff
@@ -164,93 +193,23 @@ func (s *NotificationServer) DeleteNotification(ctx context.Context, req *pb.Del
 		return nil, status.Error(codes.InvalidArgument, "Failed to provide notification name")
 	}
 
-	err := s.Store.DeleteNotification(req.GetName())
+	tx, err := s.Store.Begin()
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
```
jzelinskie: any reason you don't defer the rollback here and instead explicitly call it for each branch below?

jzelinskie (changes requested):

api/v2/rpc.go
KeyboardNerd: I don't think Rollback() should ever be called after Commit() on success.

api/v2/clairpb/clair.proto
KeyboardNerd: all built-in listers/detectors. The listers and detectors are not configurable now, should I add that to this PR?

api/v2/rpc.go
jzelinskie: Rollback() from database/sql actually is safe to do this and is considered an idiomatic pattern to do so. I think it makes sense both ways, though.

api/v2/clairpb/clair.proto
jzelinskie: Nope, we can do that later, just as long as nothing here prevents us from doing so.

database/models.go
```diff
@@ -17,52 +17,108 @@ package database
...
+	NamespacedFeature
 
-	Feature    Feature
-	Version    string
 	AffectedBy []Vulnerability
+}
 
-	// For output purposes. Only make sense when the feature version is in the context of an image.
-	AddedBy Layer
+// AffectedFeature is used determine whether a namespaced feature is affected.
```
jzelinskie: AffectedFeature is used to determine whether a namespaced feature is affected by a Vulnerability.

KeyboardNerd:
I'm using my analyze_local_images fork's py tools to test the APIs, which seems to work properly. I'll do more concurrency tests and the web hook tests.

Besides the implementation of new schema and analyzing functions, I also added the knob to control feature listers and namespace detectors to scan the layers and updaters to fetch vuln src.

api/v2/clairpb/convert.go
```diff
@@ -18,36 +18,55 @@ import (
...
-		if err != nil {
-			return nil, err
-		}
-		newVuln = v
+	vuln, err := VulnerabilityFromDatabaseModel(dbVuln.Vulnerability)
+	if err != nil {
+		return nil, err
+	}
+
+	// encrypt page numbers
```
jzelinskie: this isn't a particularly useful comment

api/v2/clairpb/convert.go
```diff
@@ -18,36 +18,55 @@ import (
...
+	nextToken, err := dbVuln.Next.Encrypt(paginationKey)
```
jzelinskie: What happens when there is no next page? Is there a special constant for that or does this function need to handle `nil`?

api/v2/clairpb/convert.go
```diff
@@ -59,35 +78,27 @@ func NotificationFromDatabaseModel(dbNotification database.VulnerabilityNotifica
...
-		Limit:    int32(limit),
-		Page: &Page{
-			ThisToken: pageToken,
-			NextToken: nextPageStr,
-			Old:       oldVuln,
-			New:       newVuln,
-		},
-	}, nil
-}
+	pbOldVuln, err := PagedVulnerableAncestriesFromDatabaseModel(dbOldVuln, paginationKey)
```
jzelinskie: What happens if there is no old vulnerability? Doesn't this need to handle `nil`?

api/v2/rpc.go
```diff
@@ -46,211 +43,221 @@ type AncestryServer struct {
...
+	pbAncestry := pb.AncestryFromDatabaseModel(ancestry)
+	if req.GetWithFeatures() || req.GetWithVulnerabilities() {
+		ancestryWFeature, ok, err := tx.FindAncestryFeatures(ancestry.Name)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		if !ok {
+			return nil, status.Error(codes.NotFound, "requested ancestry \""+ancestry.Name+" \" is not found")
```
jzelinskie: raw string literals will make this easier to read

```go
return nil, status.Error(`requested ancestry "` + ancestry.Name + `" is not found`)
```

api/v2/rpc.go
```diff
@@ -46,211 +43,221 @@ type AncestryServer struct {
...
+		pbAncestry.ScannedDetectors = ancestryWFeature.ProcessedBy.Detectors
+		pbAncestry.ScannedListers = ancestryWFeature.ProcessedBy.Listers
+
+		if req.GetWithVulnerabilities() {
+			featureVulnerabilities, err := tx.FindAffectedNamespacedFeatures(ancestryWFeature.Features)
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+
+			// ensure every feature is found
```
jzelinskie:
```go
// Ensure that every feature can be found.
```

I'd also move this comment to be above `if !fv.Valid` instead of the whole for loop.

database/models.go
```diff
@@ -17,52 +17,116 @@ package database
...
+// Namespace and Feature Name is unique. Affected Feature is bound to
+// vulnerability.
+type AffectedFeature struct {
+	Namespace   Namespace
+	FeatureName string
+	// known fixed in version for the vulnerability
```
jzelinskie: full sentences for comments

database/models.go
```diff
@@ -71,17 +135,78 @@ type Vulnerability struct {
 	Severity    Severity
 
 	Metadata MetadataMap
+}
+
+// VulnerabilityWithAffected is an vulnerability with all known affected
```
jzelinskie, line 142: **a** vulnerability

database/models.go
```diff
@@ -71,17 +135,78 @@ type Vulnerability struct {
...
+	// should be unique in a stream. Every indexes in the map should be larger
+	// than previous page.
+	Affected map[int]string
+
+	Limit   int
+	Current PageNumber
+	Next    PageNumber
 }
 
+// Notification is a hook used to notify and give token to user to retrieve the
```
jzelinskie: Let's rename this NotificationHook.

NotificationHook is a message sent to another service to inform of a change to a Vulnerability or the Ancestries affected by a Vulnerability. It contains the name of a notification that should be read and marked as read via the API.

database/models.go
```diff
@@ -71,17 +135,78 @@ type Vulnerability struct {
...
+
+	Old *Vulnerability
+	New *Vulnerability
+}
+
+// PageNumber is used to do pagination.
+type PageNumber struct {
+	// StartID is an implementation detail for paginating by an ID required to
+	// be unique to every ancestry and always increasing.
+	StartID int
```
jzelinskie: Ideally, this is just a string and we encrypted this before it ever leaves the interface, making it an opaque string to everyone but the implementation of the data store interface.

pkg/strutil/strutil.go
```diff
@@ -0,0 +1,41 @@
+package strutil
```
jzelinskie, line 15: copyright header

pkg/strutil/strutil_test.go
```diff
@@ -0,0 +1,20 @@
+package strutil
```
jzelinskie, line 15: copyright header

cmd/clair/main.go
```diff
@@ -30,26 +30,30 @@ import (
...
 
 	// Register database driver.
 	_ "github.com/coreos/clair/database/pgsql"
 
 	// Register extensions.
+	"github.com/coreos/clair/ext/featurefmt"
 	_ "github.com/coreos/clair/ext/featurefmt/apk"
 	_ "github.com/coreos/clair/ext/featurefmt/dpkg"
 	_ "github.com/coreos/clair/ext/featurefmt/rpm"
+	"github.com/coreos/clair/ext/featurens"
```
jzelinskie: import these with `strutil`, `stopper`, etc... leave these only for the side-effect imports that require `_`

cmd/clair/main.go
```diff
@@ -85,6 +89,43 @@ func stopCPUProfiling(f *os.File) {
 	log.Info("stopped CPU profiling")
 }
 
+func configClairVersion(config *Config) {
+	// prop to user of the current clair registered compontent
```
jzelinskie: I'm not sure what you're trying to say here. Not sure if `prop` is the word you're looking for.

config.example.yaml
```diff
@@ -51,10 +51,29 @@ clair:
     keyfile:
     certfile:
 
+  worker:
+    enableddetectors:
```
jzelinskie: let's call these `namespace_detectors`

config.example.yaml
```diff
@@ -51,10 +51,29 @@ clair:
...
+      - os-release
+      - lsb-release
+      - apt-sources
+      - alpine-release
+      - redhat-release
+
+    enabledlisters:
```
jzelinskie: let's call this `feature_listers`

ext/featurefmt/apk/apk_test.go
```diff
@@ -19,58 +19,65 @@ import (
...
+	"github.com/coreos/clair/ext/versionfmt/dpkg"
 	"github.com/coreos/clair/pkg/tarutil"
 )
 
 func TestAPKFeatureDetection(t *testing.T) {
+	testFeatures := []database.Feature{
+		{
+			Name:    "musl",
+			Version: "1.1.14-r10",
+		},
```
jzelinskie: Does gofmt prevent this?

```go
testFeatures := []database.Feature{
  {Name: "musl", Version: "1.1.14-r10"},
  {Name: "busybox", Version: "1.24.2-r9"},
}
```

ext/featurefmt/driver.go
```diff
@@ -117,10 +107,19 @@ func RequiredFilenames() (files []string) {
 	return
 }
 
+// ListListers returns a list of all listers' names
```
jzelinskie: Let's avoid a tongue twister here:

>ListListers returns the names of all the registered feature listers.

ext/featurens/driver.go
```diff
@@ -106,6 +110,14 @@ func RequiredFilenames() (files []string) {
 	return
 }
 
+func ListDetectors() []string {
```
jzelinskie: >ListDectectors returns the names of all registered namespace detectors.

ext/versionfmt/rpm/parser.go
```diff
@@ -121,6 +121,19 @@ func (p parser) Valid(str string) bool {
...
 
+func (p parser) InRange(versionA, rangeB string) (bool, error) {
+	cmp, err := p.Compare(versionA, rangeB)
+	if err != nil {
+		return false, err
+	}
+	return cmp < 0, nil
+}
+
+// only for temporary!
```
jzelinskie: o rly

ext/vulnsrc/debian/debian.go
```diff
@@ -62,19 +62,32 @@ func init() {
...
 	}
 
 	// Get the SHA-1 of the latest update's JSON data
-	latestHash, err := datastore.GetKeyValue(updaterFlag)
+	latestHash, ok, err := tx.FindKeyValue(updaterFlag)
 	if err != nil {
 		return resp, err
 	}
 
+	if err := tx.Rollback(); err != nil {
```
jzelinskie: why is this rolling back always here?

ext/vulnsrc/driver.go
```diff
@@ -88,3 +87,11 @@ func Updaters() map[string]Updater {
 
 	return ret
 }
+
+func ListUpdaters() []string {
```
jzelinskie: >ListUpdaters returns the names of registered vulnerability updaters.

ext/vulnsrc/rhel/rhel.go
```diff
@@ -90,11 +90,24 @@ func init() {
...
+	}
+
 	// Get the first RHSA we have to manage.
-	flagValue, err := datastore.GetKeyValue(updaterFlag)
+	flagValue, ok, err := tx.FindKeyValue(updaterFlag)
 	if err != nil {
 		return resp, err
 	}
+
+	tx.Rollback()
```
jzelinskie: why always rollback here?

ext/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -98,12 +98,22 @@ func (u *updater) Update(datastore database.Datastore) (resp vulnsrc.UpdateRespo
...
+	}
+
 	// Get the latest revision number we successfully applied in the database.
-	dbRevisionNumber, err := datastore.GetKeyValue("ubuntuUpdater")
+	dbRevisionNumber, ok, err := tx.FindKeyValue("ubuntuUpdater")
 	if err != nil {
 		return resp, err
 	}
 
+	tx.Rollback()
```
jzelinskie: why always rollback here?

updater.go
```diff
@@ -349,39 +387,313 @@ func (lv *lockableVulnerability) appendFunc(metadataKey string, metadata interfa
...
 	return response
 }
+
+func findLock(datastore database.Datastore, updaterLockName string) (string, time.Time, bool, error) {
+	tx, err := datastore.Begin()
+	if err != nil {
+		log.WithError(err).Error()
+	}
+	defer tx.Rollback()
+	return tx.FindLock(updaterLockName)
```
jzelinskie, line 474: this is missing a commit?

updater.go
```diff
@@ -349,39 +387,313 @@ func (lv *lockableVulnerability) appendFunc(metadataKey string, metadata interfa
...
+	log.WithField("count", len(changes)).Debug("creating vulnerability notifications...")
+	if len(changes) == 0 {
+		return nil
+	}
+
+	tx, err := datastore.Begin()
+	if err != nil {
+		return err
+	}
+
```
jzelinskie: why not defer rollback?

updater.go
```diff
@@ -349,39 +387,313 @@ func (lv *lockableVulnerability) appendFunc(metadataKey string, metadata interfa
...
+			Name:      vuln.Name,
+			Namespace: vuln.Namespace.Name,
+		})
+	}
+
+	tx, err := datastore.Begin()
+	if err != nil {
+		return nil, err
+	}
+
```
jzelinskie, line 640: why not defer rollback?

worker.go
```diff
@@ -44,177 +43,458 @@ var (
...
 	}
 
-	if imageFormat == "" {
-		return commonerr.NewBadRequestError("could not process a layer which does not have a format")
+	if err := commonerr.CombineErrors(errs...); err != nil {
+		return nil, nil, nil, err
 	}
 
-	log.WithFields(log.Fields{logLayerName: name, "path": cleanURL(path), "engine version": Version, "parent layer": parentName, "format": imageFormat}).Debug("processing layer")
+	// TODO: to fix, when there are duplicated layers
```
jzelinskie: hmm

worker.go
```diff
@@ -44,177 +43,458 @@ var (
...
-// detectNamespaces returns a list of unique namespaces detected in a layer and its ancestry.
-func detectNamespaces(name string, files tarutil.FilesMap, parent *database.Layer) (namespaces []database.Namespace, err error) {
-	nsSet := map[string]*database.Namespace{}
-	nsCurrent, err := featurens.Detect(files)
+func processAncestry(datastore database.Datastore, name string, layers []database.LayerWithContent) error {
+	tx, err := datastore.Begin()
 	if err != nil {
-		return
+		return err
 	}
```
jzelinskie: defer rollback?

worker.go
```diff
@@ -44,177 +43,458 @@ var (
...
-			return features, err
+// computeAncestryFeatures computes the features in an ancestry based on all
+// layers.
+func computeAncestryFeatures(ancestryLayers []database.LayerWithContent) ([]database.NamespacedFeature, error) {
+	// version format -> namespace
+	namespaces := map[string]database.Namespace{}
+	// version format -> feature ID -> feature
+	features := map[string]map[string]database.NamespacedFeature{}
+	for _, layer := range ancestryLayers {
+		// at start of the loop, namespaces and features always contain the previous
```
jzelinskie: capital at the start of sentences, end with periods, lol grammar

worker_test.go
```diff
@@ -15,61 +15,324 @@
...
-	"github.com/coreos/clair/database"
-	"github.com/coreos/clair/ext/versionfmt/dpkg"
-	"github.com/coreos/clair/pkg/commonerr"
-
 	// Register the required detectors.
+	"github.com/coreos/clair/database"
+	"github.com/coreos/clair/ext/featurefmt"
 	_ "github.com/coreos/clair/ext/featurefmt/dpkg"
+	_ "github.com/coreos/clair/ext/featurefmt/rpm"
+	"github.com/coreos/clair/ext/featurens"
```
jzelinskie: pull these up to where they used to be

jzelinskie (changes requested):
oh boy, here we go

database/models.go
KeyboardNerd: ok, so in that case the implementation should initialize an encrypted version of
 VulnerabilityNotificationFirstPage and VulnerabilityNotificationLastPage 
when it's registered. And the pagination key should be kept being used only inside pgsql.

ext/vulnsrc/debian/debian.go
KeyboardNerd: Because the transaction won't mutate the database

worker.go
KeyboardNerd: actually it's not true! the layers are deduplicated before the function call

jzelinskie:
I think we should rename the API package to just "api" or "v3" (because Clair v3) in order to avoid confusion.

After that I think this is safe to merge!

jzelinskie (approved):

PR #442, opened 2017-08-06
Adding 'clair-scanner' cli tool as one of integration tools

arminc:
I have created a spin-off from 'analyze-local-images' before it got deprecated, called clair-scanner. Clair-scanner uses Clair to check for vulnerabilities and blocks when there is one. But it also has an option to whitelist vulnerabilities which allows Clair to be used in a straight trough process as part of CI/CD pipeline. 

Although clair-scanner is derived from 'analyze-local-images' the code is different and is maintained.

### DIFF
--- a/Documentation/integrations.md
+++ b/Documentation/integrations.md
@@ -22,6 +22,6 @@ This document tracks projects that integrate with Clair. [Join the community](ht
 
 [Portus](http://port.us.org/features/6_security_scanning.html#coreos-clair): an authorization service and frontend for Docker registry (v2).
 
-[clair-scanner](https://github.com/arminc/clair-scanner): a spin-off from 'analyze-local-images' who blocks on vulnerabilities with whitelist possibility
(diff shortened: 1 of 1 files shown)


Documentation/integrations.md
```diff
@@ -22,4 +22,6 @@ This document tracks projects that integrate with Clair. [Join the community](ht
 
 [Portus](http://port.us.org/features/6_security_scanning.html#coreos-clair): an authorization service and frontend for Docker registry (v2).
 
+[clair-scanner](https://github.com/arminc/clair-scanner): a spin-off from 'analyze-local-images' who blocks on vulnerabilities with whitelist possibility
```
jzelinskie: This sentence confused me for a bit. How about something along the lines of:

>a project similar to 'analyze-local-images' with a whitelisting feature

jzelinskie (changes requested):

Documentation/integrations.md
arminc: I had that but thought it's not clear that it blocks... I agree this is much simpler

jzelinskie (approved):
LGTM! Thanks for this project :D

PR #444, opened 2017-08-17
spring cleaning

jzelinskie:
this change
* creates full articles of documentation for things that were previously just in the README
* moves compose and grafana into contrib/
* renames the example config to match etcd's repository
* regenerates our bill of materials


Quentin-M (approved):
does it compile?

PR #447, opened 2017-08-18
API: change api port to api addr, rename RunV2 to Run.

### LINKED ISSUE #446: Problem running clair as a container in docker swarm
I am trying to run clair as part of a docker swarm stack which runs clair, postgres and clairctl as services which can talk to each other over an overlay network.
The services all appear to run ok. The clair logs seems to show that it is running and I can see that the postgres database is populated. Requests to http://clair:6061/health from the clairctl container return HTTP 200.
But requests to http://clair:6060 are refused even though the clair container is exposing ports 6060 and 6061. 
If I do "docker exec -it <container_id> sh" to see what's going on inside the clair container, I can see that it is listening on :::6060 for the health request but listening on localhost:6061 for the API. 
Looking at the clair code in api/v3/server.go, I see this:

`func Run(GrpcPort int, tlsConfig *tls.Config, CertFile, KeyFile string, store database.Datastore) {
	l, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", GrpcPort))`

i.e. the clair API is explicitly listening on localhost, and there's no way to override that.
All this leads me to think that what I'm trying to do will never work - is that the case?

Chris

KeyboardNerd:
Fixes #446


api/v3/server.go
```diff
@@ -149,7 +149,7 @@ func servePrometheus(mux *http.ServeMux) {
 
 // Run initializes grpc and grpc gateway api services on the same port
 func Run(GrpcPort int, tlsConfig *tls.Config, CertFile, KeyFile string, store database.Datastore) {
-	l, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", GrpcPort))
+	l, err := net.Listen("tcp", fmt.Sprintf(":%d", GrpcPort))
```
jzelinskie: Can we pass the whole addr here rather than just the port?

jzelinskie (changes requested):

config.example.yaml
```diff
@@ -25,18 +25,19 @@ clair:
...
-    # v3 grpc/RESTful API server port
-    grpcport : 6060
+    # v3 grpc/RESTful API server address
+    addr: 0.0.0.0:6060
 
-    # Health server port
+    # Health server address
     # This is an unencrypted endpoint useful for load balancers to check to healthiness of the clair server.
-    healthport: 6061
+    healthaddr: 0.0.0.0:6061
```
jzelinskie: wrap these in quotes, just in case it confuses the yaml

```yaml
addr: "0.0.0.0:6060"
```

jzelinskie (changes requested):

jzelinskie (approved):

PR #450, opened 2017-08-21
database/pgsql: move token lib

jzelinskie:
This is no longer used in the API, but encapsulated by the DB driver so that they can implement pagination in alternative ways.


ecordell (approved):
LGTM

PR #454, opened 2017-08-24
adding insecure TLS to Helm chart

InTheCloudDan:
This adds the ability to set allow insecure TLS on the container.


jzelinskie (approved):
LGTM! Thanks!

Issue #401, opened 2017-05-24
Clair should support loading data without internet access


jzelinskie:
I'm opening this thread to open the floor for discussion on how this should be designed.

alekssaul:
Thinking out loud here, can the datasource locations be parameterized so that users can seed it via hosting them on an http endpoint?

yixingjia:
Some thoughts on how to ship the data: 

1: fully data. 

a>Clair ship the full upto date postgresql database from time to time  (eg, all.sql.gz, best by pg_dump)
b>: when user fresh install clair. he need to specify a volume  like 
-v ./data:/docker-entrypoint-initdb.d
and put the all.sql.gz file to the ./data directory.

Thus when clair up it will have the latest data

2> increment update 
a> Clari dump the latest vulnerability data from the PostgreSQL.
eg: pg_dump -t feature -t featureversion -t namespace -t vulnerability -t vulnerability_affects_featureversion -t vulnerability_fixedin_feature -U postgres > data.sql
b> user can download this data.sql
c> user can put the data.sql to the ./data directory 
d> user login to the postgresql container and exec >psql -U postgres < data.sql

This may require the user make a fully re-scan after the increment update.

furuholm:
Agree with @alekssaul. I have patched the source code to fetch data from my own web server once or twice to support offline mode. The only change I had to do was to change the urls of the CVE sources.

rikkuness:
Database dumping sort of works for me, just have some issues with getting the right format of dump that will import to the other database without destroying any relations on the offline running instance.

I'm dumping with `pg_dump -Fc -t keyvalue -t namespace -t vulnerability -t vulnerability_affected_feature -U postgres -f /tmp/clair-data.pg` and I can drop that onto a clean database and Clair running with update disabled can use that data.

The issue I'm having with it is doing a subsequent updates to the database and it violating constraints, even doing the dump with the `-c` option isn't helping me much.

Are there more or less tables that should be replicated? Ideally I'd like to keep the other ancestry data in the offline database intact and just update the vulnerability data.

rikkuness:
On a clean Clair database I can restore the tables as seen
```
# pg_restore -1 -U postgres -d postgres clair-data.pg
# echo $?
0
```

If I then try run again straight after
```
# pg_restore -1 -U postgres -d postgres clair-data.pg
pg_restore: [archiver (db)] Error while PROCESSING TOC:
[...]

PR #460, opened 2017-08-31
Make source of vulnerability data configurable

furuholm:
This is a proposal for solving #401. I've added a new method to the Updater interface called SetSourceUrl which is invoked if the config file contains relevant source urls.

I've tried to stick to the conventions but there were a couple of things that I was not sure what to do with:

1. I have added SourceURLs as a var in updater.go. Another option would be to pass the config to fetch and use config.SourceURLs instead of SourceURLs in fetch. This would be more explicit, so I think I would prefer doing it this way, but I tried to stick to the conventions as much as possible...
2. Should I add all default urls to the sample config? I haven't but reading the comments in config.yaml.sample made me consider doing so.

### DIFF
--- a/cmd/clair/config.go
+++ b/cmd/clair/config.go
@@ -61,6 +61,7 @@ func DefaultConfig() Config {
 		Updater: &clair.UpdaterConfig{
 			EnabledUpdaters: vulnsrc.ListUpdaters(),
 			Interval:        1 * time.Hour,
+			SourceURLs:      make(map[string]string),
 		},
 		Worker: &clair.WorkerConfig{
 			EnabledDetectors: featurens.ListDetectors(),
--- a/cmd/clair/main.go
+++ b/cmd/clair/main.go
@@ -124,6 +124,8 @@ func configClairVersion(config *Config) {
 	}
 
 	clair.EnabledUpdaters = strutil.CompareStringListsInBoth(config.Updater.EnabledUpdaters, updaters)
+
+	clair.SourceURLs = config.Updater.SourceURLs
 }
 
 // Boot starts Clair instance with the provided config.
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -68,12 +68,21 @@ clair:
     # Frequency the database will be updated with vulnerabilities from the default data sources
     # The value 0 disables the updater entirely.
     interval: 2h
+
     enabledupdaters: 
       - debian
       - ubuntu
       - rhel
       - oracle
       - alpine
+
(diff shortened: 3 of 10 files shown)


jzelinskie:
Hey, thanks for this PR.

I have a bunch of thoughts:
- If we're going to do something like this, I'd like dedicated sections to each vulnsrc where they can be enabled or have this source specified
- This makes the assumption that all of these upstream vulnsrcs can be mirrored
- In addition, we could build a tool that pulls down all the new vulnerabilities, and then lets you take that behind the firewall and then Clair runs an update against that neutral format.

Please add your new stuff to the sample config; I'm going to think about how to make the config file's schema make more sense with this new functionality.

Also, gofmt your code ;)

furuholm:
Great comments!

-  Your restructuring ideas to the config file's schema makes sense.
-  I'm actually less concerned about the assumption that all of these upstream vulnsrcs can be mirrored (there **could** be a mirror for any of these, even if you couldn't set one up for yourself) and more concerned about the possibility of new updaters whose source can't be described in a URL. Then the interface would be kind of broken.
-  A tool for pulling the data in a neutral fashion sounds like a great idea. Both for complete offline mode and during development as I find some of the sources to be somewhat unreliable. Having your own "proxy" source would solve this problem. Would this be a command line tool or a file that you can serve to clair? Or both?

Have added the sources to the config file and fixed the formatting.

Cheers!

jzelinskie:
Hey, sorry for letting this sit.

>concerned about the possibility of new updaters whose source can't be described in a URL. Then the interface would be kind of broken.

Totally agree here. Thus the hesitation to merge. Besides offline, if there isn't any other reason to make this configurable and risk having a brittle interface and config, I think this PR is ready to be closed.

>Would this be a command line tool or a file that you can serve to clair? Or both?

Ideally a tool that dumps an update file that can be carried offline and applied to an offline cluster.
The new database schema in Clair v3 (aka the master branch) theoretically should enable the capability to copy data from an online copy of Clair (or a tool that just ran the Clair updater). There is definitely some more thought that needs to go into this; since vulnerabilities and other data may be inserted in a different order,  IDs probably wouldn't match across databases. It's unlikely that the tool could be as naive as a pg_dump on a few tables, but maybe we can get something close to that.

furuholm:
A pg_dump sounds like a good input source. In general I don't like having external tools making changes to the database though. There is too much that can go wrong as the database evolves. How about a vulnsrc extension that imports pg_dumps?

You can close the PR. Let's continue the discussion in #401.

PR #463, opened 2017-09-06
contrib/helm/clair: fix the ingress template

bmcustodio:
Make `servicePort` reference `externalApiPort` instead of `externalPort`.


jzelinskie (approved):
nice catch! thanks

PR #467, opened 2017-09-12
Adding postgres to the chart so it can be used without dependencies.

grebois:


contrib/helm/clair/templates/configmap.yaml
```diff
@@ -13,7 +13,9 @@ data:
         options:
           # PostgreSQL Connection string
           # https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
-          source: "{{ .Values.config.postgresURI }}"
+          # This should be done using secrets or Vault, but for now this will also work
+
```
jzelinskie: no need for this blank new line

contrib/helm/clair/values.yaml
```diff
@@ -56,3 +55,16 @@ config:
   - apk
   - dpkg
   - rpm
+## Configuration values for the postgresql dependency.
+## ref: https://github.com/kubernetes/charts/blob/master/stable/postgresql/README.md
+##
+postgresql:
+  cpu: 1000m
+  memory: 1Gi
+## Harcording for now till https://github.com/kubernetes/helm/issues/2196
```
jzelinskie:
```yaml
# These values are hardcoded until Helm supports secrets.
# For more info see: https://github.com/kubernetes/helm/issues/2196
```

contrib/helm/clair/values.yaml
```diff
@@ -36,7 +36,6 @@ resources:
     cpu: 100m
     memory: 128Mi
 config:
-  postgresURI: "postgres://user:password@host:5432/postgres?sslmode=disable"
```
jzelinskie, line 39: Just comment this out in case someone wants to bring their own postgres

contrib/helm/clair/templates/configmap.yaml
```diff
@@ -13,7 +13,9 @@ data:
...
+          source: "postgres://{{ .Values.postgresql.postgresUser }}:{{ .Values.postgresql.postgresPassword }}@{{ template "postgresql.fullname" . }}:5432/{{ .Values.postgresql.postgresDatabase }}?sslmode=disable"
```
jzelinskie: This should check if a postgresURI is explicitly set, if not default to using the postgresql dependency.

contrib/helm/clair/values.yaml
```diff
@@ -56,3 +55,16 @@ config:
   - apk
   - dpkg
   - rpm
+## Configuration values for the postgresql dependency.
```
jzelinskie: single `#` for comments

jzelinskie (changes requested):
Thanks so much for this PR!
☸️

grebois:
Many thanks @jzelinskie, fixed all your comments.

Btw, I've deployed this chart and run some tests with [Klar](https://github.com/optiopay/klar) but could not get it running, not sure yet if the problem is on their side or with this particular implementation; https://github.com/optiopay/klar/issues/55

jzelinskie (approved):
LGTM! Awesome!

jzelinskie:
To my knowledge, Klar doesn't yet work with the unreleased API changes on the master branch of Clair.

grebois:
@jzelinskie I see, any suggestion of similar tool to scan local images?

jzelinskie:
clairctl, but only for a stable release of Clair (2.0.x).
To my knowledge no local scanning tools are written yet for the unstable v3 gRPC API.

grebois:
Would it make sense then to set the chart default to quay.io/coreos/clair ? instead of clair-git?

jzelinskie:
I don't think so because this chart is on the master branch.
I don't mind backporting it to the v2 branch with the changed image and configmap, though.

grebois:
Ok, so I changed the image to quay.io/coreos/clair and now I get:

```
2017-09-13 13:38:17.683622 C | main: failed to load configuration: could not load configuration: no database source specified
```

yet the config for v2.0.1 looks compatible;

https://github.com/coreos/clair/blob/v2.0.1/contrib/k8s/config.yaml

removing this 2 lines 

https://github.com/coreos/clair/blob/v2.0.1/contrib/k8s/config.yaml#L19-L20

makes it work, so I assume is quay.io/coreos/clair:latest is based on v.2.0.0

If yes, could you update it to v2.0.1?

jzelinskie:
Oh wow! Yup, `latest` was pointing to v1.2.6! I fixed it so it's [pointing to v2.0.1 now](https://quay.io/repository/coreos/clair?tab=tags).
Thanks

grebois:
Looks great now, thanks!

contrib/helm/clair/templates/configmap.yaml
nabadger: I'm finding that if I specify postgresURI in my helm values, then the template produces an invalid configmap.

The source option containing the connection string gets merged with the comment above it (I can fix it by removing one of the whitespace chomping dashes).

PR #469, opened 2017-09-20
Fix config.yaml.sample

zamarrowski:
Changed address of Clair API to bind correctly.

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -33,7 +33,7 @@ clair:
 
   api:
     # v3 grpc/RESTful API server address
-    addr: "127.0.0.1:6060"
+    addr: "0.0.0.0:6060"
 
     # Health server address
     # This is an unencrypted endpoint useful for load balancers to check to healthiness of the clair server.


jzelinskie:
Binding to 0.0.0.0 actually is a shortcut for saying "bind on all interfaces".
If a user is just trying to get something running it's more likely to function than localhost, especially if the user is trying to access the instance from a non-local network (or a virtual network that runs, say, docker).

I'm willing to accept this change if you have a good argument against this.

zamarrowski:
Yeah, when you bind to 0.0.0.0 you are saying "bind on all interfaces" inside container. In fact, the Docker documentation says: 

> You can see that Docker has exposed these container ports on 0.0.0.0, the wildcard IP address that will match any possible incoming port on the host machine. 

We needed change to 0.0.0.0 because we couldn't access to the API from outside of container.

jzelinskie (approved):
Oh! I totally read the diff wrong! I thought we already were using `0.0.0.0`.
LGTM

Issue #441, opened 2017-08-03
git-remote-http zombies


sweharris:
A `ps` listing on my machine showed a lot of zombie processes, all children of the main clair process

eg

```
root     15236 15220  1 Jul25 ?        02:26:41 /clair -config=/config/config.yaml
root     15881 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     15947 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     16903 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     17402 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     18491 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     18731 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     19556 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     19714 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
root     20534 15236  0 Aug01 ?        00:00:00 [git-remote-http] <defunct>
```

They're all dated "Aug01", starting from 8:15 through to 13:15 UTC, pretty much 2 every 10 minutes.

There _might_ have been some networking issues around that time; I wasn't paying attention (this server is behind at least 2 layers of proxy).

If it helps identify the version
````
quay.io/coreos/clair-git   latest              ee6cf4643b22        2 weeks ago         422 MB

    {
        "Id": "sha256:ee6cf4643b22aba6dfcb18c653ae60b13f46f9a747b4226a69090712b2510f26",
        "RepoTags": [
            "quay.io/coreos/clair-git:latest"
        ],
        "RepoDigests": [
            "quay.io/coreos/clair-git@sha256:0a9b504a4e174aec9a5e3d7ca239ed58aaca0c282226b845a61e4b9978b6131c"
        ],
   
````

bseb:
+1 on this Issue

I have seen this issue caused by a network interruption and as well firewall misconfiguration. It seems that an http timeout retrieving vulnerability  data results in these processes, and they do not clear until the Clair container is restarted.

jzelinskie:
Thanks for the report. I have some other projects that shell out to git that have had similar issues in the past. I'll see if I can share that code with this project to make our handling of git more robust.

bseb:
[...]

PR #477, opened 2017-10-13
Add dumb-init to Dockerfile Entrypoint

bseb:
This allows for proper zombie process reaping. See discussion in https://github.com/coreos/clair/issues/441


jzelinskie:
It looks like [Alpine has a package for dumb-init](https://pkgs.alpinelinux.org/package/edge/community/x86_64/dumb-init) that we could use instead of curling it down. This URL looks like it's for edge, though. I'm not sure if it's available on stable.

bseb:
@jzelinskie Looks like dumb-init is indeed available in the alpine repos for 1.8. Ive changed the Dockerfile accordingly.

jzelinskie (approved):
LGTM! Thanks!

PR #482, opened 2017-10-30
Added my new python project Yair

yfoelling:
Please have a look at my project.
I think many developers may want to use this tool to scan their image.
Operation teams may use it to decide if an image is secure _enough_ to be deployed etc.

It is much easier to use than the alternatives and gives you much better output.


jzelinskie (approved):
LGTM! Thanks for building such a cool project.

Quentin-M:
Dope!

PR #484, opened 2017-11-07
Usage of running clair via docker is wrongly written.

odg0318:
Command line needs --net=host when docker is used to run clair.

### DIFF
--- a/Documentation/running-clair.md
+++ b/Documentation/running-clair.md
@@ -77,7 +77,7 @@ If this error is raised, manually execute `docker-compose start clair`.
 $ mkdir $PWD/clair_config
 $ curl -L https://raw.githubusercontent.com/coreos/clair/master/config.yaml.sample -o $PWD/clair_config/config.yaml
 $ docker run -d -e POSTGRES_PASSWORD="" -p 5432:5432 postgres:9.6
-$ docker run -d -p 6060-6061:6060-6061 -v $PWD/clair_config:/config quay.io/coreos/clair-git:latest -config=/config/config.yaml
+$ docker run --net=host -d -p 6060-6061:6060-6061 -v $PWD/clair_config:/config quay.io/coreos/clair-git:latest -config=/config/config.yaml
 ```
 
 #### Source


jzelinskie:
`net=host` is an optional flag if you want to run Clair using the hosts networking namespace. It is not required for executing Clair, although I could be convinced to document this if we find users are confused about this functionality.

odg0318:
@jzelinskie I mean that in sample configuration localhost is used to connect to postgres. But when container is running in bridge network, there is no way to access localhost. I think that the document about running clair via docker works wrong.

jzelinskie:
Ah! It turns out that you can elide `net=host` on macOS. I'm fine with adding this in to correct the instructions for Linux users.

Sorry about that.

jzelinskie (approved):

Documentation/running-clair.md
```diff
@@ -77,7 +77,7 @@ If this error is raised, manually execute `docker-compose start clair`.
 $ mkdir $PWD/clair_config
 $ curl -L https://raw.githubusercontent.com/coreos/clair/master/config.yaml.sample -o $PWD/clair_config/config.yaml
 $ docker run -d -e POSTGRES_PASSWORD="" -p 5432:5432 postgres:9.6
```
jzelinskie, line 79: don't we need to add net=host for postgres too?

jzelinskie (changes requested):

Documentation/running-clair.md
odg0318: You are right

odg0318: Nope, postgres container publishes 5432 port to host. It is shown to host as localhost at the side of clair container which is running in host network.

jzelinskie: Ok. I'll approve this change if you can fix the spelling mistake in the commit message.

`Usage of Rrunning clair via docker is wrongly written.` => `Usage of running clair via docker is wrongly written.`

Thanks! It's been a long time since I've used Docker directly on Linux (I use mac) and so I've forgotten the semantics of net=host.

odg0318: I fixed my spelling mistake and changed author to me.

jzelinskie (approved):
Thank you!

PR #485, opened 2017-11-07
Driver: Add proxy support

yebinama:
I have to use a proxy on my servers running Clair.

Setting HTTP_PROXY variable works fine for the updater since http.Get and then DefaultTransport is used.

```
var DefaultTransport RoundTripper = &Transport{
        Proxy: ProxyFromEnvironment,
        ...
}
```

On the other hand, as the imagefmt driver declares a custom http.Transport to enable/disable certificate verification, it does not support HTTP_PROXY. I modified it to have the same behavior as the updater.


jzelinskie (approved):

jzelinskie:
Awesome thanks for the contribution!

PR #488, opened 2017-11-21
Fix path name for values.yaml in helm chart

caulagi:
Fix path name for values.yaml in helm chart


jzelinskie (approved):
LGTM! Thanks for the fix!

PR #494, opened 2017-12-06
Add Idle Connection Limits to NVD fetch

ErikThoreson:
nvd.go and xml.go: added Max idle connection limits and publisher datetime from xml feed

Reason:
The NVD site frequently timeouts connections when downloading the feeds this causes an issue when running clair behind a load balancer or proxy that attempts to re-establish the connection and which can result in a stacking of idle connections. The NVD site will then reset the connection due to reaching their set limit.

Example Fetter Error that would be resolved by the limits:

NOTE: If this is accepted. I'd love it if this could be put in as a patch release for the 2.0 version as well.


ErikThoreson:
Actually - I found another issue with the fetcher. Going to put in a fix for that and create a new pr instead.

PR #496, opened 2017-12-08
 adding publisher datetime and updating nvd feed download

ErikThoreson:
nvd.go and xml.go: added publisher datetime from xml feed and updated the XML feed download to complete before moving onto downloading the next feed

Reason:
The NVD site frequently resets connections when downloading the feeds if they are not downloaded in a timely manner. When all the feeds are requested at once but then not downloaded immediately the NVD site will reset the connection. It is more reliable to download each after making the request instead of opening all the connections then downloading them one at a time. 

Example feed error:

NOTE: If this is accepted. I'd love it if this could be put in as a patch release for the 2.0 version as well.


ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -79,30 +79,38 @@ func (a *appender) BuildCache(datastore database.Datastore) error {
 		a.dataFeedHashes = make(map[string]string)
 	}
 
-	// Get data feeds.
+	//Get data feeds.
```
jzelinskie: Not sure why you changed this line.

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -79,30 +79,38 @@ func (a *appender) BuildCache(datastore database.Datastore) error {
...
 
 	// Parse data feeds.
-	for dataFeedName, dataFeedReader := range dataFeedReaders {
-		var nvd nvd
-		if err = xml.NewDecoder(dataFeedReader).Decode(&nvd); err != nil {
-			log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not decode NVD data feed")
-			return commonerr.ErrCouldNotParse
-		}
+	for dataFeedName, dataFileName := range dataFeedReaders {
+
```
jzelinskie: no need for a new line here

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -79,30 +79,38 @@ func (a *appender) BuildCache(datastore database.Datastore) error {
...
+		if f, err := os.Open(dataFileName); err == nil {
+			//add check here
```
jzelinskie: this comment isn't particularly valuable

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -79,30 +79,38 @@ func (a *appender) BuildCache(datastore database.Datastore) error {
...
+			var nvd nvd
+			r := bufio.NewReader(f)
```
jzelinskie: This, among a few other changes you made, prevents us from streaming reads/writes and rather buffers them into memory. This will increase memory usage. While this doesn't consume a lot of memory, I'd rather we see if we can maintain the streaming property.

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -143,51 +151,40 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
 		dataFeedHashes[dataFeedName] = hash
 	}
 
-	// Create io.Reader for every data feed.
-	dataFeedReaders := make(map[string]NestedReadCloser)
+	// Create map containg the name and filename for every data feed.
```
jzelinskie: containing

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -143,51 +151,40 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
...
 				return dataFeedReaders, dataFeedHashes, commonerr.ErrCouldNotDownload
 			}
 
 			// Store it to a file at the same time if possible.
-			if f, err := os.Create(fileName); err == nil {
-				nrc := NestedReadCloser{
-					Reader:            io.TeeReader(gr, f),
-					NestedReadClosers: []io.ReadCloser{r.Body, gr, f},
+			f, err := os.Create(fileName)
+			defer f.Close()
```
jzelinskie: This will leave these files open until the function returns, rather than between loops.

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -143,51 +151,40 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
...
+			if err == nil {
+				_, err = f.Write(b)
```
jzelinskie: ioutil.WriteFile is probably better to use than os.Create with this conditional

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -217,6 +213,39 @@ func getHashFromMetaURL(metaURL string) (string, error) {
 	return "", errors.New("invalid .meta file format")
 }
 
+func getBytesFromFeedURL(feedURL string) ([]byte, error) {
+
```
jzelinskie: no new line

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -217,6 +213,39 @@ func getHashFromMetaURL(metaURL string) (string, error) {
...
+		return nil, err
+	}
+
+	req.Header.Add("Accept-Encoding", "gzip")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
```
jzelinskie: no new line

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -217,6 +213,39 @@ func getHashFromMetaURL(metaURL string) (string, error) {
...
+	defer resp.Body.Close()
+
+	reader, err := gzip.NewReader(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	b, err := ioutil.ReadAll(reader)
```
jzelinskie: This is what is preventing us from streaming this data.

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -217,6 +213,39 @@ func getHashFromMetaURL(metaURL string) (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
jzelinskie: You just checked this error twice

jzelinskie (changes requested):
There's two major issues I see with this PR:

1. Rather than us streaming the data as we deserialize and write to disk, this buffers it directly into memory. Performance is still unlikely to be a big deal and I'm okay with this change, if it makes the code more simple.

2. I don't know if NVD's website will rate-limit or block Clair instances if it make too many requests, rather than one that downloads in bulk (this is probably the reason why they make the bulk download available).

ErikThoreson:
@jzelinskie thanks for the feedback. I've corrected what I can now. I'll look into using streaming instead of buffering items into memory. If I can still do download them sequentially instead of all at once, the NVD site hopefully won't complain at me.

jzelinskie:
Yup. When this PR is ready, I'd like to see it rebased on master so that there are no longer any merge commits, too.

ErikThoreson:
@jzelinskie moved to io.Copy which streams the file in small chunks to the file. I put back most of the original code and rebased as well.

jzelinskie:
This change looks good, but and I'm sorry for the wait. I'm going to email someone at NVD to confirm that this isn't going to cause Clair installations to get rate-limited and fail updates.

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -87,22 +88,27 @@ func (a *appender) BuildCache(datastore database.Datastore) error {
...
 
 	// Parse data feeds.
-	for dataFeedName, dataFeedReader := range dataFeedReaders {
-		var nvd nvd
-		if err = xml.NewDecoder(dataFeedReader).Decode(&nvd); err != nil {
-			log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not decode NVD data feed")
-			return commonerr.ErrCouldNotParse
-		}
+	for dataFeedName, dataFileName := range dataFeedReaders {
+		if f, err := os.Open(dataFileName); err == nil {
```
jzelinskie: Can we invert this?

```go
f, err := os.Open()
if err != nil {
  log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not open NVD data file")
  return commonerr.ErrCouldNotParse
}

var nvd nvd
...
```

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -87,22 +88,27 @@ func (a *appender) BuildCache(datastore database.Datastore) error {
...
+			var nvd nvd
+			r := bufio.NewReader(f)
+			if err = xml.NewDecoder(r).Decode(&nvd); err != nil {
+				log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not decode NVD data feed")
+				return commonerr.ErrCouldNotParse
```
jzelinskie: we need another f.Close() above this return

jzelinskie (changes requested):
Sorry for taking forever. With these small changes, I'll actually merge the thing, haha.

ErikThoreson:
cool. i'm testing the updates. hopefully will re-submit soon.

ErikThoreson:
i'm going to close this and open a new one.

PR #513, opened 2018-01-23
Improve documentation on helm installation

lrepolho:
The docs were missing the helm dependencies download before running the install, so Postgres was never installed.

Also I'm suggesting the user to download the latest version of helm since executing it with 2.5 failed for me, now with 2.7.2 it works fine.


jzelinskie (approved):
thanks!

PR #522, opened 2018-02-09
Improvements to the helm chart

vdboor:
With this PR I'd like to fix the hurdles I've run into.

These are:
* one bug: the `source: postgresURI ` line was merged the the comment above, because the if statement strips newlines at both ends.
* the memory limits are too low, causing Clair to be scheduled on hosts that doesn't have enough memory available - and eventually OOM the container or machine.
* Switch to the stable release, as the git version doesn't work out of the box.
* Explicitly mention that `postgresql.enabled` can be set to false, because the `requirements.yaml` allows that.

Do you also have plans to move this chart to the official helm chart repositories?

### DIFF
--- a/contrib/helm/clair/values.yaml
+++ b/contrib/helm/clair/values.yaml
@@ -34,7 +34,7 @@ resources:
     memory: 1Gi
   requests:
     cpu: 100m
-    memory: 128Mi
+    memory: 500Mi
 config:
   # postgresURI: "postgres://user:password@host:5432/postgres?sslmode=disable"
   paginationKey: "XxoPtCUzrUv4JV5dS+yQ+MdW7yLEJnRMwigVY/bpgtQ="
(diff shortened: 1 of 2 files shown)


contrib/helm/clair/values.yaml
```diff
@@ -5,9 +5,9 @@ replicaCount: 1
 logLevel: info
 insecureTls: false
 image:
-  repository: quay.io/coreos/clair-git
-  tag: latest
-  pullPolicy: Always
+  repository: quay.io/coreos/clair
+  tag: v2.0.1
```
jzelinskie: Because this code is in master, it should be tied to deploying master. When we cut a stable release, we should tie this to a particular version.

jzelinskie (changes requested):
Just one minor nitpick.
Thanks a bunch for giving this some love.

vdboor:
Thanks for looking into this. I've updated my pull request to omit that image change!

I do think it might be useful to mention this choice in the docs, so people can choose to install the stable version instead. I didn't get the master version running last week.

jzelinskie (approved):
Agreed. Can you file an issue?

PR #528, opened 2018-03-07
Helm: Fixed a typo in maintainers field.

KeyboardNerd:
Maintainers field was incorrectly listing email as one item.


jzelinskie (approved):
Nice catch! Thanks!

PR #532, opened 2018-03-20
README: fixed issues address

KeyboardNerd:
Is this like a inside joke? ;)


jzelinskie (approved):
haha! just copypasta. thanks sida

PR #538, opened 2018-04-04
generate protobufs in docker

jzelinskie:
This removes the requirement of the build machine having protobuf and
any extensions installed.


bison (approved):
:cool:

PR #541, opened 2018-04-11
[WIP] Arch linux support

hhomar:
Hi

I've added image scanning for Arch Linux, but I've run into a couple of problems. I've put the problems in as separate commits:

1. When versionfmt.InRange is run and the versionA and rangeB are the same, InRange will return false e.g.

VersionA = 5.4.2-1, RangeB = 5.4.2-1 returns false, which means that the vulnerability is not included in the affected namespaced feature.

This doesn't seem right to me, although I could be misunderstanding the idea behind it?

2. The bigger problem is that Arch Linux stores installed packages as directories under /var/lib/pacman/local as opposed to a single file like Debian. This causes an issue when processing the layers as it will only return the features found in the last layer.

I've put in a hack to append the current features from each layer, which fixes the problem for Arch Linux, but it obviously breaks other image scanners.

I'm a bit stuck on how to process.

Thanks


jzelinskie:
1. Why not define a new versionfmt for Arch version strings that works the way they should. It looks like [there's a program](https://www.archlinux.org/pacman/vercmp.8.html) for comparing version strings explicitly for Arch.
2. This is actually desirable. We'll aggregate the features together at the Ancestry level. Because layers can be reused across different Ancestries, you actually don't want information from any previous layers. This is actually the big difference between Clair v2 and v3.

jzelinskie:
@hhomar any updates?

PR #568, opened 2018-07-05
Updating the ubuntu vulnsrc to use the new git repo

MackJM:
Hi,
Since ubuntu has officially moved their cve-tracker to use git instead of bzr. I've updated the vulnsrc for ubuntu to use git. A lot of the code was copied from Alpine.

I've built and tested locally and it all runs fine.

Cheers,

### DIFF
--- a/ext/vulnsrc/ubuntu/ubuntu.go
+++ b/ext/vulnsrc/ubuntu/ubuntu.go
@@ -18,14 +18,12 @@ package ubuntu
 
 import (
 	"bufio"
-	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"regexp"
-	"strconv"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


ext/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -162,120 +160,81 @@ func (u *updater) Clean() {
...
+	if err != nil {
+		return "", vulnsrc.ErrGitFailure
 	}
 
-	return nil
+	commit = strings.TrimSpace(string(out))
+	return
 }
 
 func getRevisionNumber(pathToRepo string) (int, error) {
```
jzelinskie: Why didn't you delete this whole function?

jzelinskie (changes requested):
Everything else looks good!

ext/vulnsrc/ubuntu/ubuntu.go
MackJM: I missed it :disappointed:... I'll remove it

ext/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -162,120 +160,76 @@ func (u *updater) Clean() {
...
+		return "", vulnsrc.ErrGitFailure
 	}
 
-	return revno, nil
+	commit = strings.TrimSpace(string(out))
+	return
 }
 
-func collectModifiedVulnerabilities(revision int, dbRevision, repositoryLocalPath string) (map[string]struct{}, error) {
+func collectModifiedVulnerabilities(revision, dbRevision, repositoryLocalPath string) (map[string]struct{}, error) {
```
jzelinskie: Rename revision "commit" and get rid of dbRevision since we aren't using it anymore.

jzelinskie (changes requested):

jzelinskie (approved):

jzelinskie:
Just got done testing this locally -- thanks sooo much for this PR! I also did the work for the master branch. Bzr has caused so many issues in the past, I'm glad we can finally get rid of it.

MackJM:
Glad I could help. I wasn't sure if I should do this on master or release-2.0... Happy I picked right :smile:

PR #570, opened 2018-07-06
Update to correct repo

zdtsw:
clair:latest is the stable one
  clair-git:latest is buggy to use
  add proxy setting if clair runs behind proxy


jzelinskie:
This is intentional as you're currently viewing HEAD of Clair and not a stable branch.
To install use contrib files for a stable version of Clair, please check out the `release-2.0` branch.

PR #571, opened 2018-07-13
adding publisher datetime and updating nvd feed download

ErikThoreson:
nvd.go and xml.go: added publisher datetime from xml feed and updated the XML feed download to complete before moving onto downloading the next feed

Reason:
The NVD site frequently resets connections when downloading the feeds if they are not downloaded in a timely manner. When all the feeds are requested at once but then not downloaded immediately the NVD site will reset the connection. It is more reliable to download each after making the request instead of opening all the connections then downloading them one at a time.

Example feed error:

NOTE: This is a ref to this PR. Just a bit more cleaned up. https://github.com/coreos/clair/pull/496


jzelinskie (approved):

reasonerjt:
@jzelinskie @ErikThoreson 
Will this be cherry picked to release-2.0 branch?
I'm seeing the connection reset issue when using Clair 2.0.4

PR #572, opened 2018-07-14
use multi-stage build

ghost:
The final image size is 52.2MiB, almost 10 times smaller than it currently is.


jzelinskie (approved):
Thanks! I think we just totally forgot to do this after adding multi-stage builds to Quay!

PR #577, opened 2018-07-18
ext/featurens: Add fedora regex matching

etsauer:
Ensure fedora images can be scanned


jzelinskie:
Does fedora use something different to track vulns or the regular OVAL listings?

etsauer:
@jzelinskie according to [this](https://fedoraproject.org/wiki/Features/OpenSCAP) it looks to be the same.

etsauer:
bump.

jzelinskie:
Hey there, I'm just double checking on whether or not we should be pulling data from here or it's better to get it from another place.

jzelinskie:
So I poked around and got some information from @brianredbeard.
If we're going to detect Fedora, we should also track vulnerabilities from Fedora by using their official database for tracking content, [Bodhi](https://bodhi.fedoraproject.org/updates/?type=security)

[FedMsg](https://fedora-fedmsg.readthedocs.io/en/latest) is also another tool for tracking Fedora content we should explore, although it looks like pure Python.

jam01:
Bump... Can we reopen this? I've built a couple of images based on fedora that'd are failing to scan.

jzelinskie:
This PR needs more than just identifying the Fedora namespace. It also needs to include a vulnsrc for what Fedora is using to track vulnerabilities.

urbaniak:
Any updates on supporting Fedora?

ldelossa:
Fedora is a bleeding edge distribution with no security tracker. Fedora support is a non-goal.

PR #586, opened 2018-08-08
Helm Chart: Use Secret for config file. Fix some minor issues

### LINKED ISSUE #581: Shouldn't configmap.yaml for the Helm chart be a Secret instead? It contains sensitive information
**Description of Problem**:
The ConfigMap for Clair on the Helm chart contains sensitive information of the DB (username and password). Shouldn't it be a Secret instead?

ghost:
- Config file for Clair now is inside a Secret. Fixes #581.

- Update helper template to use latest Helm conventions for "name" and
"fullname".

- Update labels to use the recommended labels by Helm. See:
https://docs.helm.sh/chart_best_practices/#standard-labels

- Remove service.name from values file because it is not used.
Service discovery for parent charts can be made by using the
"clair.fullname" template since template functions are shared
between charts.


jzelinskie (approved):
LGTM! Appreciate the update!

PR #597, opened 2018-08-27
Added vulnerability source URL to the config file

MackJM:
- The source URL for updaters in the vulnerability source can now change through the config
- Changed the array of enabledUpdaters to a map
- When no url is present to the default is still used

### DIFF
--- a/cmd/clair/config.go
+++ b/cmd/clair/config.go
@@ -54,12 +54,19 @@ type Config struct {
 
 // DefaultConfig is a configuration that can be used as a fallback value.
 func DefaultConfig() Config {
+	var updaterMap map[string]string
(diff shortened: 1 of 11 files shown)


MackJM:
I'm working on getting this working in release-2.0 as well

jzelinskie:
I believe there used to be another PR implementing this behavior in the past. I want to dig up that discussion so that I can see if there was anything useful.

MackJM:
@jzelinskie sounds good, I wasn't sure this was the best way of doing it. But it was the best I could come up with at the time.

MackJM:
I'm going to rework this a little better, and add nvd source.

Issue #596, opened 2018-08-27
RHEL doesn't pull any vulnerabilities on master 


MackJM:
<!--

The Clair developers only support the Clair API.
They do not support or operate any third party client (e.g. clairctl, klar, harbor).
If you are using a third party client, please create issues on their respective repositories.

Are you using a development build of Clair (e.g. quay.io/coreos/clair-git)?
Your problem might be solved by switching to a stable release (e.g. quay.io/coreos/clair).

-->

**Description of Problem**: RHEL doesn't pull any vulnerabilities. 
From the logs I see:

`{"Event":"no update","Level":"debug","Location":"rhel.go:170","Time":"2018-08-27 18:31:10.943123","package":"Red Hat"}`

When I double check in the database I see no namespaces for RHEL:

postgres=# SELECT * FROM namespace;                                                                                                                                                                                                                                                        
 id |      name       | version_format 
----+-----------------+----------------
  1 | ubuntu:17.10    | dpkg
  2 | oracle:6        | rpm
  3 | alpine:v3.5     | dpkg
  4 | ubuntu:12.04    | dpkg
  5 | debian:9        | dpkg
  6 | ubuntu:14.04    | dpkg
  7 | ubuntu:15.10    | dpkg
  8 | ubuntu:18.04    | dpkg
  9 | oracle:7        | rpm
 10 | ubuntu:12.10    | dpkg
 11 | alpine:v3.8     | dpkg
 12 | alpine:v3.7     | dpkg
 13 | alpine:v3.3     | dpkg
 14 | debian:10       | dpkg
 15 | debian:8        | dpkg
 16 | ubuntu:16.10    | dpkg
 17 | alpine:v3.4     | dpkg
 18 | oracle:5        | rpm
 19 | alpine:v3.6     | dpkg
 20 | ubuntu:15.04    | dpkg
 21 | ubuntu:16.04    | dpkg
 22 | debian:unstable | dpkg
 23 | ubuntu:13.04    | dpkg
 24 | ubuntu:14.10    | dpkg
 25 | ubuntu:17.04    | dpkg
(25 rows)


**Expected Outcome**: List of vulnerabilities for RHEL/Centos


**Actual Outcome**: No vulnerabilities for RHEL/Centos 


**Environment**:
- Clair version/image: HEAD of master (ddeb64339d6ad59b5bf197459acef5c9a60f53f9)
- Clair client name/version: N/A
- Host OS: Ubuntu 16.04
[...]

PR #600, opened 2018-08-30
Check http code when downloading vulnerabilities data

Allda:
When http requests to Red Hat's oval data doesn't return StatusOK it
returns empty vulnerability list with no error. This should not happen.

This commits check the return code of http request to Red Hat's oval data.

This is also related to #596

### DIFF
--- a/ext/vulnsrc/debian/debian.go
+++ b/ext/vulnsrc/debian/debian.go
@@ -90,6 +90,11 @@ func (u *updater) Update(datastore database.Datastore) (resp vulnsrc.UpdateRespo
 		return resp, commonerr.ErrCouldNotDownload
 	}
 
+	if r.StatusCode != http.StatusOK {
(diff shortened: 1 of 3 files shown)


MackJM:
Although this would definitely help with what happened with RHEL. The same issue exists in debian and oracle. I suggest we move this to a package that we can share for all three. That way we can also set the User-Agent for our requests as well.

jzelinskie:
There doesn't seem to be much reusable code in this change to pull out into something shareable. However, I would like to eventually create an OVAL library that can be reused.

If you add this check for the few other vulnsrcs, I'll merge it.

Allda:
@jzelinskie I added HTTP status check to others vulnsrcs modules.

Allda:
Closing this PR since @jzelinskie added new http module which already does the status code check.

PR #601, opened 2018-08-30
 Add Status endpoint with Clair configuration

KeyboardNerd:
Implement a status endpoint providing the current lister/detector in
the current instance of Clair.


KeyboardNerd:
It also simplifies the run-in-docker file.

jzelinskie (approved):
LGTM. I also made the simplification of the run_in_docker script in my WIP PR that was playing with gogoprotobuf, so it makes a lot of sense to me.

PR #605, opened 2018-09-05
Implement Ancestry Layer-wise feature API

KeyboardNerd:
Ancestry now has the per-layer feature information instead of per-ancestry feature information.


api/v3/clairpb/convert.go
```diff
@@ -124,12 +124,13 @@ func VulnerabilityWithFixedInFromDatabaseModel(dbVuln database.VulnerabilityWith
...
 func AncestryFromDatabaseModel(dbAncestry database.Ancestry) *GetAncestryResponse_Ancestry {
-	ancestry := &GetAncestryResponse_Ancestry{
-		Name: dbAncestry.Name,
-	}
+	ancestry := &GetAncestryResponse_Ancestry{Name: dbAncestry.Name}
 	for _, layer := range dbAncestry.Layers {
-		ancestry.Layers = append(ancestry.Layers, LayerFromDatabaseModel(layer))
+		ancestryLayer := &GetAncestryResponse_AncestryLayer{}
+		ancestryLayer.Layer = LayerFromDatabaseModel(layer)
+		ancestry.Layers = append(ancestry.Layers, ancestryLayer)
```
jzelinskie: I think this can be reduced to

```go
ancestry.Layers = append(ancestry.Layers, &GetAncestryResponse_AncestryLayer{
  Layer:  LayerFromDatabaseModel(layer),
})
```

api/v3/rpc.go
```diff
@@ -115,54 +120,62 @@ func (s *AncestryServer) GetAncestry(ctx context.Context, req *pb.GetAncestryReq
...
 
-		if req.GetWithVulnerabilities() {
-			featureVulnerabilities, err := tx.FindAffectedNamespacedFeatures(ancestryWFeature.Features)
-			if err != nil {
-				return nil, status.Error(codes.Internal, err.Error())
-			}
+		respAncestry = &pb.GetAncestryResponse_Ancestry{Name: name}
+		respAncestry.ScannedDetectors = ancestry.ProcessedBy.Detectors
+		respAncestry.ScannedListers = ancestry.ProcessedBy.Listers
+		respAncestry.Layers = []*pb.GetAncestryResponse_AncestryLayer{}
```
jzelinskie: slightly more readable:

```go
respAncestry = &pb.GetAncestryResponse_Ancestry{
  Name: name,
  ScannedDectectors: ancestry.ProcessedBy.Detectors,
  ScannedListers: ancestry.ProcessedBy.Listers,
  Layers: []*pb.GetAncestryResponse_AncestryLayer{},
}
```

When you write struct literals like this, if you include names to the fields, you don't need to include all of the fields.

```go
type Derp struct { A, B int }

var example1 = Derp{A: 1} // Ok
var example2 = Derp{B: 1} // Ok
var example3 = Derp{1} // Explode
var example4 = Derp{1,2} // Ok, but order matters
```

database/models.go
```diff
@@ -29,34 +29,47 @@ type Processors struct {
...
-// ancestry, which is processed by `ProcessedBy`.
-type AncestryWithFeatures struct {
+// AncestryWithContent has the ancestry's name and the Ancestry Layers
+// associated with it.
+type AncestryWithContent struct {
 	Ancestry
 
-	ProcessedBy Processors
-	Features    []NamespacedFeature
+	// TODO(sidchen) deduplicate the Layers.
```
jzelinskie: is this something we want to do? won't the different layers have different context?

database/models.go
```diff
@@ -29,34 +29,47 @@ type Processors struct {
...
+	Layers []AncestryLayer
 }
 
-// Layer corresponds to a layer in an image processed by `ProcessedBy`.
+// AncestryLayer is a layer with all detected namespaced features.
+type AncestryLayer struct {
+	Layer
+
+	// DetectedFeatures are the features introduced by this layer.
+
```
jzelinskie: rm this newline

database/pgsql/ancestry.go
```diff
@@ -43,80 +43,89 @@ func (tx *pgSession) UpsertAncestry(ancestry database.Ancestry, features []datab
...
-		nf.Feature.VersionFormat = nf.Namespace.VersionFormat
-		awf.Features = append(awf.Features, nf)
+
+		feature.Feature.VersionFormat = feature.Namespace.VersionFormat // This looks strange.
+		features[int(layerIndex.Int64)] = append(features[int(layerIndex.Int64)], feature)
 	}
 
-	return awf, true, nil
+	// By the assumption of Ancestry Layer Index, we have the ancestry's layer
+	// index corresponding to the index in the array.
```
jzelinskie: This should really be in the Session interface doc comment for FindAncestryWithContent.

KeyboardNerd:
We need to have proper test infrastructure. The previous code doesn't work at all but passes all test checks. Now it's fixed though.

jzelinskie (changes requested):
some nitpicks in general, but this is overall looking okay. once we get everything in, it might be worth doing some cleanup

database/models.go
KeyboardNerd: Well, the problem here is that Ancestry has `Layers` field, which contains a list of `Layer`, and AncestryLayer also has `Layer`. The information is duplicated. I just don't really know a good way to deduplicate these.

database/pgsql/ancestry.go
KeyboardNerd: Well, this is an implementation detail for the ancestry_layer table. Here, I'm using the `ancestry_index` as the key to associate features to ancestry layers.

For people who implements this interface, they don't need to know what `ancestry_index` is. All they should do is get corresponding features for each layer in the Ancestry. They can do it in different ways, for example, query layer-wise feature directly using a single query.

jzelinskie: You're totally right.

database/models.go
```diff
@@ -29,34 +29,46 @@ type Processors struct {
...
-type AncestryWithFeatures struct {
+// AncestryWithContent has the ancestry's name and the Ancestry Layers
+// associated with it.
+type AncestryWithContent struct {
 	Ancestry
 
-	ProcessedBy Processors
-	Features    []NamespacedFeature
+	// TODO(sidchen) deduplicate the Layers.
+	Layers []AncestryLayer
```
jzelinskie: We should comment here about the order of the layers.

database/models.go
jzelinskie: Why not have both types be the same but leave the DetectedFeatures for each Layer empty when it isn't AncestryWithContent?

KeyboardNerd: I remember the design is that: Every struct, when returned, MUST have all fields filled to represent the whole information. Therefore, leaving DetectedFeatures empty will indicate that the Layer has NO detected features.

jzelinskie (approved):

PR #606, opened 2018-09-05
Adding httputil and version packages to master

MackJM:
- Debian/RHEL/Oracle vulnsrc now use httputil to download files
- httputil sets the User-Agent to the requests as Clair/<version> (https://github.com/coreos/clair)
- httputil holds Status2xx() which returns if the response is a http success (2xx)
- GetClientAddr moved from api/httputil to pkg/httputil
- the version packge holds a Version string which is set at build time from the git tag and sha
- the .git directory was removed from .dockerignore so that we can use the git tag to set the version


jzelinskie (approved):
LGTM

PR #607, opened 2018-09-05
enforce gofmt in travis

jzelinskie:


KeyboardNerd (approved):
LGTM

Yeah, by default gofmt won't simplify but since it seems the mechanism is very simple and trivial, I'll think it's reasonable to have it.
https://github.com/golang/go/blob/master/src/cmd/gofmt/simplify.go

PR #609, opened 2018-09-06
Using httputil for NVD

MackJM:
nvd was missed when moving to httputil, this fixes it


jzelinskie:
@MackJM can you please update this to address all the changes in #610? It's blocking a release for fixing NVD in stable.

jzelinskie:
We decided it's too complicated to backport this.

PR #610, opened 2018-09-06
Using httputil for NVD

MackJM:
nvd was missed when moving to httputil, this fixes it

### DIFF
--- a/ext/vulnmdsrc/nvd/nvd.go
+++ b/ext/vulnmdsrc/nvd/nvd.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -36,6 +35,7 @@ import (
 	"github.com/coreos/clair/database"
 	"github.com/coreos/clair/ext/vulnmdsrc"
 	"github.com/coreos/clair/pkg/commonerr"
+	"github.com/coreos/clair/pkg/httputil"
 )
 
 const (
@@ -166,9 +166,15 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
 			}
 
 			// Download data feed.
-			r, err := http.Get(fmt.Sprintf(dataFeedURL, dataFeedName))
+			r, err := httputil.GetWithUserAgent(fmt.Sprintf(dataFeedURL, dataFeedName))
 			if err != nil {
 				log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not download NVD data feed")
+				return dataFeedReaders, dataFeedHashes, commonerr.ErrCouldNotDownload
+			}
+			// r is closed in BuildCache()
+
+			if !httputil.Status2xx(r) {
+				log.WithField("StatusCode", r.StatusCode).Error("Failed to download NVD data feed")
 				return dataFeedReaders, dataFeedHashes, commonerr.ErrCouldNotDownload
 			}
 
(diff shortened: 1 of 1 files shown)


ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -166,11 +166,17 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
 			}
 
 			// Download data feed.
-			r, err := http.Get(fmt.Sprintf(dataFeedURL, dataFeedName))
+			r, err := httputil.GetWithUserAgent(fmt.Sprintf(dataFeedURL, dataFeedName))
 			if err != nil {
 				log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not download NVD data feed")
 				return dataFeedReaders, dataFeedHashes, commonerr.ErrCouldNotDownload
 			}
+			// r is closed in BuildCache()
```
jzelinskie: not if an error occurs

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -199,12 +205,16 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
...
 func getHashFromMetaURL(metaURL string) (string, error) {
-	r, err := http.Get(metaURL)
+	r, err := httputil.GetWithUserAgent(metaURL)
 	if err != nil {
 		return "", err
 	}
 	defer r.Body.Close()
 
+	if !httputil.Status2xx(r) {
+		return "", errors.New("Unsuccesuful status code: " + string(r.StatusCode))
```
jzelinskie: `Unsuccessful`

Also having some more context in this message would be nice. I have no idea what request was being tried that returned a non-200 status code.

jzelinskie (changes requested):

MackJM:
I was wrong at first with that comment, because that's where it crashed when I had defer there.

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -171,7 +171,7 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
 				log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not download NVD data feed")
 				return dataFeedReaders, dataFeedHashes, commonerr.ErrCouldNotDownload
 			}
-			// r is closed in BuildCache()
+			defer r.Body.Close()
```
jzelinskie: This won't close anything until the whole function getDataFeeds() exits. This is not what we want to do.

It could end up being more trouble than it's worth, but wrapping everything inside the loop iteration into a it's own function might be the best way to get defer to execute at the right time.

MackJM: Yeah, it was a little outside of what I set out with my initial PR, but I agree with you. I think I wrapped it up nicely now. Without spending too much effort.

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -165,46 +165,64 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
 				}
 			}
 
-			// Download data feed.
-			r, err := http.Get(fmt.Sprintf(dataFeedURL, dataFeedName))
-			if err != nil {
-				log.WithError(err).WithField(logDataFeedName, dataFeedName).Error("could not download NVD data feed")
-				return dataFeedReaders, dataFeedHashes, commonerr.ErrCouldNotDownload
-			}
+			downloadAndSave := func() error {
```
jzelinskie: I think you should pull this out into a top-level function called downloadFeed, rather than an inline function. After that LGTM.

jzelinskie (changes requested):

ext/vulnmdsrc/nvd/nvd.go
```diff
@@ -165,46 +165,68 @@ func getDataFeeds(dataFeedHashes map[string]string, localPath string) (map[strin
...
+	gr, err := gzip.NewReader(r.Body)
+	if err != nil {
+		log.WithError(err).WithFields(log.Fields{"StatusCode": r.StatusCode, "DataFeedName": dataFeedName}).Error("could not read NVD data feed")
+		return commonerr.ErrCouldNotDownload
+	}
+
+	// Store it to a file at the same time if possible.
+	f, err := os.Create(fileName)
+	if err != nil {
+
```
jzelinskie: Sorry for the OCD, but drop this newline and we're good to go.

jzelinskie (changes requested):

jzelinskie (approved):

PR #611, opened 2018-09-06
Remove dependency on graceful

jzelinskie:
This library has no longer been needed since Go 1.8 added `http.Server.Shutdown()`.


api/api.go
```diff
@@ -93,9 +85,7 @@ func listenAndServeWithStopper(srv *graceful.Server, st *stopper.Stopper, certFi
 	}
 
 	if err != nil {
-		if opErr, ok := err.(*net.OpError); !ok || (ok && opErr.Op != "accept") {
-			log.Fatal(err)
-		}
+		log.Fatal(err)
```
KeyboardNerd: https://golang.org/pkg/net/http/#Server.Shutdown 
By the example, ListenAndServe will return ErrServerClosed when Shutdown is called. Should we change it to be that only when err is not ErrServerClosed, we log Fatal.

jzelinskie: Nice save!

KeyboardNerd (approved):
LGTM

PR #612, opened 2018-09-06
Configurable package endpoint URLs

MackJM:
- Added 'endpoints' under updater in the config
- Endpoints are set at init, or the default is used if blank or omited
- Added SetEndpointURL to vulnsrc and vulnmdsrc
- Added example endpoint config to config.example.yaml

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -55,6 +55,15 @@ clair:
     # Frequency the database will be updated with vulnerabilities from the default data sources
     # The value 0 disables the updater entirely.
     interval: 2h
+    # (Optional) Endpoints can be used to change the endpoint that a package uses
+    # If omited the default is used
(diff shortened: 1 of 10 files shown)


MackJM:
@jzelinskie any thoughts on this?

reasonerjt:
This change is valuable for 2.0 to mitigate the impact when the endpoint of vuln data updates.

jzelinskie:
So, I feel like this is reasonable for v2.x, but not for v3.x because there are no guarantees in the future that every vulnsrc can be represented by a URL prefix.

MackJM:
@jzelinskie I believe it should be good to go.

MackJM:
@jzelinskie I dunno if you have time but I'd love to close this.

PR #613, opened 2018-09-06
Introduce pkg/pagination

jzelinskie:
This change refactors a lot of the code dealing with pagination so that
fernet implementation details do not leak.

- Deletes database/pgsql/token
- Introduces a pagination package
- Renames idPageNumber to Page and add a constructor and method.


pkg/pagination/pagination.go
```diff
@@ -0,0 +1,103 @@
...
+type Key string
+
+// MustGenerateNewKey generates a new pagination key.
+//
+// This function will panic if your system does not have enough entropy to
+// produce the required randomness.
+func MustGenerateNewKey() Key {
+	var key fernet.Key
+	if err := key.Generate(); err != nil {
+		panic("pagination: failed to read enough randomness to generate a new key")
```
KeyboardNerd: I understand that by current implementation, it'll only return error because of insufficient randomness. But will this error be the only error that key.Generate function will generate in the future?

pkg/pagination/pagination.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// KeyFromString creates the key for a given string.
+//
+// Strings must be 32-bit URL-safe base64 representations of the key bytes.
+func KeyFromString(keyString string) (k Key, err error) {
+	var key *fernet.Key
+	key, err = fernet.DecodeKey(keyString)
+	if err != nil {
+		err = errors.New("invalid pagination key string: must be 32-bit URL-safe base64")
```
KeyboardNerd: Is this intended to be un-exported error?

pkg/pagination/pagination.go
KeyboardNerd: I understand that by current implementation, it'll only return error because of insufficient randomness. But will this error be the only error that key.Generate function will generate in the future?

pkg/pagination/pagination.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	decoded, _ := fernet.DecodeKey(string(k))
+	return fernet.EncryptAndSign(buf.Bytes(), decoded)
+}
+
+// UnmarshalToken decrypts a token using provided key and decodes the result
+// into the provided interface.
+func (k Key) UnmarshalToken(token string, v interface{}) error {
+	decoded, _ := fernet.DecodeKey(string(k))
```
KeyboardNerd: Do we really want to suppress the error here? Yes, it was not handled before. If there's a valid reason, I think it makes sense to comment it.

KeyboardNerd (commented):
A few nitpicks about the error handling. Otherwise, it looks good.

pkg/pagination/pagination.go
jzelinskie: If you only ever make Keys from the constructors, they'd always be valid.

However, I'll change this so that the Key type actually keeps the decoded bytes in memory rather than the string representation.

pkg/pagination/pagination.go
jzelinskie: I'll export it. I also just noticed that it's 32 **bytes** not bits. Woops

pkg/pagination/pagination.go
jzelinskie: I went down the rabbit hole and talked to a bunch of the people working on Go crypto about this. I think it's safe to assume that fernet probably isn't going to change anything in this function and that any errors from crypto/rand.Read() should panic.

KeyboardNerd (approved):
LGTM

PR #614, opened 2018-09-07
Replace Ancestry with AncestryWithContent struct in database models

KeyboardNerd:
As one of the steps to simplifies the codebase, the AncestryWithContent
struct is renamed to Ancestry, and Ancestry is removed. It will cause
the PostAncestry request to be slower.


api/v3/rpc.go
```diff
@@ -118,79 +114,41 @@ func (s *AncestryServer) GetAncestry(ctx context.Context, req *pb.GetAncestryReq
...
-					ancestryLayer.DetectedFeatures = append(ancestryLayer.DetectedFeatures, pb.NamespacedFeatureFromDatabaseModel(dbFeature))
-				}
-			}
-
-			respAncestry.Layers = append(respAncestry.Layers, ancestryLayer)
-		}
-	} else {
-		dbAncestry, ok, err := tx.FindAncestry(name)
+	for _, layer := range ancestry.Layers {
+		pbLayer, err := GetPbAncestryLayer(tx, layer, req.GetWithFeatures(), req.GetWithVulnerabilities())
```
jzelinskie: I was actually thinking about totally removing these from the proto, too. Let's just always return everything.

jzelinskie (changes requested):
Beyond this design decision, everything else looks pretty good.

api/v3/rpc.go
KeyboardNerd: hmm idk if we want to always return the vulnerabilities though

jzelinskie: Quay almost exclusively requests with vulnerabilities. IIRC the only scenario it doesn't is in tests.  Clients that don't care about the vulnerabilities can just ignore them and this avoids us maintaining this branch needlessly.

jzelinskie (approved):

PR #615, opened 2018-09-10
Extend the sleep duration if update fails

reasonerjt:
If one of the updaters failed in the first update.  It will sleep for 1
minute and rerun all the updaters again.  This will generate redundant
requestss to the remote sites.
This commit mitigate such issue by extending the sleep duration after
each failure in update.

Signed-off-by: Daniel Jiang <jiangd@vmware.com>

### DIFF
--- a/updater.go
+++ b/updater.go
@@ -83,7 +83,7 @@ func RunUpdater(config *UpdaterConfig, datastore database.Datastore, st *stopper
 	log.WithField("lock identifier", whoAmI).Info("updater service started")
 
 	for {
-		var stop, status bool
+		var stop, updateSuccessful bool
 
 		// Determine if this is the first update and define the next update time.
 		// The next update time is (last update time + interval) or now if this is the first update.
@@ -105,7 +105,7 @@ func RunUpdater(config *UpdaterConfig, datastore database.Datastore, st *stopper
 				// Launch update in a new go routine.
 				doneC := make(chan bool, 1)
 				go func() {
-					status = update(datastore, firstUpdate)
+					updateSuccessful = update(datastore, firstUpdate)
 					doneC <- true
 				}()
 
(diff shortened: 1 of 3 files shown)


jzelinskie:
We should update this to use exponential backoff.
The proper way to do so would to be make a pkg/timeutil containing:

```go
func ExpBackoff(prev, max time.Duration) time.Duration {
	if prev == 0 {
		return time.Second
	}
	if prev > max/2 {
		return max
	}
	return 2 * prev
}
```

reasonerjt:
@jzelinskie Thanks for the suggestion, I'll update this PR in the next couple of days.

reasonerjt:
@jzelinskie 
Sorry for the late follow up.
I have made update according to your suggestion, please have a look when you have time.

Thanks.

updater.go
```diff
@@ -176,7 +186,7 @@ func sleepUpdater(approxWakeup time.Time, st *stopper.Stopper) (stopped bool) {
 
 // update fetches all the vulnerabilities from the registered fetchers, upserts
 // them into the database and then sends notifications.
-func update(datastore database.Datastore, firstUpdate bool) {
+func update(datastore database.Datastore, firstUpdate bool) (status bool) {
```
jzelinskie: I'm not sure you've done anything with status? Is the last line of this function supposed to be `return true`?

jzelinskie (changes requested):

updater.go
reasonerjt: Thanks for the review.

It's `named return value`, for details see:
https://tour.golang.org/basics/7

In this case it returns the value of `status` which was assigned in this line:
https://github.com/coreos/clair/pull/615/files#diff-de69c370937723718c3c4e015f8e8ba0L185

If you think this is confusing, another option is to update the declaration to:
```
func update(datastore database.Datastore, firstUpdate bool)  bool { ...
```
and update all the `return` statements to:
```
return status
```

reasonerjt: @jzelinskie please kindly let me know your thought.

jzelinskie: Oh sorry. I literally just missed that line because it was under the GitHub fold.

If we rename status to something more descriptive like "updateSuccessful", then this LGTM

reasonerjt:
@jzelinskie 
Thanks for the reply.  I've made update to the variable name.

jzelinskie (approved):

jzelinskie:
Thank you for your patience!

PR #617, opened 2018-09-11
Introduce pkg/grpcutil

jzelinskie:
This change refactors all of the gRPC code used in the v3 package to be generic and reusable.

### DIFF
--- a/pkg/grpcutil/gateway.go
+++ b/pkg/grpcutil/gateway.go
@@ -1,0 +1,81 @@
+// Copyright 2018 clair authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package grpcutil
+
+import (
+	"crypto/tls"
+	"net/http"
+	"strings"
+
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+// RegisterServiceHandlerFunc is a function that registers ServiceHandlers with
+// a ServeMux.
+type RegisterServiceHandlerFunc func(context.Context, *runtime.ServeMux, *grpc.ClientConn) error
+
+// NewGateway creates a new http.Handler and grpc.ClientConn with the provided
+// gRPC Services registered.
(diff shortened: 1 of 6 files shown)


pkg/grpcutil/gateway.go
```diff
@@ -0,0 +1,85 @@
...
+	} else {
+		dialOpts = append(dialOpts, grpc.WithInsecure())
+	}
+
+	srvmux := runtime.NewServeMux(
+		// Include empty fields with default values.
+		runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{EmitDefaults: true}),
+	)
+
+	conn, err := grpc.DialContext(context.TODO(), addr, dialOpts...)
```
KeyboardNerd: It was using context.Background() in the old code. Should we still use that?

pkg/grpcutil/gateway.go
```diff
@@ -0,0 +1,85 @@
...
+
+// NewGatewayHandler creates a new http.Handler with the provided gRPC Services
+// registered.
+func NewGatewayHandler(addr string, tlsConfig *tls.Config, funcs []RegisterServiceHandlerFunc) (http.Handler, error) {
+	// Configure the right DialOptions the for TLS configuration.
+	var dialOpts []grpc.DialOption
+	if tlsConfig != nil {
+		var gwTLSConfig *tls.Config
+		gwTLSConfig = tlsConfig.Clone()
+		gwTLSConfig.InsecureSkipVerify = true
```
KeyboardNerd: Can we add a comment on why we use InsecureSkipVerify here?

pkg/grpcutil/gateway.go
```diff
@@ -0,0 +1,85 @@
...
+		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(gwTLSConfig)))
+	} else {
+		dialOpts = append(dialOpts, grpc.WithInsecure())
+	}
+
+	srvmux := runtime.NewServeMux(
+		// Include empty fields with default values.
+		runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{EmitDefaults: true}),
```
KeyboardNerd: Do you remember why do we want to include the empty fields?

KeyboardNerd (commented):
Some comments on gateway.go
We should also add tests on server connectivity in the future.

pkg/grpcutil/gateway.go
jzelinskie: We should use `TODO()` until the code has been entirely refactored to support context from the start of the application. `Background()` is how you'd initialize a context if you were doing so from the proper location, but in this case, we'll have to handle that another time.

KeyboardNerd: Sure

pkg/grpcutil/gateway.go
jzelinskie: Nope

KeyboardNerd: Yeah, just get rid of it, I don't think there will be any issues.

KeyboardNerd (approved):
LGTM

PR #619, opened 2018-09-11
database: remove LayerWithContent from interface

KeyboardNerd:
- Layer is renamed to LayerMetadata
- LayerWithContent is renamed to Layer
- Remove redundant interfaces to pertain layer's metadata

### DIFF
--- a/database/pgsql/vulnerability.go
+++ b/database/pgsql/vulnerability.go
@@ -16,7 +16,6 @@ package pgsql
 
 import (
 	"database/sql"
-	"encoding/json"
 	"errors"
 	"time"
 
@@ -218,17 +217,6 @@ func (tx *pgSession) insertVulnerabilities(vulnerabilities []database.Vulnerabil
 	}
 
 	return vulnIDs, nil
-}
-
-// castMetadata marshals the given database.MetadataMap and unmarshals it again to make sure that
-// everything has the interface{} type.
-// It is required when comparing crafted MetadataMap against MetadataMap that we get from the
-// database.
-func castMetadata(m database.MetadataMap) database.MetadataMap {
-	c := make(database.MetadataMap)
-	j, _ := json.Marshal(m)
-	json.Unmarshal(j, &c)
-	return c
 }
 
 func (tx *pgSession) lockFeatureVulnerabilityCache() error {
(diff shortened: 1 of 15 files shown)


KeyboardNerd:
After this, there will be a PR for relating features with Lister/Detector to enable the API wise filtering.

worker.go
```diff
@@ -222,20 +220,29 @@ func processLayers(datastore database.Datastore, imageFormat string, requests []
...
 
-	for _, res := range partialRes {
-		if err := persistPartialLayer(datastore, res); err != nil {
+	for _, layer := range partialLayers {
+		if err := persistPartialLayer(datastore, layer); err != nil {
 			return nil, err
 		}
+
+		log.WithFields(
+			log.Fields{
```
jzelinskie: put this on the line above

database/pgsql/vulnerability.go
```diff
@@ -220,17 +219,6 @@ func (tx *pgSession) insertVulnerabilities(vulnerabilities []database.Vulnerabil
...
-// castMetadata marshals the given database.MetadataMap and unmarshals it again to make sure that
-// everything has the interface{} type.
-// It is required when comparing crafted MetadataMap against MetadataMap that we get from the
-// database.
-func castMetadata(m database.MetadataMap) database.MetadataMap {
-	c := make(database.MetadataMap)
-	j, _ := json.Marshal(m)
-	json.Unmarshal(j, &c)
-	return c
-}
```
jzelinskie, line 232: Why don't we need this anymore?

jzelinskie (changes requested):
only a nitpick and a question

database/pgsql/vulnerability.go
KeyboardNerd: No, we don't need it. We will not compare crafted MetadataMap against MetadataMap from database query result any more because Updater will only compare MetadataMaps from database.

KeyboardNerd: https://github.com/coreos/clair/blob/master/updater.go#L515 It may be by design that we will only create notification when there's big change in the vulnerability. Maybe, in the future we can make it configurable.

jzelinskie (approved):

PR #621, opened 2018-09-14
pkg/gitutil: init

jzelinskie:
This refactors the code we're using to manage temporary git repositories
into a utility package.

### DIFF
--- a/ext/vulnsrc/alpine/alpine.go
+++ b/ext/vulnsrc/alpine/alpine.go
@@ -1,4 +1,4 @@
-// Copyright 2017 clair authors
+// Copyright 2018 clair authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,7 +20,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -31,7 +30,7 @@ import (
 	"github.com/coreos/clair/ext/versionfmt"
 	"github.com/coreos/clair/ext/versionfmt/dpkg"
 	"github.com/coreos/clair/ext/vulnsrc"
-	"github.com/coreos/clair/pkg/commonerr"
+	"github.com/coreos/clair/pkg/gitutil"
 )
 
 const (
@@ -53,7 +52,7 @@ func (u *updater) Update(db database.Datastore) (resp vulnsrc.UpdateResponse, er
 
 	// Pull the master branch.
 	var commit string
-	commit, err = u.pullRepository()
+	u.repositoryLocalPath, commit, err = gitutil.CloneOrPull(secdbGitURL, u.repositoryLocalPath, updaterFlag)
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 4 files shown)


ext/vulnsrc/alpine/alpine.go
```diff
@@ -183,40 +182,6 @@ func parseVulnsFromNamespace(repositoryPath, namespace string) (vulns []database
 	return
 }
```
KeyboardNerd, line 184: Wow.
Very duplication.
So Git.

pkg/gitutil/gitutil.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Prepare a command to clone the repository.
+	cmd := exec.Command("git", "clone", remote, ".")
+	cmd.Dir = path
+
+	// Execute the command.
+	var commandOutput []byte
+	commandOutput, err = cmd.CombinedOutput()
+	if err != nil {
+		os.RemoveAll(path)
```
KeyboardNerd: log returned error

yeah, previously it wasn't handled at all.

pkg/gitutil/gitutil.go
```diff
@@ -0,0 +1,140 @@
...
+	return revParseHead(path)
+}
+
+// CloneOrPull performs a git pull if the desiredPath exists on the filesystem.
+// Otherwise, it generates a new temporary directory and performs a git clone
+// to that path.
+func CloneOrPull(remote, desiredPath string) (path, head string, err error) {
+	// Create a temporary directory if the path is unspecified.
+	if desiredPath == "" {
+		path, err = ioutil.TempDir(os.TempDir(), remote)
```
KeyboardNerd: I don't think we should use `remote` for the prefix 
https://github.com/coreos/clair/pull/621/files#diff-154f97505971bb2f44a0f30045d133d2R38
We can use the updater flag I think.

pkg/gitutil/gitutil.go
```diff
@@ -0,0 +1,140 @@
...
+		err = ErrFailedPull
+		return
+	}
+
+	return revParseHead(path)
+}
+
+// CloneOrPull performs a git pull if the desiredPath exists on the filesystem.
+// Otherwise, it generates a new temporary directory and performs a git clone
+// to that path.
```
KeyboardNerd: Different from the documentation here, the code only generate a new temporary directory when the desired path is not given. Is that the expected behavior?

KeyboardNerd (changes requested):

ext/vulnsrc/alpine/alpine.go
jzelinskie: [](https://github.com/thiderman/doge)

pkg/gitutil/gitutil.go
```diff
@@ -0,0 +1,146 @@
...
+
+// CloneOrPull performs a git pull if there is a git repository located at
+// desiredPath. Otherwise, it performs a git clone to that path.
+//
+// If repoPath is left empty, a temporary directory is generated with the
+// provided prefix and returned.
+func CloneOrPull(remote, repoPath, tempDirPrefix string) (path, head string, err error) {
+	// Create a temporary directory if the path is unspecified.
+	if repoPath == "" {
+		path, err = ioutil.TempDir(os.TempDir(), remote)
```
KeyboardNerd: remote -> tempDirPrefix

pkg/gitutil/gitutil.go
```diff
@@ -0,0 +1,146 @@
...
+		}).Error("failed to git rev-parse repository")
+		err = ErrFailedPull
+		return
+	}
+
+	return revParseHead(path)
+}
+
+// CloneOrPull performs a git pull if there is a git repository located at
+// desiredPath. Otherwise, it performs a git clone to that path.
```
KeyboardNerd: nit: desired path?

KeyboardNerd (changes requested):

KeyboardNerd (approved):

PR #639, opened 2018-10-12
Use SHA256 instead of SHA1 for fingerprinting

Katee:
The current use of SHA1 for fingerprinting is safe. However, there is very little downside to switching to SHA256. Some static analysis tools warn about the use of SHA1.


jzelinskie (approved):
LGTM!

PR #645, opened 2018-10-17
Switch to NVD JSON feed and include CVSSv3

Katee:
The NVD JSON feed is still "beta". According to the [NVD changlog](https://nvd.nist.gov/vuln/Data-Feeds/JSON-feed-changelog) the last breaking change (a field rename which would not have affected this code) happened 2017-12-18. **CVSSv3 is only in the JSON feed.**

Some things in the feed relevant to Clair have changed:
* The format for `PublishedDateTime` has changed from `2018-01-10T17:29:00.930-05:00` to `2018-01-10T22:29Z`.
* The CVSSv2 and CVSSv3 vector strings are provided directly. If desired, I'm happy to remove the parsing code here or in another PR.

The changes in this PR are meant to be backwards compatible:
* `NVDmetadataCVSSv2` remains unchanged. `PublishedDateTime` is still in `NVDmetadataCVSSv2` even though it isn't specific to CVSSv2.
* As before, `Metadata` returns nil if CVSSv2 isn't available (and this now has a test).

### DIFF
--- a/ext/vulnmdsrc/nvd/json.go
+++ b/ext/vulnmdsrc/nvd/json.go
@@ -22,27 +22,39 @@ import (
 )
 
 type nvd struct {
-	Entries []nvdEntry `xml:"entry"`
+	Entries []nvdEntry `json:"CVE_Items"`
 }
 
 type nvdEntry struct {
-	Name              string  `xml:"http://scap.nist.gov/schema/vulnerability/0.4 cve-id"`
-	CVSS              nvdCVSS `xml:"http://scap.nist.gov/schema/vulnerability/0.4 cvss"`
-	PublishedDateTime string  `xml:"http://scap.nist.gov/schema/vulnerability/0.4 published-datetime"`
+	CVE               nvdCVE    `json:"cve"`
+	Impact            nvdImpact `json:"impact"`
+	PublishedDateTime string    `json:"publishedDate"`
 }
 
-type nvdCVSS struct {
-	BaseMetrics nvdCVSSBaseMetrics `xml:"http://scap.nist.gov/schema/cvss-v2/0.2 base_metrics"`
+type nvdCVE struct {
+	Metadata nvdCVEMetadata `json:"CVE_data_meta"`
 }
 
-type nvdCVSSBaseMetrics struct {
-	Score            float64 `xml:"score"`
-	AccessVector     string  `xml:"access-vector"`
-	AccessComplexity string  `xml:"access-complexity"`
-	Authentication   string  `xml:"authentication"`
-	ConfImpact       string  `xml:"confidentiality-impact"`
-	IntegImpact      string  `xml:"integrity-impact"`
-	AvailImpact      string  `xml:"availability-impact"`
+type nvdCVEMetadata struct {
+	CVEID string `json:"ID"`
+}
+
+type nvdImpact struct {
+	BaseMetricV2 nvdBaseMetricV2 `json:"baseMetricV2"`
+}
+
+type nvdBaseMetricV2 struct {
+	CVSSv2 nvdCVSSv2 `json:"cvssV2"`
+}
+
+type nvdCVSSv2 struct {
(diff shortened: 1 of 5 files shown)


ext/vulnmdsrc/nvd/json.go
```diff
@@ -0,0 +1,162 @@
...
+package nvd
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type nvd struct {
```
skelterjohn, line 24: Is there a reason you chose to not export any of these type names? Specifically, it means that code in another package cannot construct one of these unless it takes a json blob and runs it through the parser. Constructing them directly can be useful for testing.

Katee: This matches the previous `xml.go` file. The purpose of having these structs unexported is so changes to the underlying feed don't require a change to Clair's data model. Other packages should only be depending on Clair's data models, not this intermediate representation.

In `nvd.go` the `NVDmetadataCVSSv3` struct could be changed to include all the CVSSv3 fields, depends on what the Clair maintainers want.

Tests in the same package have access even if things are not exported.

ext/vulnmdsrc/nvd/json.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2017 clair authors
```
jzelinskie:
```suggestion
// Copyright 2018 clair authors
```

ext/vulnmdsrc/nvd/json.go
```diff
@@ -0,0 +1,162 @@
...
+	AttackComplexity   string  `json:"attackComplexity"`
+	PrivilegesRequired string  `json:"privilegesRequired"`
+	UserInteraction    string  `json:"userInteraction"`
+	Scope              string  `json:"scope"`
+	ConfImpact         string  `json:"confidentialityImpact"`
+	IntegImpact        string  `json:"integrityImpact"`
+	AvailImpact        string  `json:"availabilityImpact"`
+}
+
+var vectorValuesToLetters map[string]string
```
jzelinskie: No need to use `init()`:

```suggestion
var vectorValuesToLetters = map[string]string{
	"NETWORK":          "N",
	"ADJACENT_NETWORK": "A",
	"LOCAL":            "L",
	"HIGH":             "H",
	"MEDIUM":           "M",
	"LOW":              "L",
	"NONE":             "N",
	"SINGLE":           "S",
	"MULTIPLE":         "M",
	"PARTIAL":          "P",
	"COMPLETE":         "C",

	// CVSSv3 only
	"PHYSICAL":  "P",
	"REQUIRED":  "R",
	"CHANGED":   "C",
	"UNCHANGED": "U",
}
```

jzelinskie (changes requested):
just two minor nitpicks

Katee:
I'd also like to pass through the CVSSv3 expoloitability score and impact score, I'm adding a commit for that now.

ext/vulnmdsrc/nvd/json.go
```diff
@@ -0,0 +1,184 @@
...
+	vectorValuesToLetters["MULTIPLE"] = "M"
+	vectorValuesToLetters["PARTIAL"] = "P"
+	vectorValuesToLetters["COMPLETE"] = "C"
+
+	// CVSSv3 only
+	vectorValuesToLetters["PHYSICAL"] = "P"
+	vectorValuesToLetters["REQUIRED"] = "R"
+	vectorValuesToLetters["CHANGED"] = "C"
+	vectorValuesToLetters["UNCHANGED"] = "U"
+}
```
jzelinskie: I couldn't figure out how to make the GitHub suggestion delete this, too. ☹️

Katee: Whoops! I didn't notice that merging from the GitHub UI. Fixed. (Do you want any history re-written?)

jzelinskie (approved):

PR #647, opened 2018-10-19
vulnsrc: Refactor debian and alpine sources

KeyboardNerd:
* Add fsutil to handle listing directory function
* Add FindKeyValue wrapper in database module
* Refactor debian and alpine sources to use them


ext/vulnsrc/alpine/alpine_test.go
```diff
@@ -15,28 +15,24 @@
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestYAMLParsing(t *testing.T) {
 	_, filename, _, _ := runtime.Caller(0)
 	path := filepath.Join(filepath.Dir(filename))
+	secdb, err := parseSecDBFile(path + "/testdata/v34_main.yaml")
```
jzelinskie: We should use `filepath.Join()` again here instead of `+`.
 While it's obvious now, we you refactor and change code, it becomes less obvious over time.

ext/vulnsrc/alpine/alpine.go
```diff
@@ -195,42 +132,59 @@ type secDBFile struct {
...
-	for _, pack := range file.Packages {
-		pkg := pack.Pkg
-		for version, vulnStrs := range pkg.Fixes {
-			err := versionfmt.Valid(dpkg.ParserName, version)
-			if err != nil {
-				log.WithError(err).WithField("version", version).Warning("could not parse package version. skipping")
+	file = &secDBFile{}
+	err = yaml.Unmarshal(data, file)
+	return
+}
```
jzelinskie, line 145: Prefer decoders when available to Unmarshal.
Decoders stream data and can avoid ballooning memory when processing large files.
I know these aren't very large right now, but they could become large at some point in the future.

```go
func parseSecDBFile(filePath string) (file *secDBFile, err error) {
  var f io.ReadCloser
  f, err = os.Open(filePath)
  if err != nil {
    return
  }
  defer f.Close()

  file = &secDBFile{}
  err = yaml.NewDecoder(f).Decode(&file)
  return
}
```

pkg/fsutil/fsutil.go
```diff
@@ -0,0 +1,64 @@
...
+const (
+	// FilesOnly filters Dir function to return only files
+	FilesOnly dirFilter = iota
+	// DirectoriesOnly filters Dir function to return only directories
+	DirectoriesOnly
+)
+
+// Dir lists the files or folders under the given path and filter based on the
+// dirFilter.
+func Dir(path string, filter dirFilter) ([]string, error) {
```
jzelinskie: I'd name this `Readdir` for transparency.

pkg/fsutil/fsutil.go
```diff
@@ -0,0 +1,64 @@
...
+	"strings"
+)
+
+type dirFilter int
+
+const (
+	// FilesOnly filters Dir function to return only files
+	FilesOnly dirFilter = iota
+	// DirectoriesOnly filters Dir function to return only directories
+	DirectoriesOnly
```
jzelinskie, line 31: A filter that doesn't exclude anything would be a good value for `iota`

ext/vulnsrc/alpine/alpine.go
```diff
@@ -195,42 +132,59 @@ type secDBFile struct {
...
+
+func extractVulnerabilities(file *secDBFile) (vulns []database.VulnerabilityWithAffected) {
```
jzelinskie: This looks like a good place to make this a method and turn parseSecDBFile into a constructor:

secDBFile => secDB
parseSecDBFile => newSecDB
extractVulnerabilites => secDB.Vulnerabilities()

jzelinskie (changes requested):

KeyboardNerd:
Updated according to comments

jzelinskie (approved):

PR #648, opened 2018-10-28
Bump Go versions and use '.x' to always get latest patch versions

HaraldNordgren:


KeyboardNerd:
but why

jzelinskie (approved):
LGTM, thanks.

PR #653, opened 2018-10-31
Pinning helm postgres dep to the working 1.0.0

brosander:
It looks like the postgres chart just had a huge shakeup, image, maintainer change:
https://github.com/helm/charts/commit/7f1e47c5a8b10f24547aa385d782294622f8f439#diff-fb61f7f926e5bfb00f3b15a715d90302

The new chart isn't playing nice with clair (even after updating the postgres args to account for new format on user, password, db) so I pinned the old one for now.

### DIFF
--- a/contrib/helm/clair/requirements.yaml
+++ b/contrib/helm/clair/requirements.yaml
@@ -1,5 +1,5 @@
 dependencies:
   - name: postgresql
-    version: "*"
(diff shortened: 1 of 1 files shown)


jzelinskie (approved):
LGTM. Hopefully we can update it soon, if it makes sense to do so.

PR #672, opened 2018-12-12
Implement Feature types

KeyboardNerd:
The PR is to differentiate packages by their types, and vulnerabilities to affect different types of features.

Change set:
1. Feature Detectors: Extract both binary and source packages from image layers

2. Vulnerability matching algorithm change: Each vulnerability matches a specific type of package, either binary or source package.

| Vulnerability source | Matching Package Type |
|---------------------|---------------|
| Debian: "https://security-tracker.debian.org/tracker" | matching source package name |
| Ubuntu: "https://git.launchpad.net/ubuntu-cve-tracker" | matching source package name |
| Alpine: "https://github.com/alpinelinux/alpine-secdb" | matching binary package name |
| Oracle: "https://linux.oracle.com/oval/" | matching binary package name |
| RHEL: "https://www.redhat.com/security/data/oval/" | matching binary package name |

Please correct the above assumption for each vulnerability data source if any error.

3. API change: Clair returns all found binary and source packages to the client, and let the client to filter.
For each Feature, we add a field to indicate the type of the feature, currently it's Binary or Source package.


KeyboardNerd:
We can choose which type of the package to be returned on API based on the package manager.
e.g.
For any feature detected in sRPM, we show the source package
For any feature detected in RPM, we show the binary package

KeyboardNerd:
The size of `models.go` grows exponentially, we should split that into smaller files.
Also models.go describes the objects used by Clair core logic to compute the vulnerabilities. We should have a separate thing in the actual pgsql implementation to represent the pgsql rows.

database/affected_feature_type.go
```diff
@@ -1,26 +0,0 @@
-// Copyright 2018 clair authors
```
KeyboardNerd, line 1: This is renamed to `database/feature_type.go`

database/database.go
```diff
@@ -27,20 +26,20 @@ import (
...
 
 	// ErrInvalidParameters is an error that occurs when the parameters are not valid.
-	ErrInvalidParameters = errors.New("database: parameters are not valid")
+	ErrInvalidParameters = NewStorageError("parameters are not valid")
 
 	// ErrMissingEntities is an error that occurs when an associated immutable
 	// entity doesn't exist in the database. This error can indicate a wrong
 	// implementation or corrupted database.
-	ErrMissingEntities = errors.New("database: associated immutable entities are missing in the database")
+	ErrMissingEntities = NewStorageError("associated immutable entities are missing in the database")
```
KeyboardNerd, line 42: one by one we're moving to use special error type for the database.

database/models.go
```diff
@@ -155,18 +155,33 @@ type Namespace struct {
 	VersionFormat string
 }
 
+func NewNamespace(name string, versionFormat string) *Namespace {
```
KeyboardNerd, line 158: we're moving to use the proper initializer for the database structs

database/pgsql/complex_test.go
```diff
@@ -137,65 +132,39 @@ func TestConcurrency(t *testing.T) {
...
 	}
 	defer store.Close()
-
 	nsFeatures, vulnerabilities := testGenRandomVulnerabilityAndNamespacedFeature(t, store)
+	tx, err := store.Begin()
+	if !assert.Nil(t, err) {
+		t.FailNow()
+	}
 
-	fmt.Printf("%d features, %d vulnerabilities are generated", len(nsFeatures), len(vulnerabilities))
```
KeyboardNerd, line 160: Remove the concurrent insertion because the vulnerabilities <-> feature relationships are cached after the features are inserted into the database

KeyboardNerd (commented):
comments for PR reviewer

database/models.go
```diff
@@ -179,6 +194,11 @@ type NamespacedFeature struct {
 	Namespace Namespace
 }
 
+func NewNamespacedFeature(namespace *Namespace, feature *Feature) *NamespacedFeature {
+	// TODO: namespaced feature should use pointer values
```
jzelinskie, line 198: Is it only desirable to use pointers here to save memory?

database/pgsql/complex_test.go
```diff
@@ -137,65 +132,39 @@ func TestConcurrency(t *testing.T) {
...
-			Name:          uuid.New(),
-			VersionFormat: "dpkg",
-		}
-	}
-	return r
-}
-
 func TestCaching(t *testing.T) {
 	store, err := openDatabaseForTest("Caching", false)
 	if !assert.Nil(t, err) {
```
jzelinskie: we should just use `require` here instead of assert

database/pgsql/complex_test.go
```diff
@@ -137,65 +132,39 @@ func TestConcurrency(t *testing.T) {
...
 		t.FailNow()
 	}
 	defer store.Close()
-
 	nsFeatures, vulnerabilities := testGenRandomVulnerabilityAndNamespacedFeature(t, store)
```
jzelinskie, line 140: I'd actually prefer to keep the newline above this to separate the initialization of the storage from the rest of the setup

database/pgsql/complex_test.go
```diff
@@ -137,65 +132,39 @@ func TestConcurrency(t *testing.T) {
...
+	require.Nil(t, tx.InsertVulnerabilities(vulnerabilities))
+	if err := tx.Commit(); err != nil {
+		panic(err)
+	}
 
-	tx, err := store.Begin()
+	tx, err = store.Begin()
 	if !assert.Nil(t, err) {
 		t.FailNow()
 	}
```
jzelinskie: I'm pretty sure assert->failnow() is equivalent to just using `require`

database/pgsql/feature.go
```diff
@@ -240,62 +246,39 @@ func (tx *pgSession) PersistNamespacedFeatures(features []database.NamespacedFea
...
+			vulnerableFeatures[i].NamespacedFeature = features[i]
 		}
 	}
 
-	rows, err := tx.Query(searchNamespacedFeaturesVulnerabilities, pq.Array(toQuery))
+	rows, err := tx.Query(searchNamespacedFeaturesVulnerabilities, pq.Array(featureIDs))
 	if err != nil {
 		return nil, handleError("searchNamespacedFeaturesVulnerabilities", err)
 	}
+
```
jzelinskie: keep the defer in the same block as the query

database/pgsql/migrations/00001_initial_schema.go
```diff
@@ -159,13 +159,18 @@ var (
...
 
+			// vulnerability_affected_feature is a de-normalized table to store
+			// the affected features in a independent place other than the
+			// feature table to reduce table lock issue, and makes it easier for
+			// decoupling updater and the Clair main logic.
 			`CREATE TABLE IF NOT EXISTS vulnerability_affected_feature (
 				id SERIAL PRIMARY KEY, 
 				vulnerability_id INT REFERENCES vulnerability ON DELETE CASCADE,
 				feature_name TEXT NOT NULL,
+				feature_type INT NOT NULL REFERENCES feature_type ON DELETE CASCADE,
```
jzelinskie, line 170: doesn't this mean if the last package of a particular kind gets deleted, so will that entry in the feature_type table?
e.g. if the last source package is deleted `source` ceases to be a package type

jzelinskie (changes requested):

database/models.go
KeyboardNerd: I think it's idiomatic to use the pointer value for structs

database/pgsql/migrations/00001_initial_schema.go
KeyboardNerd: I think feature_type here is the parent table, so it's the other way around

KeyboardNerd: https://www.postgresql.org/docs/9.5/ddl-constraints.html
> CASCADE specifies that when a referenced row is deleted, row(s) referencing it should be automatically deleted as well.

row in vulnerability_affected_feature is referencing the feature_type table.
so when a feature type is deleted, we delete the row referencing that.

jzelinskie (approved):

PR #680, opened 2019-01-02
Replace arrays with slices

Allda:
This solution should be easier to read and maintain because there is no need of using indexes.

NOTE: There are few more places which use indexes. If this PR will be approved I can refactor rest of codebase.

### DIFF
--- a/database/pgsql/feature.go
+++ b/database/pgsql/feature.go
@@ -76,11 +76,9 @@ func (tx *pgSession) PersistFeatures(features []database.Feature) error {
 	})
 
 	// TODO(Sida): A better interface for bulk insertion is needed.
-	keys := make([]interface{}, len(features)*3)
-	for i, f := range features {
-		keys[i*3] = f.Name
-		keys[i*3+1] = f.Version
-		keys[i*3+2] = f.VersionFormat
+	keys := make([]interface{}, 0, len(features)*3)
+	for _, f := range features {
+		keys = append(keys, f.Name, f.Version, f.VersionFormat)
 		if f.Name == "" || f.Version == "" || f.VersionFormat == "" {
 			return commonerr.NewBadRequestError("Empty feature name, version or version format is not allowed")
 		}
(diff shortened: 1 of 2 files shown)


jzelinskie:
I suspect that based on the comment above the code:
```go
// TODO(Sida): A better interface for bulk insertion is needed.
```
This is done for performance reasons. `append` is going to allocate whenever adding the new value will exceed the existing `cap`, then it'll double the size of the slice.

The existing code allocates only once.

jzelinskie:
Btw, we can get the best of both worlds with code like this:
```go
keys := make([]interface{}, len(features)*3)
for _, f := range features {
  keys = append(keys, f.Name, f.Version, f.VersionFormat)
  ...
}
```

database/pgsql/layer.go
```diff
@@ -235,12 +235,13 @@ func (tx *pgSession) persistLayerFeatures(features []dbLayerFeature) error {
...
 		return features[i].featureID < features[j].featureID
 	})
-
-	keys := make([]interface{}, len(features)*3)
-	for i, feature := range features {
-		keys[i*3] = feature.layerID
-		keys[i*3+1] = feature.featureID
-		keys[i*3+2] = feature.detectorID
+	var keys []interface{}
+	// keys := make([]interface{}, len(features)*3)
```
jzelinskie: why comment this out? this is probably what we want to keep rather than `var keys []interface{}`, which introduces the possibility of a nil slice.

database/pgsql/feature.go
```diff
@@ -76,11 +76,12 @@ func (tx *pgSession) PersistFeatures(features []database.Feature) error {
...
 
 	// TODO(Sida): A better interface for bulk insertion is needed.
-	keys := make([]interface{}, len(features)*3)
-	for i, f := range features {
-		keys[i*3] = f.Name
-		keys[i*3+1] = f.Version
-		keys[i*3+2] = f.VersionFormat
+	var keys []interface{}
+	for _, f := range features {
+		keys = append(keys,
```
jzelinskie: Let's keep the append on one line.

jzelinskie (changes requested):
I think we can get the best of both worlds by keeping the usage of `make`, followed by using `append`.

```go
keys := make([]interface{}, len(features)*3)
for _, f := range features {
  keys = append(keys, f.Name, f.Version, f.VersionFormat)
  ...
}
```

Allda:
This doesn't seem to work because make([]interface{}, len(features)*3) creates a slice with default values and append() adds items at the end.

jzelinskie:
Ah, you probably have to do `make([]interface{}, 0, len(features)*3)` so that you set the capacity and not the length.

Allda:
Now it works fine after setting just capacity.

jzelinskie (approved):

PR #681, opened 2019-01-02
Vulnsrc rhel: handle "none" CVE impact

Allda:
Some RHEL CVEs [1] contains "none" string in impact field. This is throwing
warning message when fetching vulnerabilities. The new code handles this
case and it uses advisory severity instead.

[1] https://www.redhat.com/security/data/oval/com.redhat.rhsa-20080038.xml


jzelinskie (approved):
LGTM. Thanks!

Issue #185, opened 2016-05-24
Map Source Packages => Binary Packages


Quentin-M:
In some cases, the Feature names we need to extract to match vulnerabilities isn't the most meaningful for end-users. For instance, when `libssl-dev` is installed, we currently stores `openssl` (the source name), that's how we are able to match its vulnerabilities. However, when it comes to fix the potential issues, reading `libssl-dev` is much more valuable. Thus, it would be nice to have a new field to store that.

The way I see it, we can either add the field to `FeatureVersion` or to `Layer_diff_FeatureVersion`. We can't add it to `Feature` directly because during vulnerability updates, these names are unknown.

jzelinskie:
I've renamed this issue.

We need to document this, too. It's pretty unintuitive that Clair operates on **source packages** which may not correlate one-to-one with the packages users interact with via package managers.

KeyboardNerd:
To make it clear, Clair doesn't operate solely on *source packages*. For example, *rpm* extension extracts the installed binary package name as the feature name. This is due to the fact that OVAL provides vulnerabilities on binary packages. https://github.com/coreos/clair/blob/master/ext/featurefmt/rpm/rpm.go#L64 

In Clair V2, Clair feature does not represent accurate package information because the purpose is to be associated with the vulnerabilities. Therefore, using feature name as package name is actually misleading to the users.

In Clair V3, in our road map, the API response for getting Ancestry ( Image ) vulnerabilities will actually return the installed package names, which is the goal of this issue.

KeyboardNerd:
part 1 differentiate the feature types: #672

ldelossa:
@KeyboardNerd I'm starting to work on this issue. 

As I see it the desire here is to link a CVE with source information to a Binary package. In order to do this during feature detection we need to parse the source package associated with a binary package. I see that RPM is already doing this in the code. Do I have that correct ?

KeyboardNerd:
This source package / binary package information is useful because some vulnerability sources provide vulnerabilities on "source packages" (dpkg) and others on "binary packages" (RPM if I remember correctly).
It's still unclear to us if marking a binary package with the vulnerabilities on source package is a good idea since I don't quite know about how does vulnerabilities are announced.

ldelossa:
Closing this as it's handled in ClairV4.

PR #682, opened 2019-01-03
features: Store source package to database

Allda:
Recent commit extracted source packages from images, but it didn't store
the new metadata to the database. This caused inconsistent data model in
several places and the current master branch was broken.

This commit store source package name and version to feature DB table.

### DIFF
--- a/database/models.go
+++ b/database/models.go
@@ -169,6 +169,20 @@ type Feature struct {
 	VersionFormat string
 }
 
+// Valid checks if the Feature Layer is compliant to the spec.
+func (f *Feature) Valid() bool {
+	if f == nil {
+		return false
+	}
+
+	if f.Name == "" || f.Version == "" || f.SourceName == "" ||
+		f.SourceVersion == "" || f.VersionFormat == "" {
+		return false
+	}
+
+	return true
+}
+
 // NamespacedFeature is a feature with determined namespace and can be affected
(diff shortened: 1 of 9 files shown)


Allda:
@KeyboardNerd Any chance you could take a look at this PR sometime this week? Thank You

KeyboardNerd:
Adding source package column work is not finished because we had discussion to change the database design.

see work in progress: https://github.com/coreos/clair/pull/672
The above PR, once merged will change the database design from:
"adding feature source package name column" to
"adding layer wise feature relation table"
so that we can support different types of relationship that maybe introduced in the future.

The feature relationship is used to ask the following question:
> if a package is affected by specific vulnerability, will the related package assumed to be affected as well?

One example is that, if we only have a vulnerability database for source packages, not the binary packages, we can infer that the binary package has the same set of vulnerabilities as the source package. This inference model was implemented in one of the feature detector, and this is not really trivial to anyone. We want to explicitly program and extend the model: https://github.com/coreos/clair/issues/185

Since we want our model to record the relationship as accurate as possible, we can add the relation not only binary -> source package relationship. And so, that's the intention of our decision to change the DB design.

I'll reject this PR since it will not work in the future. 
We're trying our best to finish the above PR I mentioned in order to fix the current broken HEAD.

PR #688, opened 2019-01-15
Add auth token field for notifier as alternative authentication option

shudipta:
We are using Clair to build a Docker image scanner for Kubernetes. Our image scanner behaves as a Kubernetes extended apiserver. We run Clair as a Deployment. When Clair tries to send notification to our server, it gets 401 unauthorized error. We want to use the service account token mounted in the Clair server pod as the token for our extended apiserver. This pr adds the option to use a tokenfile (token file path) or token itself for the notifier section.

cc: @tamalsaha

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -83,5 +83,8 @@ clair:
       keyfile: 
       certfile: 
 
+      tokenfile:
+      token:
+
       # Optional HTTP Proxy: must be a valid URL (including the scheme).
       proxy:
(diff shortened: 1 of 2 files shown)


jzelinskie:
Hey there, sorry about the delay in review.

I think it's reasonable to assume that users will want to attach more than just the Authorization header.
Why not just add a headers key to the config that ends up looking something like this:

```yaml
http:
  headers:
  - authorization:
    - Bearer asdfasdf
```

ldelossa:
@shudipta please re-open if you'd like to consider @jzelinskie suggestion and continue this work.

PR #691, opened 2019-01-17
Notification: Add Messagebus notification plugin

Allda:
Plugin sends notification to user using messagebus based on activemq.
Plugin is fully configurable using config file. It requires valid
brokers, certificates and destination topic.


Allda:
Sorry, this should not be PR in upstream.

PR #695, opened 2019-01-21
pgsql: fix unchecked error

saromanov:
Hi! I've just fixed unchecked error. Please review it


jzelinskie:
Just need to double check that this isn't intentional. Often code with caches ignores errors because caches are allowed to fail open.

KeyboardNerd (approved):
LGTM

KeyboardNerd:
This is not intentional because the following code depends on the cache variable, which if is corrupted will mess up the database.

KeyboardNerd:
I'll just merge it

PR #700, opened 2019-01-29
Make nodePort configurable with helm

anon6789:
This pull request adds option to set the NodePort to desired port. The port can be set with the following values file:

```
service:
   type: "NodePort"
   apiNodePort: 31111
   healthNodePort: 31112
```

Only when `Values.service.type` equals `"NodePort"` and `Values.service.apiNodePort` or `Values.service.healthNodePort` is specified the NodePort port will be set.


jzelinskie (approved):
LGTM!

PR #712, opened 2019-02-22
clair: Ancestry Builder

KeyboardNerd:
worker is removed, we have Ancestry builder! using builder pattern to construct an Ancestry.


analyzer.go
```diff
@@ -0,0 +1,147 @@
...
+	reason string
+}
+
+func (e *AnalyzeError) Error() string {
+	return e.reason
+}
+
+func newAnalyzeError(reason string) *AnalyzeError {
+	return &AnalyzeError{reason}
+}
```
jzelinskie, line 34: Any reason for this to be a struct and not a string?

```go
type AnalyzeError string

func (e AnalyzeError) Error() {
  return string(e)
}

var StorageError = AnalyzeError("failed to query the database")
```

ancestry.go
```diff
@@ -0,0 +1,283 @@
...
+		layerIndex: 0,
+		detectors:  detectors,
+		namespaces: make(map[database.Detector]*layerIndexedNamespace),
+		features:   make(map[database.Detector][]layerIndexedFeature),
+	}
+}
+
+// AddLeafLayer adds a leaf layer to the ancestry builder, and computes the
+// namespaced features.
+func (b *AncestryBuilder) AddLeafLayer(layer *database.Layer) error {
```
jzelinskie: Doesn't look like this can actually fail. No need to return an error.

ancestry.go
```diff
@@ -0,0 +1,283 @@
...
+// Every detector inspects a set of files for the features
+// therefore, if that set of files gives a different set of features, it
+// should replace the existing features.
+func (b *AncestryBuilder) addLayerFeatures(detector database.Detector, features []database.LayerFeature) {
+	if len(features) == 0 {
+		// TODO(sidac): we need to differentiate if the detector finds that all
+		// features are removed ( a file change ), or the package installer is
+		// removed ( a file deletion ), or there's no change in the file ( file
+		// does not exist in the blob ) Right now, we're just assuming that no
+		// change in the file because that's the most case.
```
jzelinskie: >... most **common** case.

ancestry.go
```diff
@@ -0,0 +1,283 @@
...
+			}
+
+			currentFeatures = append(currentFeatures, b.createLayerIndexedFeature(namespace, &features[i]))
+		}
+	}
+
+	b.features[detector] = currentFeatures
+}
+
+func (b *AncestryBuilder) updateNamespace(layerNamespace *database.LayerNamespace) {
```
jzelinskie, line 144: This could use a doc comment (even tho it's not exported). It could also probably use a better name, but honestly I just tried to think of one for a bit and couldn't.

database/detector.go
```diff
@@ -93,12 +93,12 @@ func (s DetectorType) Valid() bool {
 type Detector struct {
 	// Name of an extension should be non-empty and uniquely identifies the
 	// extension.
-	Name string
+	Name string `json:"Name"`
```
jzelinskie: json/javascript is always camelCase

blob.go
```diff
@@ -0,0 +1,61 @@
+package clair
```
jzelinskie, line 15: copyright!

blob.go
```diff
@@ -0,0 +1,61 @@
...
+
+	log "github.com/sirupsen/logrus"
+)
+
+func retrieveLayerBlob(ctx context.Context, blobSha256 string, path string, headers map[string]string) (io.ReadCloser, error) {
+	if strings.HasPrefix(path, "http://") || strings.HasPrefix(path, "https://") {
+		return downloadLayerBlob(ctx, blobSha256, path, headers)
+	}
+
+	return loadLayerBlobFromFS(blobSha256)
```
jzelinskie: Is blobaSha256 supposed to be `path`? Also, may as well just `return os.Open(path)` here.

blob.go
```diff
@@ -0,0 +1,61 @@
...
+
+	client := &http.Client{Transport: tr}
+	r, err := client.Do(request)
+	if err != nil {
+		log.WithError(err).Error("could not download layer")
+		return nil, RetrieveBlobError
+	}
+
+	// Fail if we don't receive a 2xx HTTP status code.
+	if is2xx(r.StatusCode) {
```
jzelinskie: `httputil.is2xx(r)`

blob.go
```diff
@@ -0,0 +1,61 @@
...
+
+func retrieveLayerBlob(ctx context.Context, blobSha256 string, path string, headers map[string]string) (io.ReadCloser, error) {
+	if strings.HasPrefix(path, "http://") || strings.HasPrefix(path, "https://") {
+		return downloadLayerBlob(ctx, blobSha256, path, headers)
+	}
+
+	return loadLayerBlobFromFS(blobSha256)
+}
+
+func downloadLayerBlob(ctx context.Context, blobSha256 string, uri string, headers map[string]string) (io.ReadCloser, error) {
```
jzelinskie: This function doesn't use blobSha256.
We could also move this into httputil and call it 
```go
httputil.Get(ctx context.Context, uri string, headers http.Header) (io.ReadCloser, error)
```

We should also use the context in the request: https://godoc.org/net/http#Request.WithContext

ancestry.go
```diff
@@ -0,0 +1,283 @@
...
+		// does not exist in the blob ) Right now, we're just assuming that no
+		// change in the file because that's the most case.
+		return
+	}
+
+	existingFeatures := b.features[detector]
+	currentFeatures := make([]layerIndexedFeature, 0, len(features))
+	// Features that are not in the current layer should be removed.
+	for i := range existingFeatures {
+		feature := existingFeatures[i]
```
jzelinskie, line 106: rename this `existingLayerIndexedFeature`

We don't want full on hungarian notation, but we also don't want everything to be called `feature`

ancestry.go
```diff
@@ -0,0 +1,283 @@
...
+
+	return nil
+}
+
+// IsAncestryCached checks if the ancestry is already cached in the database with the current set of detectors.
+func IsAncestryCached(store database.Datastore, name string, layerHashes []string) (bool, error) {
+	if name == "" {
+		// TODO(sidac): we'll use the computed ancestry name in the future.
+		// During the transition, it still requires the user to use the correct
+		// ancestry name.
```
jzelinskie, line 269: How big of a change do you think this will be? Why not just strip it out of the API right now?

analyzer.go
```diff
@@ -0,0 +1,147 @@
...
+	return layer, nil
+}
+
+// EnabledDetectors retrieves a list of all detectors installed in the Clair
+// instance.
+func EnabledDetectors() []database.Detector {
+	return append(featurefmt.ListListers(), featurens.ListDetectors()...)
+}
+
+// Init populates the database with registered detectors.
```
jzelinskie: Rename this `RegisterConfiguredDetectors`. Init isn't very descriptive, and I consider things that need to be `init`ed to be a bad pattern, rather than having constructors that provide initialized things.

jzelinskie (changes requested):

blob.go
KeyboardNerd: yeah, I had the intention to checksum the result blob. Need to clean that up.

ancestry.go
KeyboardNerd: if it's real hungarian notation it'll be: `sliceElif` lol

ancestry.go
KeyboardNerd: Because when getting the ancestry, it requires a name.
We can either do: 
return the computed name to the client
or
the client still gives us the name, but we use an extra table to keep the client given name to the internal ancestry name mapping.

blob.go
KeyboardNerd: travis should fail when no copyright header, I'll add that todo

analyzer.go
KeyboardNerd: not really

ancestry.go
KeyboardNerd: yeah, the error was intended for if a namespace can't be determined for a feature, but that may happen very often. I'm just logging that for now.

blob.go
KeyboardNerd: oops

KeyboardNerd: TODO: add blob_test.go

ancestry.go
jzelinskie: I think we should get rid of names and make the computed name an implementation detail. Just let the customer provide the layer list.

KeyboardNerd: OR (which I prefer)
The user always query with a list of layer hashes in the request body. This just requires us to change the API a little bit.

KeyboardNerd: exactly

jzelinskie (approved):

PR #714, opened 2019-02-25
(WIP) Update gRPC and gRPC gateway

KeyboardNerd:
- Add the missing the gRPC and gRPC gateway versions.
- Add script to build the v3 api using versioned dependencies (vendor directory)


KeyboardNerd:
We need to eventually get rid of the vendor directory! 
Why not just use the Dockerfile to build the vendor folder?

KeyboardNerd:
gRPC server breaks after this PR
However, `all checks have passed`
need to add test on the muxedserver to start the server correctly

ldelossa:
@KeyboardNerd re-open if you plan on finalizing this work.

PR #720, opened 2019-02-27
clair: Fix namespace update logic

KeyboardNerd:
Only when a namespace's version is updated, we understand and replace
the namespace instead of using detector as the indicator.

### DIFF
--- a/ancestry.go
+++ b/ancestry.go
@@ -17,6 +17,7 @@ package clair
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


ancestry.go
```diff
@@ -143,64 +143,112 @@ func (b *AncestryBuilder) addLayerFeatures(detector database.Detector, features
...
-			introducedIn: b.layerIndex,
+	newNSNames := strings.Split(layerNamespace.Name, ":")
+	if len(newNSNames) != 2 {
+		log.Error("invalid namespace name")
+	}
+
+	newNSName := newNSNames[0]
+	newNSVersion := newNSNames[1]
+	for i, ns := range b.namespaces {
+		nsNames := strings.Split(ns.Namespace.Name, ":")
```
KeyboardNerd, line 158: not tracking the namespace name and version separately is a bad idea : (

jzelinskie: At least in the meantime, we could have utility functions that encapsulate this.

```go
nsName, nsVersion := nsutil.Split(layerNamespace.Name)

...

func Split(nsString string) (name, version) {
  strs := strings.Split(nsString, ":")
  if len(strs) != 2 {
    panic("invalid namespace name")
  }
  return strs[0], strs[1]
}
```

jzelinskie (approved):
I don't think my suggestion for the ns splitting needs to be done in this PR. This LGTM

PR #722, opened 2019-02-28
featuturefmt: Extract PotentialNamespace

### LINKED ISSUE #706: Enable multiple namespaces for installed layer feature
Hi,
right now Clair consumes Red Hat's Oval security data which covers only a limited subset of RH's products (RHEL only). To cover the whole RH portfolio Clair needs to switch from Oval to VMaaS security data. The pull request for switching already exists (https://github.com/coreos/clair/pull/697), but it also filters RHEL product only. The filter cannot be removed now because of the following reasons:

Each vulnerability and the affected package is tight to a specific product. The product is represented by CPE string and you can see an example in this file (https://www.redhat.com/security/data/metrics/cvemap.xml )

The CPE can be stored as a namespace when fetching new vulnerabilities. On the other way when Clair scans layer and extract features, it could also extract CPEs from an image. The CPEs will be stored as a file in the layer's filesystem. The only problem is that there can be multiple CPEs per layer because packages installed in a given layer can come from multiple products (you can imagine CPEs as yum repositories).

So my proposal would be to enable multiple namespaces per layer's feature. The database model already enables it (there is N:M relation in namespaced_feature table), but the internal logic only allows one namespace per layer's feature.

### Expected Outcome
Enable multiple namespaces per layer's feature. This would enable to correctly scan Red Hat's images.

### Actual Outcome
Clair currently supports only one namespace per layer's feature

### Environment

- Clair version/image:  v2.0.7

Allda:
PotentialNamespace is feature namespace extracted while detecting features in a layer. The PotentialNamespace is stored in nemaspace table + reference is also stored in layer_feature.

This will allow having multiple namespaces for single layer feature.

Partially resolves: https://github.com/coreos/clair/issues/706


database/models.go
```diff
@@ -166,22 +170,23 @@ func NewNamespace(name string, versionFormat string) *Namespace {
...
 type Feature struct {
-	Name          string      `json:"name"`
-	Version       string      `json:"version"`
-	VersionFormat string      `json:"versionFormat"`
-	Type          FeatureType `json:"type"`
+	Name               string      `json:"name"`
+	Version            string      `json:"version"`
+	VersionFormat      string      `json:"versionFormat"`
+	Type               FeatureType `json:"type"`
+	PotentialNamespace Namespace   `json:"potentialNamespace"`
```
KeyboardNerd: Please move PotentialNamespace to LayerFeature struct, because Feature is shared by multiple layers.

Allda: Feature listers return `database.Feature` and not `database.LayerFeature`. Can I change the return type of listers or how can I extract namespace?

KeyboardNerd: yes, please change the feature lister to return LayerFeature.

Allda:
@KeyboardNerd The PR is ready to be reviewed. 

Here is the feature lister which uses a potential namespace which I used for testing.
https://gist.github.com/Allda/99d30f5d37d35dd3e2d0244470d6f150

There is one thing which is causing problems now and I believe it will be resolved once you made your changes in updater.go

When feature lister detects multiple namespaces (cpes) it returns each package multiple times with different CPE. It works fine, store namespaces, features, layer_feature, but it fails when storing data to ancestry_feature because of duplicated reference to namespaced_feature_id.

Allda:
I found one more bug, wait with a review, please.

Allda:
> I found one more bug, wait with a review, please.

Fixed

database/pgsql/migrations/00001_initial_schema.go
```diff
@@ -89,7 +89,8 @@ var (
 				layer_id INT REFERENCES layer ON DELETE CASCADE, 
 				feature_id INT REFERENCES feature ON DELETE CASCADE,
 				detector_id INT REFERENCES detector ON DELETE CASCADE,
-				UNIQUE (layer_id, feature_id));`,
+				namespace_id INT REFERENCES namespace ON DELETE CASCADE,
```
KeyboardNerd: please explicitly state the namespace_id is nullable

database/pgsql/layer.go
```diff
@@ -37,10 +37,11 @@ const (
 		SELECT id FROM layer WHERE hash = $1`
 
 	findLayerFeatures = `
-		SELECT f.name, f.version, f.version_format, t.name, lf.detector_id
-			FROM layer_feature AS lf, feature AS f, feature_type AS t
+		SELECT f.name, f.version, f.version_format, t.name, lf.detector_id, ns.name, ns.version_format
+			FROM layer_feature AS lf, feature AS f, feature_type AS t, namespace AS ns
 			WHERE lf.feature_id = f.id
 				AND t.id = f.type
+				AND lf.namespace_id = ns.id
```
KeyboardNerd: what if ls.namespace_id is null? there won't be a matched ns.id.

database/dbutil.go
```diff
@@ -58,6 +58,19 @@ func ConvertFeatureSetToFeatures(features mapset.Set) []Feature {
 	return uniqueFeatures
 }
 
+func ConvertFeatureSerToLayerFeatures(features mapset.Set) []LayerFeature {
```
KeyboardNerd: typo

KeyboardNerd:
Schema looks ok.

Please add some tests on the change, especially the features with potential namespace.

database/pgsql/layer.go
Allda: Resolved by using Left join

database/pgsql/testdata/data.sql
```diff
@@ -2,7 +2,8 @@
 INSERT INTO namespace (id, name, version_format) VALUES
   (1, 'debian:7', 'dpkg'),
   (2, 'debian:8', 'dpkg'),
-  (3, 'fake:1.0', 'rpm');
+  (3, 'fake:1.0', 'rpm'),
+  (4, 'cpe:/o:redhat:enterprise_linux:7::server', 'rpm');
```
KeyboardNerd, line 6: Does 'cpe:/o:redhat:enterprise_linux:7::server' has any version information?

If this is the actual format for CPE, I'll have to change the namespace db a bit because right now we assume namespace name consists of two components: `<name>:<version>` as a string.

KeyboardNerd: Also, please rebase to master, and test. We found a bug in our tests that prevents database test failure from showing up.

Allda: The version is usually included in CPE (in this case it is 7), but from the definition of CPE the version can be omitted.

https://access.redhat.com/blogs/766093/posts/2387501

Allda: Yes, I noticed the bug. Today I spend a few hours trying to figure out why my test is not failing and after I found the cause and I fixed it I noticed that you fixed it yesterday

database/pgsql/feature.go
```diff
@@ -359,13 +360,14 @@ func (tx *pgSession) findFeatureIDs(fs []database.Feature) ([]sql.NullInt64, err
 		return nil, err
 	}
 
-	fMap := map[database.Feature]sql.NullInt64{}
+	fMap := map[string]sql.NullInt64{}
```
KeyboardNerd: why do we need the string as key?

database/pgsql/layer_test.go
```diff
@@ -135,7 +157,8 @@ func TestPersistLayer(t *testing.T) {
 				layer, ok, err := tx.FindLayer(test.name)
 				assert.Nil(t, err)
 				assert.True(t, ok)
-				database.AssertLayerEqual(t, test.layer, &layer)
+				result := database.AssertLayerEqual(t, test.layer, &layer)
```
KeyboardNerd: it already has the assertion inside.

KeyboardNerd (commented):
A few points, I'll do extra testing today.

database/pgsql/feature.go
Allda: That's probably forgotten code when potential namespace was part of Feature and not LayerFeature. I will fix it.

KeyboardNerd (approved):
lgtm

PR #723, opened 2019-02-28
database: make locks SOI & add Extend method

jzelinskie:


database/pgsql/lock.go
```diff
@@ -19,71 +19,77 @@ import (
...
+	var (
+		desiredLockedUntil = time.Now().Add(desiredDuration)
+
+		lockedUntil time.Time
+		lockOwner   string
+	)
+
 	defer observeQueryTime("Lock", "soiLock", time.Now())
-	_, err := tx.Exec(soiLock, name, owner, until)
+	err := tx.QueryRow(soiLock, lockName, whoami, desiredLockedUntil).Scan(&lockOwner, lockedUntil)
```
KeyboardNerd: do we need `&lockedUntil`?

jzelinskie: In the case where you don't get the lock, you want the time it'll expire, so you can try and grab it again around then.

edit: woops forgot the `&`

KeyboardNerd (approved):

PR #727, opened 2019-03-06
pgsql: Fix pgsql test

KeyboardNerd:


jzelinskie (approved):
We can fix everything that's broken in another PR.

Issue #734, opened 2019-03-08
Support Potential Namespace in Ancestry builder


KeyboardNerd:
### Description of Problem / Feature Request

If a potential namespace exists in a feature, it should be used as the feature's namespace instead of doing the look up.

### Environment

- Clair version/image: master
- Clair client name/version: 
- Host OS: 
- Kernel (e.g. `uname -a`): 
- Kubernetes version (use `kubectl version`): 
- Helm version (use `helm version`): 
- Network/Firewall setup:

KeyboardNerd:
Background:
In the Ancestry Builder, every `layerIndexedFeature` holds a pointer to the `layerIndexedNamespace`. 

Why pointer? When we detect a namespace is upgraded, every `layerIndexedFeature` referencing to that namespace as in `layerIndexedNamespace` will be updated automatically.
[Reference](https://github.com/coreos/clair/blob/master/ancestry.go#L144)

To implement this feature,
[lookupNamespace](https://github.com/coreos/clair/blob/master/ancestry.go#L212) defines the how the namespace of a feature is found. 

Currently, we use the versionformat ( package manager ) and the mapping from namespace to package manager ( e.g. Ubuntu has dpkg, and therefore, all dpkg packages are under ubuntu ), to look up the namespace we assume to be correct.

Now, the new feature is that every `layerFeature` has a `PotentialNamespace`, so, the logic goes as following: 
1. When a layer feature has the potential namespace, we are going to use that namespace.
2. In Ancestry Builder, first check if the potential namespace is already existing or not in (AncestryBuilder.namespaces)[https://github.com/coreos/clair/blob/master/ancestry.go#L45]
3. If no, we first insert the `PotentialNamespace` to `AncestryBuilder.namespaces`, the layer index will be the same as the layer index of the feature. We can omit the detector field because it doesn't make sense.
4. We use the existing `layerIndexedNamespace` in `AncestryBuilder.namespaces`. 

Then add some tests in `ancestry_test.go` with that PotentialNamespace thing: https://github.com/coreos/clair/blob/master/ancestry_test.go

PR #738, opened 2019-03-14
Use PotentialNamespace in ancestry

Allda:
If PotentialNamespace is available for a given feature it is used instead
of detector namespace. A detector is empty in such a case.

Resolves: https://github.com/coreos/clair/issues/734

### DIFF
--- a/ancestry.go
+++ b/ancestry.go
@@ -211,9 +211,20 @@ func (b *AncestryBuilder) createLayerIndexedFeature(namespace *layerIndexedNames
 
 func (b *AncestryBuilder) lookupNamespace(feature *database.LayerFeature) (*layerIndexedNamespace, bool) {
 	matchedNamespaces := []*layerIndexedNamespace{}
-	for i, namespace := range b.namespaces {
-		if namespace.Namespace.VersionFormat == feature.VersionFormat {
-			matchedNamespaces = append(matchedNamespaces, &b.namespaces[i])
+	if feature.PotentialNamespace.Name != "" {
+		a := &layerIndexedNamespace{
+			Namespace: database.LayerNamespace{
+				Namespace: feature.PotentialNamespace,
+			},
+			IntroducedIn: b.layerIndex,
+		}
+		matchedNamespaces = append(matchedNamespaces, a)
+	} else {
+
+		for i, namespace := range b.namespaces {
+			if namespace.Namespace.VersionFormat == feature.VersionFormat {
+				matchedNamespaces = append(matchedNamespaces, &b.namespaces[i])
+			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


Allda:
@KeyboardNerd I tested the PR with real data which uses PotentialNamespace and results were exactly as I expected.

ancestry.go
```diff
@@ -211,9 +211,20 @@ func (b *AncestryBuilder) createLayerIndexedFeature(namespace *layerIndexedNames
...
-		if namespace.Namespace.VersionFormat == feature.VersionFormat {
-			matchedNamespaces = append(matchedNamespaces, &b.namespaces[i])
+	if feature.PotentialNamespace.Name != "" {
+		a := &layerIndexedNamespace{
+			Namespace: database.LayerNamespace{
+				Namespace: feature.PotentialNamespace,
+			},
+			IntroducedIn: b.layerIndex,
+		}
+		matchedNamespaces = append(matchedNamespaces, a)
```
KeyboardNerd, line 221: We should always ensure that all elements in matched namespaces are in the ancestrybuilder.namespaces, and any feature with the same namespace must point to the same namespace object in the ancestrybuilder.namespaces slice. 

Otherwise, this will break the updateNamespace logic. 

BTW, in order to use the CPE string as a namespace, please remove the panic condition for updateNamespace logic.

KeyboardNerd: ^^^ 
This may not be the case.
I think for now at least: when a feature has a potential namespace, we will use that namespace directly instead of getting into the namespace lookup and upgrade process. So that the output AncestryFeature is guaranteed to have the namespace.

@jzelinskie What do you think about this logic here?

jzelinskie: I think this logic is fine for now. I don't want to put the cart before the horse. Let's get things working and then test our assumptions later.

KeyboardNerd (changes requested):
Please also update the convert.go DetectorFromDatabaseModel function. Right now I think the API may fail because the detector is empty.

Allda:
> Please also update the convert.go DetectorFromDatabaseModel function. Right now I think the API may fail because the detector is empty.

API doesn't fail when a detector is empty.

Here is an example output:
`
{
  "name": "gnupg2",
  "namespace": {
    "name": "cpe:/a:redhat:cloudforms_managementengine:5.8::el7",
    "detector": {}
  },
  "version": "2.0.22-5.el7_5",
  "version_format": "rpm",
  "detector": {
    "name": "redhat",
    "version": "1.0",
    "dtype": "DETECTOR_D_TYPE_FEATURE"
  },
  "feature_type": "binary"
},`

KeyboardNerd (approved):

PR #740, opened 2019-03-20
Detect vulnerabilities in NPM-installed packages

### LINKED ISSUE #750: NPM vulnerability scanner
### Feature Request

Clair should report on vulnerabilities in NodeJS packages installed using NPM (or Yarn or a similar tool).

### Expected Outcome

If I have built a Docker image which in the build phase runs `npm install`, but one of my dependencies has a known security issue, then it should be identified when it is scanned by Clair.

### Actual Outcome

Currently nothing (unless it's a system-level package)

### Environment

N/A

cnorthwood:
This PR detects installed NPM packages based on the presence of a `project.json` file and records them as a feature in the NPM namespace. It imports vulnerabilities from [the NodeJS security working group advisories repo](https://github.com/nodejs/security-wg/tree/master/vuln).

Fixes #750.


cnorthwood:
Hi there, I'm a new contributor and also a new Golang dev (this is my first time actually using the language), so any feedback is gracefully appreciated.

I'm also unclear how the namespaces feature is intended to work. Should these features be namespaced in someway? (perhaps as an "NPM" namespace?)

My next steps are to investigate adding the Node Security Advisories as a vulnerability feed into Clair. Would you prefer to have this as a separate PR, or would you rather have one large PR to review?

Thanks

KeyboardNerd:
Namespace is intended for differentiating the vulnerabilities for different platforms.
For example, a CVE report says it affects curl package in CentOS and Ubuntu images but not Windows because it's compiled in different way.

For your case of NPM, please look up the vulnerability data sources and see if it has platform information, for example, a package running on NodeJS v10.0 is affected by not NodeJS v11.0.
If there's no thing that differentiates the platform, we can just use NPM as the namespace.

cnorthwood:
thanks @KeyboardNerd, I'll have a look at how namespacing would work - having a look at the existing featurefmts (apk/dpkg/rpm), I can't see how they're namespaced (perhaps they're not?) and update the PR

KeyboardNerd:
Yes, that's because featurefmt only looks at the package installation information, which may not contain where the package is installed or compiled in. 

If we want all packages installed by NPM to be under NPM namespace, you can fill in the PotentialNamespace field in LayerFeature to be NPM namespace.
Otherwise, you need to somehow extract the namespace from layer blob by implementing a module in featurens.

cnorthwood:
> If we want all packages installed by NPM to be under NPM namespace

I guess the question is therefore, "do we want all packages installed by NPM to be under the NPM namespace?". I'm unclear on the wider context of namespaces so far, so whether or not it makes sense to use them.

I've also noticed that the current mechanism for asking for files to scan only returns those which `RequiredFilenames` ask for, so I've updated tarutil to support specifying those by regexp, as a NPM package could be installed anywhere in the layer, and have raised #749 to implement this feature, so this PR depends on that one.

cnorthwood:
okay, so I think I've written this, but I'm not 100% clear how to test it in situ, the database appears to be empty after the importer has run (but this includes the other reporters) so I'm not sure how to verify, as the v1 API appears to have gone which I think would have allowed me to see it.

The main thing I'm not clear on is the version logic for AffectedVersion and FixedInVersion. The NPM database specifies these as ranges, not distinct versions (e.g., a vulnerability could affect 5.4.2 and 6.1.3, but be fixed in 5.4.3 and 6.1.4. However, despite 5.4.3 having a fix, and 6.1.3 having a greater version, 6.1.3 should still be affected), so I'm not sure if the approach I've taken here fits the internal model of Clair or if more work is needed (and I suspect if more work is needed, that would require some of the domain concepts within Clair to change)

KeyboardNerd:
The ranged version support is initially implemented in versionfmt:
The interface has these two functions:
InRange(versionA, rangeB string) (bool, error) checks if versionA is in rangeB ( a string representing range(s) of versions )
GetFixedIn(rangeA string) (string, error) was intended to compute the version that the user should upgrade to in order to avoid the vulnerability, but maybe it's not necessary. 

You can implement a npm versioning scheme in versionfmt, and at the InRange function will work properly. 

In the AffectedFeature struct: 
AffectedVersion: should be the version string you want to put as rangeB in InRange function
FixedInVersion: if there's no such thing, you should just put an empty string. This is the recommended upgraded version available to the user. 
VersionFormat: should be npm version format's name.

cnorthwood:
I've updated the semver comparison to correctly match against the formats in the Node security advisories, and rebased against master. I believe this PR is ready for review now.

cnorthwood:
Hi @KeyboardNerd, this should be ready for review now - is there anything blocking from my side? Thanks

cnorthwood:
just chasing this up - is there anything blocking this that I can unblock to help get this reviewed/accepted?

KeyboardNerd:
@jzelinskie

ext/vulnsrc/npm/npm.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		vulns = append(vulns, vulnsInFile...)
+	}
+
+	return vulns, nil
+}
+
+func parseNpmVuln(filePath string) (vulns []database.VulnerabilityWithAffected, err error) {
+	var data []byte
```
ldelossa, line 79: @cnorthwood 

Since this will be ran in a loop. Can we attempt to do this in a streaming fashion instead of loading the full set of bytes into memory ? 

Opening a file descriptor to this file via `os.Open` https://golang.org/pkg/os/#Open. The FD implements io.Reader and io.Writer. You can then use  a json.Decoder for streaming: https://golang.org/pkg/encoding/json/#Decoder

ext/vulnsrc/npm/npm.go
```diff
@@ -0,0 +1,204 @@
...
+		ids = []string{"NSWG-ECO-" + strconv.Itoa(securityAdvisory.ID)}
+	}
+
+	for _, id := range ids {
+		vuln := database.VulnerabilityWithAffected{
+			Vulnerability: database.Vulnerability{
+				Name:        id,
+				Namespace:   ns,
+				Description: securityAdvisory.Overview,
+				Link:        "https://github.com/nodejs/security-wg/blob/master/vuln/npm/" + strconv.Itoa(securityAdvisory.ID) + ".json",
```
ldelossa, line 108: Nit - url should be constant.

ext/vulnsrc/npm/npm.go
```diff
@@ -0,0 +1,204 @@
...
+	if err = validSecurityAdvisory(securityAdvisory); err != nil {
+		return
+	}
+
+	var ids []string
+	// use CVE IDs if they exist
+	if len(securityAdvisory.CVEs) > 0 {
+		ids = securityAdvisory.CVEs
+	} else {
+		ids = []string{"NSWG-ECO-" + strconv.Itoa(securityAdvisory.ID)}
```
ldelossa, line 99: nit - NSWG-ECO- should be constant.

ldelossa (changes requested):
@cnorthwood would like to change the json parsing to streaming. the others are just nits.

ext/vulnsrc/npm/npm.go
cnorthwood: I'll have a look at this, but am unlikely to have time to do so for a few weeks now. The JSON files are themselves pretty small so I didn't think the memory overhead a significant issue.

ldelossa: @cnorthwood fair enough, the consideration is more for thinking about scale.

ldelossa:
@cnorthwood @KeyboardNerd I'm attempting to test that this PR reports the correct vulnerabilities. The current implementation is not reporting effected packages. I'm using the following test docker container:

```
FROM node:12.7.0

RUN npm install -g concat-stream@1.3.1
RUN npm install -g angular-http-server@1.4.1
RUN npm install -g merge-objects@1.0.4
RUN npm install -g call@3.0.0
RUN npm install -g moment@2.11.1
```

Clair only reports vulnerabilities from debian:9 in this case. I will do some digging as to why. Immediate indication is no "namespaced_feature" records being created linking the features and their namespace. Features themselves are indexed correctly and namespace is created in the database.

cnorthwood:
Would be keen to get input from a maintainer on this, I'm not really clear on the inner architecture of Clair to make sure it's raising these issues in the expected manner, especially with namespacing which is fairly new

ldelossa:
@cnorthwood appreciate your work here. With the recent GA of clair 4.0 we reworked how matching occurs in general. 

Clair 4.0 does *not* have npm support, and it should. If you would still like to contribute and focus on Clair 4.0, NPM support would be a welcomed addition.

PR #741, opened 2019-03-20
v3: Analyze layer content in parallel

KeyboardNerd:
This resolves the performance issue we see.

### DIFF
--- a/api/v3/rpc.go
+++ b/api/v3/rpc.go
@@ -15,7 +15,10 @@
 package v3
 
 import (
+	"sync"
+
 	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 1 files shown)


api/v3/rpc.go
```diff
@@ -83,32 +84,43 @@ func (s *AncestryServer) PostAncestry(ctx context.Context, req *pb.PostAncestryR
...
+				return nil, status.Error(codes.InvalidArgument, "ancestry layer path should not be empty")
+			}
+
+			g.Go(func() error {
+				clairLayer, err := clair.AnalyzeLayer(analyzerCtx, s.Store, layer.Hash, req.Format, layer.Path, layer.Headers)
+				if err != nil {
+					return err
+				}
+
+				layerMap[layer.Hash] = clairLayer
```
jzelinskie, line 116: This needs a lock even though the keys are independent because of races when the runtime is expanding the size of the map.

Even if we added the size to the make,`layerMap := make(map[string]*database.Layer, len(req.Layers))`, I wouldn't trust this race without a lock.

jzelinskie (changes requested):

api/v3/rpc.go
KeyboardNerd: makes sense

jzelinskie (approved):

PR #742, opened 2019-03-20
set / as default path but allow for individual configuration in values.yml

alsokate:
The current hardcoded path `/` in the ingress template of the Helm charts does not work in the context of AWS ALBs.  The listener is explicit in its routing, and the path needs to be set to `/*` in order for any of the requests to Clair endpoints to succeed.  This is a fix that sets `/` as a default but allows for others using the same system to set a custom configuration without having to maintain their own copy of the Helm charts.

Viz:
(with default `/` path)
Screenshot 2019-03-19 10 33 19
Screenshot 2019-03-19 10 34 25

(with path configured to `/*` in a local, customized clone of the helm charts)
Screenshot 2019-03-19 10 28 17
Screenshot 2019-03-19 10 35 46
Screenshot 2019-03-19 10 35 56

### DIFF
--- a/contrib/helm/clair/templates/ingress.yaml
+++ b/contrib/helm/clair/templates/ingress.yaml
@@ -1,6 +1,7 @@
 {{- if .Values.ingress.enabled -}}
 {{- $serviceName := include "clair.fullname" . -}}
 {{- $servicePort := .Values.service.externalApiPort -}}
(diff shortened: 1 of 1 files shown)


jzelinskie:
Hey there.

This is actually some default boilerplate generated with all Helm Charts, I believe.
You might be able to make this case upstream as well.

jzelinskie (approved):

PR #748, opened 2019-03-21
Detect installed Python eggs and wheels as features

cnorthwood:


ldelossa:
@cnorthwood appreciate your work here. With the recent GA of clair 4.0 we reworked how python detection works. I'm going to close this as its becoming stale. If you're interested in further contribution let's focus on Clair 4.0 (current main branch)

PR #749, opened 2019-03-21
tarutil: allow file names to be specified by regexp

cnorthwood:
Some features do not exist in set locations, but can be anywhere in the layer.
This allows those featurefmt to specify the filenames to be scanned by
regexp, as opposed to purely by path prefix.
If any current users of this express paths which use regexp special characters
this could be a breaking change for them (with the exception of . which will
continue to work as it matches against the literal character .), however
none of the code in this repo does that.

### DIFF
--- a/pkg/tarutil/tarutil.go
+++ b/pkg/tarutil/tarutil.go
@@ -80,7 +80,7 @@ func ExtractFiles(r io.Reader, filenames []string) (FilesMap, error) {
 		// Determine if we should extract the element
 		toBeExtracted := false
 		for _, s := range filenames {
-			if match, err := regexp.MatchString("^"+s, filename); err == nil && match {
+			if match, err := regexp.MatchString(s, filename); err == nil && match {
 				toBeExtracted = true
 				break
 			}
--- a/pkg/tarutil/tarutil.go
+++ b/pkg/tarutil/tarutil.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os/exec"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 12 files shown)


pkg/tarutil/tarutil.go
```diff
@@ -78,7 +80,7 @@ func ExtractFiles(r io.Reader, filenames []string) (FilesMap, error) {
 		// Determine if we should extract the element
 		toBeExtracted := false
 		for _, s := range filenames {
-			if strings.HasPrefix(filename, s) {
+			if match, err := regexp.MatchString("^"+s, filename); err == nil && match {
```
jzelinskie: This leading `^` is implicit behavior and seems unnecessary.
If we have to change other files, I don't mind.

jzelinskie (changes requested):

pkg/tarutil/tarutil.go
cnorthwood: The `^` does change the semantics (https://play.golang.org/p/yZl7bJLzr22) - I included it to keep the behaviour in line with the previous behaviour of only checking for the prefix.

I don't think there are any changes needed to the other files as the other file definitions were simple prefixes and didn't contain any special characters, it was more a headsup in case there were other subtle implications.

KeyboardNerd: Since we're moving to use the regex, I think it makes sense to enforce all the implementations with `RequiredFile` implementation to use regex.

cnorthwood: okay, I've removed the `^` and updated all users to have a regexp instead in order to maintain previous behaviour

jzelinskie (approved):
LGTM! Thanks for this change.

PR #751, opened 2019-03-22
Exclude PotentialNamespace while comparing layer features

Allda:
PotentialNamespace needs to be excluded from comparison in
addLayerFeatures, because otherwise some features can be lost in higher
layer.

### DIFF
--- a/ancestry.go
+++ b/ancestry.go
@@ -101,12 +101,12 @@ func (b *AncestryBuilder) addLayerFeatures(detector database.Detector, features 
 	}
 
 	existingFeatures := b.features[detector]
-	currentFeatures := make([]layerIndexedFeature, 0, len(features))
+	currentFeatures := make([]layerIndexedFeature, 0, len(features)+len(existingFeatures))
 	// Features that are not in the current layer should be removed.
 	for i := range existingFeatures {
 		feature := existingFeatures[i]
 		for j := range features {
-			if features[j] == *feature.Feature {
+			if features[j].CompareWithoutNamespace(*feature.Feature) {
 				currentFeatures = append(currentFeatures, feature)
 				break
 			}
--- a/database/layer.go
+++ b/database/layer.go
@@ -63,3 +63,11 @@ type LayerFeature struct {
 	By                 Detector  `json:"by"`
 	PotentialNamespace Namespace `json:"potentialNamespace"`
 }
+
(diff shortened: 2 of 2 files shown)


ancestry.go
```diff
@@ -101,12 +101,12 @@ func (b *AncestryBuilder) addLayerFeatures(detector database.Detector, features
...
 
 	existingFeatures := b.features[detector]
-	currentFeatures := make([]layerIndexedFeature, 0, len(features))
+	currentFeatures := make([]layerIndexedFeature, 0, len(features)+len(existingFeatures))
 	// Features that are not in the current layer should be removed.
 	for i := range existingFeatures {
 		feature := existingFeatures[i]
 		for j := range features {
-			if features[j] == *feature.Feature {
+			if features[j].CompareWithoutNamespace(*feature.Feature) {
```
KeyboardNerd, line 109: I'm not sure how content set file works, when the RPM database changes, will the content set file change?

Allda: Each layer can add new content-sets, but it should always inherit content-sets from previous layers. So features should appear in layer when new content-set is added and if there are no changes in content-sets across layers features should be only in a layer where they were installed.

KeyboardNerd: oh, so:
layer 1 has content set file (rhel7), and added a RPM (curl), now scan result has curl under rhel7
layer 2 has no content set file, and added a RPM (wget), now scan result has curl and wget

layer 1 is the parent layer of layer 2

We want to preserve the curl to under rhel7, but wget should have no potential namespace.

Is that correct?

Allda: Yes, that's correct. Also if layer 2 introduces new content-set (for example rhel7.5) then curl will be shown in both layers (in layer 1 with rhel7 namespace and layer 2 with rhel7.5 namespace).

KeyboardNerd: Interesting, I think this logic is very unique to content-set stuff. I think for now, since there's only one use case of PotentialNamespace, the solution in this PR is fine.

However, since there are people working on NPM etc. that will use Potential Namespace, I'm wondering how to support this type of difference in handling the potential namespace.
@jzelinskie what do you think?

Allda:
@KeyboardNerd @jzelinskie hi, any update on this one?

Allda:
Closing this once since it is already part of another PR.

PR #752, opened 2019-03-22
Adding Openshift Template to contrib.

arocki7:
This template will be helpful for those who want to host Clair on Openshift.

### DIFF
--- a/contrib/openshift/README.md
+++ b/contrib/openshift/README.md
@@ -1,0 +1,36 @@
+# Openshift Template for Clair
+
+This template is to implement clair inside Openshift.  
+
+## What's inside
+
+Clair : quay.io/coreos/clair:latest  
+Postgres: Postgres 9.5 (Persistent Storage)  
+
+## How to Deploy
+
+- Login into Openshift Console.  
+- Click 'Import from YAML/JSON'  
+- Paste the contents of template.yml file.
+- Create > Change the parameters if needed.
+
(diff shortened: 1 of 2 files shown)


jzelinskie:
Nice work! I think this would be a good addition to the contrib directory.

However, there is a lot that could be reused between the Helm chart and this. It's a burden to have to maintain Kubernetes manifests in two different places.

I suspect we could leverage `helm template` in order to generate both a vanilla and OpenShift deployment from one set of maintained templates.

arocki7:
Hi @jzelinskie, thanks for the feedback.

Helm charts are not as per OpenShift template standards where you handle everything using a single template file. In OpenShift, engineers just import/save template using one template only. 

OpenShift has upstream version of Kubernetes objects such as DeploymentConfig and route. So, it won’t be possible to use OpenShift features (such as SSL termination) using standard Kubernetes objects in OpenShift.

I tried to use helm templates from contrib. But, it is quite complicated than this template.

My suggestion is to keep them separated.

jzelinskie:
Sorry for not making this clear: my suggestion was not that the OpenShift template be deployed as a Helm Chart, but rather that there is a build system for the Clair open source project that produces 2 artifacts: the OpenShift template and the Helm Chart. The point of doing so is to reduce the burden to maintain two sets of the very similar things. The Helm Chart has already had issues with not getting its config updated.

PR #755, opened 2019-04-03
Allow OpenShift user update CA bundle

Allda:
When users run Clair in OpenShift under arbitrary user and
they want to update CA bundle with their own certificate
they need to run `ca-certificates` which requires access
to /etc/ssl/certs.

This commit updates Dockerfile based on OpenShift guidelines to allow
update CA bundle in a system.

https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,11 @@ RUN export CLAIR_VERSION=$(git describe --tag --always --dirty) && \
 FROM alpine:3.8
 COPY --from=build /go/src/github.com/coreos/clair/clair /clair
 RUN apk add --no-cache git rpm xz ca-certificates dumb-init
+
+# change ownership of ssl directory to allow custom cert in OpenShift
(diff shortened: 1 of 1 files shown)


jzelinskie:
I'm not sure I'm comfortable with adding OpenShift specific fixes in the upstream project. Considering Red Hat only supports software built on a RHEL base image, the upstream Alpine-based image is already a no-go.

[Clair-JWT](https://github.com/quay/clair-jwt) is a supported version of Clair for running on OpenShift for Quay deployments, and it is in the process of moving towards a RHEL base image.

jzelinskie:
I'm not a huge fan of this, but I think it's reasonable with a comment in the Dockerfile so that it's not lost in the noise.

jzelinskie (approved):

Issue #730, opened 2019-03-07
Insufficient (very low) resource limits per default enabled in Helm chart


naibaf0:
### Description of Problem / Feature Request

In the Helm Chart there are quite the small resource limits set as default:
```yaml
resources:
  limits:
    cpu: 200m
    memory: 1500Mi
  requests:
    cpu: 100m
    memory: 500Mi
```

These limits are quite small and resulted in Clair regularly being killed in the Kubernetes Cluster. As I just used clair as a requirement/dependency it took me quite some time to realize the reason for this.

### Expected Outcome

The helm chart would just expose the option to set resource limits, which can then be set if needed, or at the very least the chart would have limits that do not result in the pod instantly being killed when a few requests come in a short amount of time.

According to the blank helm chart created with `helm create `
> We usually recommend not to specify default resources and to leave this as a conscious
> choice for the user. This also increases chances charts run on environments with little
> resources, such as Minikube. If you do want to specify resources, uncomment the following
> lines, adjust them as necessary, and remove the curly braces after 'resources:'.

### Actual Outcome
The Helm chart has small limits that were the cause of clair in Kubernetes being constantly killed in my deployment.

### Environment

- Clair version/image: 2.0.7
- Kubernetes version (use `kubectl version`): 1.11.7
- Helm version (use `helm version`): 2.12.2
- Network/Firewall setup: 

If the limits can be disabled or removed per default I'd be happy to send a pull request.

PR #756, opened 2019-04-04
[Helm] Do not set resources per default, leave suggested values in comments

naibaf0:
I commented the `resources` values in the helm Chart and added an empty object instead. As described in #730 I stumbled upon this issue, as the default values were very much insufficient.

According to the output of `helm create`:

```
We usually recommend not to specify default resources and to leave this as a conscious
choice for the user. This also increases chances charts run on environments with little
resources, such as Minikube. If you do want to specify resources, uncomment the following
lines, adjust them as necessary, and remove the curly braces after 'resources:'.
```

The setting of resource limits should be a conscious user choice instead of a default.

### DIFF
--- a/contrib/helm/clair/Chart.yaml
+++ b/contrib/helm/clair/Chart.yaml
@@ -1,6 +1,6 @@
 name: clair
 home: https://coreos.com/clair
-version: 0.1.1
+version: 0.1.2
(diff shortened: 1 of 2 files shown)


naibaf0:
ping

PR #766, opened 2019-04-09
Fix: lock updater - return correct bool value

Allda:
The function which extends lock duration in DB returns incorrect bool
value (return default false value). This cause that context is canceled
when extending lock duration and the whole update process fails.

This commit fixes the bug and returns a correct bool value.


Allda:
@jzelinskie This PR fixes root cause of "context canceled" error which I mentioned yesterday on Slack.

jzelinskie (approved):

PR #767, opened 2019-04-16
helm: add mapping for postgres ssl certs

alsokate:
This allows for easy inclusion of an SSL certificate for Clair's connection to its Postgres database, e.g., by storing it locally and referencing it during a helm install/upgrade with the `--set-file` flag.


ldelossa:
@alsokate We will not be supporting helm for production use cases from this repository. We have moved the helm charts to be utilized only for local development. See the following PR: https://github.com/coreos/clair/pull/788

I will be closing this PR for now as local development usually does not warrant SSL cert deployment.

PR #768, opened 2019-04-18
Vulnsrc: return flag map instead of single value 

Allda:
Updates can return only one value using FlagName and FlagValue. This is
limited in some cases when updaters want to store more than one value.
    
This commit enables to return map with more than one key + value.

This PR also cleans unused code in updater.go


ian-panzica:
As far as I can tell this looks good. I see what's going on - but I don't necessarily know enough about the context I guess.

Allda:
Can you reopen this PR please and review it?

PR #770, opened 2019-04-23
KeyValue: allow records with empty value

Allda:
Upcoming change in Red Hat updater requires to store additional
key-value records in database which can contain empty value.

### DIFF
--- a/database/pgsql/keyvalue/keyvalue.go
+++ b/database/pgsql/keyvalue/keyvalue.go
@@ -35,9 +35,9 @@ const (
 )
 
 func UpdateKeyValue(tx *sql.Tx, key, value string) (err error) {
-	if key == "" || value == "" {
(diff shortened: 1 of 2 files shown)


jzelinskie:
Can you describe the "future usage" more?
This does not seem like desirable behavior without further explanation.

Allda:
@jzelinskie You can see the use case in this commit:
https://github.com/Allda/clair/commit/60834d8ec28ac5ff6d269803de9c25b8d0c0a7ed#diff-4728f02070ac77bb9121b0e8b854a996R181

The story behind this PR is that some vulnerability fetchers need to write down CVE should be processed in the next run because in the current run there was not enough metadata to process the CVE. And if there are no such a CVEs in current run the flag needs to be replaced by an empty string.

PR #774, opened 2019-04-26
Updater: fix stuck updater process

Allda:
There are several goroutines in updateWhileRenewingLock() which were
blocking updater process because they were waiting for event which never
happened when update() function was successful.

This commit adds context which unblock other goroutines when update
pass.

### DIFF
--- a/updater.go
+++ b/updater.go
@@ -132,7 +132,7 @@ func RunUpdater(config *UpdaterConfig, datastore database.Datastore, st *stopper
 			}
 
 			if acquiredLock {
-				sleepDuration, err = updateWhileRenewingLock(datastore, whoAmI, isFirstUpdate, st)
+				err = updateWhileRenewingLock(datastore, whoAmI, isFirstUpdate, st)
 				if err != nil {
 					if err == errReceivedStopSignal {
 						log.Debug("updater received stop signal")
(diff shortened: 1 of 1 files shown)


Allda:
@jzelinskie Hi, can you please take a look at this PR? It is blocking Clair from getting new vulnerability data.

updater.go
```diff
@@ -156,9 +158,14 @@ func RunUpdater(config *UpdaterConfig, datastore database.Datastore, st *stopper
 
 var errReceivedStopSignal = errors.New("stopped")
 
-func updateWhileRenewingLock(datastore database.Datastore, whoAmI string, isFirstUpdate bool, st *stopper.Stopper) (sleepDuration time.Duration, err error) {
+func updateWhileRenewingLock(datastore database.Datastore, whoAmI string, isFirstUpdate bool, st *stopper.Stopper) (err error) {
 	g, ctx := errgroup.WithContext(context.Background())
+	// done context is used when updater finishes and all other
+	// go rutines in group should finish too
+	doneCtx := context.Background()
+	doneCtx, done := context.WithCancel(doneCtx)
```
jzelinskie: Can we not wrap the existing context rather than creating a brand new unrelated one?

Allda: @jzelinskie What do you mean by brand new context? I created a new context line above.

jzelinskie: Sorry for the confusion.

I'm suggesting this:
```go
var cancel context.CancelFunc
ctx, cancel = context.WithCancel(ctx) // <-- the same context from errgroup.WithContext()
```

Allda: I am not sure if I understand your suggestion so I rather ask... I need 2 separate contexts to be able to distinguish between when an update is successfully done and when an update fails. You suggest merging those context together if I understand it right. How can we distinguish between those 2 cases then?

jzelinskie:
Hey, this looks reasonable. I actually just haven't approved it because I didn't like the complexity building up in this part of the code, but I don't think that's a reason to block this. A follow up PR can clean things up, if it is actually needed.

jzelinskie (approved):

PR #776, opened 2019-05-03
add disco to namespace mappings for ubuntu

flumm:
otherwise those vulnerabilities do not reach the database


jzelinskie (approved):
Thanks for the backport!

PR #781, opened 2019-05-17
snap package.

mikeroyal:
Includes snapcraft.yaml to build snap.


jzelinskie:
I appreciate the work and feel free to maintain this yourself, but I'm going to close this.

The maintainers have decided that we're sticking to only maintaining a development environment in the upstream project and that downstream consumers of Clair should maintain packaging themselves.

PR #787, opened 2019-05-31
Helm postgres update

ldelossa:
Introduces latest postgres container and confirms helm chart working correctly.
Introduces init container which waits on the postgres instance to be available.

### DIFF
--- a/contrib/helm/clair/values.yaml
+++ b/contrib/helm/clair/values.yaml
@@ -60,20 +60,25 @@ tolerations: []
 
 # Configuration values for the postgresql dependency.
 # ref: https://github.com/kubernetes/charts/blob/master/stable/postgresql/README.md
+# chartVersion: 5.1.2
+# if enabling a feature refer to the above link for full set of configuration info
 postgresql:
-# The dependant Postgres chart can be disabled, to connect to
-# an existing database by defining config.postgresURI
-  enabled: true
+  postgresqlUsername: clair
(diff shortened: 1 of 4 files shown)


contrib/helm/clair/values.yaml
```diff
@@ -60,20 +60,25 @@ tolerations: []
...
-  memory: 1Gi
-# These values are hardcoded until Helm supports secrets.
-# For more info see: https://github.com/kubernetes/helm/issues/2196
-  postgresUser: clair
-  postgresPassword: clair
-  postgresDatabase: clair
+  resources:
+    requests:
+      memory: 256Mi
+      cpu: 250m
```
jzelinskie, line 76: I'm actually wary of adding in resource requests as this can totally change depending on the environment. For example, these values might be great for dev, but not for production.

We need to decide whether or not the Helm chart is for development or production uses.

ldelossa: @jzelinskie these values are simply echo'd from the dependent postgresql chart: 
https://github.com/helm/charts/blob/master/stable/postgresql/values.yaml#L287

Figured hoisting them into our values.yaml made it explicit that the decision is on the user of our chart to set this correctly.

ldelossa:
Going to close this PR in favor of moving this helm chart work to a `/local-dev` directory, making it geared toward local development purposes.

PR #797, opened 2019-06-12
drone: init

jzelinskie:


Makefile
```diff
@@ -1,16 +1,47 @@
...
+.PHONY: build
+build:
+	go build -v -ldflags "-X github.com/coreos/clair/pkg/version.Version=$(COMMIT)" ./cmd/clair
+
+.PHONY: unit-test
+unit-test:
+	go test $(UNIT_TEST_PACKAGES)
+
+.PHONY: db-test
+db-test:
```
ldelossa, line 31: @jzelinskie 

Are these integration tests which require a DB to be available ?

Makefile
```diff
@@ -1,16 +1,47 @@
...
+	go test $(DB_TEST_PACKAGES)
+
 .PHONY: deploy-local
 deploy-local:
+	# This make target is designed to be ran idempotently.
```
ldelossa, line 38: :+1:

Makefile
jzelinskie: Yup! Should I make a doc comment clarifying as much?

ldelossa: @jzelinskie yup! comments would be good. One other thing on my mind is *techincally* we can get these to work with the local dev setup, tho I'm assuming you added this for CI and will call it in the Drone container it's going to run in? 

I do not feel strongly either way, just food for thought since we can launch postgres with local dev and make the postgres db available locally.

ldelossa (approved):

PR #805, opened 2019-06-20
remove unncessary map copy

ldelossa:
Believe the extra map copy is not necessary. LMK if this is not the case.


KeyboardNerd (approved):
Thanks for clean up!

PR #837, opened 2019-08-27
Source binary mapping

ldelossa:
@KeyboardNerd @jzelinskie 

This is the start of closing the source->binary mapping task. I have successfully mapped the source packages to their binary packages with this code, however I seem to be hitting a brick wall when saving layer features. 

To test you can add this table to your ClairV3 database:
```
CREATE TABLE source_binary_mapping (
	id SERIAL PRIMARY KEY,
	source_feature int REFERENCES feature(id),
	binary_feature int REFERENCES feature(id),
	unique(source_feature, binary_feature)
);
```

However we begin to face this issue: `{"Event":"failed to store layer change","Level":"error","Location":"analyzer.go:106","Time":"2019-08-26 20:19:49.353730","error":"associated immutable entities are missing in the database","layer.Hash":"f59d6d019dd5b8398eb8d794e3fafe31f9411cc99a71dabfa587bf732b4a7385"}` 

I troubleshot for awhile but every time I make small change locally, I break something in another database file. Any ideas why we get this issue? I'll continue to look tomorrow.

### DIFF
--- a/database/feature.go
+++ b/database/feature.go
@@ -25,6 +25,8 @@ type Feature struct {
 	Version       string      `json:"version"`
 	VersionFormat string      `json:"versionFormat"`
 	Type          FeatureType `json:"type"`
+	// if the feature has a source affiliated with it bind it here
+	Source *Feature `json:"source"`
 }
 
 // NamespacedFeature is a feature with determined namespace and can be affected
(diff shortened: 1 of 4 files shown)


KeyboardNerd:
yeah it could be really frustrating to see that, we should probably simplify how the data insertion works.
Basically, think a layer as a tree, a layer points to a set of features, namespaces, etc., you need to first insert all the dependencies first before inserting a layer.

ldelossa:
Closing this issue. This is addressed in ClairCore and ClairV4

PR #839, opened 2019-09-04
nvd: fix conversion of r.StatusCode to string

noahklein-zz:
string(404) == "Ɣ"
fmt.Sprintf("%d", 404) == "404"

See https://play.golang.org/p/_avQPvVyLJ4


jzelinskie (approved):
LGTM! Nice catch!
Great first contribution!

Issue #804, opened 2019-06-19
Please use OVAL export for Ubuntu vulnerabilities source


cjwatson:
### Description of Problem / Feature Request

Hi.  I'm one of the engineers responsible for running git.launchpad.net.

At the moment, the Ubuntu vulnerabilities data source is implemented by cloning/pulling https://git.launchpad.net/ubuntu-cve-tracker.  Unfortunately this is pretty expensive for us to service: it's a large repository, and if it's imperfectly packed then cloning it causes `git pack-objects` to use a substantial amount of memory trying to serve a suitable pack to clients.  If there are more than a trivial number of parallel clones at once, then this effectively DDoSes our git service (as happened today).  Obviously we'd like to improve our service's scalability, and we do have work in progress for that, but this won't happen overnight and in any case it can't scale indefinitely.

It's hard for me to tell from our logs how many of the clients are related to Clair; all I can see is a large number of git processes from widely varying IP addresses trying to clone the same large repository and resulting in a large spike in our memory use.  It does look like a likely candidate, though.

Could you please consider using the OVAL exports that our security team provide instead of cloning this git repository?  I'm not in the Ubuntu security team, but I'm told they're [here](https://people.canonical.com/~ubuntu-security/oval/).  The Oracle, RHEL, and SUSE vulnerability sources all use OVAL, so I would hope this is just a matter of plugging different URLs into much the same framework.

Let me know if there are any problems with the data that make this difficult, and I can pass that on to the Ubuntu security team.

### Expected Outcome

Clair uses OVAL exports for Ubuntu, resulting in (I hope) the same effective vulnerabilities feed but with a lighter footprint on our infrastructure.

### Actual Outcome

Clair sends many clients to clone the same large repository from git.launchpad.net, which creaks under the load.

### Environment

N/A

alexmurray:
Another easier short-term option would be mirroring https://git.launchpad.net/ubuntu-cve-tracker to github and updating clair to use that mirror instead of launchpad?

jzelinskie:
I didn't realize that Ubuntu produces OVAL reports, too!

@cjwatson Can you get someone from the Ubuntu security team to chime in? It would be a shame to make the effort to port the code to use this service only to find out that the Clair clients simply overload a different service ran by Canonical and the owner of people.canonical.com creates a new issue that their website is overloaded.

cjwatson:
[...]

PR #853, opened 2019-10-08
Port ubuntu vulnsrc to use OVAL rather than the Ubuntu CVE Tracker

alexmurray:
In response to issue #804


ext/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -13,369 +13,406 @@
...
+	return resp, nil
 }
 
-func (u *updater) Clean() {
-	if u.repositoryLocalPath != "" {
-		os.RemoveAll(u.repositoryLocalPath)
+// Get the latest modification time of a remote file
+// expressed as unix time
+func getLatestModifiedTime(url string) (int64, error) {
+	resp, err := http.Head(url)
```
ldelossa, line 204: @alexmurray  - believe you still want to defer close of the response body here. https://golang.org/pkg/net/http/ the docs groups the "head" method with the advice to close the body.

ext/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -13,369 +13,406 @@
...
-		"Patches": {},
-		// Product
-		"product": {},
-	}
+	ignoredCriterions          []string
+	ubuntuPackageCommentRegexp = regexp.MustCompile(`^(.*) package in ([a-z]+) (?:(?:was vulnerable|is related to the CVE in some way) but has been fixed \(note: '(.*)'\)|is affected and needs fixing).$`)
+	ubuntuOvalFileRegexp       = regexp.MustCompile(`com.ubuntu.([a-z]+).cve.oval.xml.bz2`)
+)
+
+type oval struct {
```
ldelossa, line 61: I don't have an issue with you defining the oval structure here, but you maybe interested in using our Oval parser fork: https://github.com/quay/goval-parser

ldelossa (changes requested):
Looks good. one critique, one suggestion. 
May I assume you tested vulnerability matching in a test env of your own?

ext/vulnsrc/ubuntu/ubuntu.go
alexmurray: I just used the example set by the existing OVAL vulnsrc's.

alexmurray:
>  May I assume you tested vulnerability matching in a test env of your own?

Sadly I have not had any success at actually getting clair running properly locally - is there a simple quick-start-guide or similar to be able to deploy it on a local machine and scan a particular container to test?

ext/vulnsrc/ubuntu/ubuntu.go
alexmurray: This was copied from the same function in the suse vulnsrc - if this is the case that should also be fixed too.

ldelossa: @alexmurray sounds like the right thing to do :+1:

alexmurray: Would  you prefer a separate PR for this or should I bundle it with this one?

ldelossa: Separate pr sounds good to me.

ldelossa:
> > May I assume you tested vulnerability matching in a test env of your own?
> 
> Sadly I have not had any success at actually getting clair running properly locally - is there a simple quick-start-guide or similar to be able to deploy it on a local machine and scan a particular container to test?

@alexmurray this is fine. I will just need a bit to test out that everything works accordingly. 
I will also review the local development docs here: https://github.com/coreos/clair/blob/master/Documentation/local-development.md

alexmurray:
Those docs assume contributors are highly familiar with kubernetes etc and already have this deployed - I am not familiar with k8s myself so perhaps a simple example using something like microk8s might be beneficial?

alexmurray:
I have rebased this PR on top of current master to fix up the conflicts from the previous version - what can I do to help get this merged?

cjwatson:
@ldelossa Please could you indicate whether there's anything else for @alexmurray to do here to satisfy your review comments?  While we've made some further changes to git.launchpad.net to make it easier to control how much this swamps our infrastructure, I'd really like to see this move along so that it stops being a problem.

ldelossa:
@alexmurray @cjwatson my apologies for the delay. We have been putting all our engineering efforts into:
https://github.com/quay/claircore
and https://github.com/quay/clair/tree/development-4.0

I will merge this.

ldelossa (approved):

PR #864, opened 2019-10-24
WIP: Layer content sharing

Allda:
This PR implements new type of Clair's plugins. 

The post-processor plugin can process layers after features and
namespaces have been gathered from layers. This will allow sharing
information between layers in case a layers are missing any information
like namespace.
    
This PR also adds first plugin which is focused on Red Hat's rpms.
Red Hat's plugin share CPE information between layer in case cpes are
not fully defined in some layers.

MR contains non-public data sources so it should not be merged before Red Hat provides new public API.


Allda:
@ldelossa Here is the code change we have been talking about. Please take a look and share your feedback. Thanks!

PR #867, opened 2019-10-28
database: add mapping for Ubuntu Eoan (19.10)

andrewsharon:
This is required for Clair to support Ubuntu Eoan (19.10)


jzelinskie (approved):
LGTM thanks!

PR #883, opened 2019-11-25
Allow limiting pgsql database connections in configuration

mirskifa:
This PR adds the ability to limit the maximum number of connections that the `database/sql` library will make to a pgsql database.

If the configuration is not specified, the pre-existing behaviour is retained (i.e. no max connection limit is set).

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -29,7 +29,11 @@ clair:
       # 32-bit URL-safe base64 key used to encrypt pagination tokens
       # If one is not provided, it will be generated.
       # Multiple clair instances in the same cluster need the same value.
-      paginationkey: 
+      paginationkey:
+
+      # Maximum number of open connections allowed to database
+      # If unspecified or <= 0 then no limit is enforced in Clair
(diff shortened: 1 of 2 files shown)


ldelossa:
@mirskifa looks good to me. Did you come across this need directly?

ldelossa (dismissed):

mirskifa:
Yes, we've been performance testing Clair and found that it was very easy to overload the service when a small amount of queuing (via database/sql blocking calls until a connection is free) allowed for a much more stable flow of responses from a group of Clair instances.

ldelossa:
@mirskifa 

Would you mind adding the database configuration option to the example yaml? Just to advertise this option is now available via the config.

mirskifa:
@ldelossa I've updated the sample.  In doing so my editor has trimmed a lot of trailing whitespace; I'm assuming that's not an issue but let me know.

ldelossa:
@mirskifa - thanks. looks good.

ldelossa (approved):

PR #884, opened 2019-11-25
Add ability to configure read-only pgsql database

mirskifa:
Postgres has the ability to configure read-only streaming replicas but by the current Clair startup attempts a number of operations that do not work on these read-only replicas.

This PR adds support for configuring a read-only database and skips the `RegisterConfiguredDetectors` and `migrateDatabase` calls that would block startup.

### DIFF
--- a/cmd/clair/main.go
+++ b/cmd/clair/main.go
@@ -131,7 +131,9 @@ func Boot(config *Config) {
 
 	defer db.Close()
 
-	clair.RegisterConfiguredDetectors(db)
+	if ro, ok := db.(database.ReadOnly); !ok || !ro.ReadOnly() {
+		clair.RegisterConfiguredDetectors(db)
+	}
 
 	// Start notifier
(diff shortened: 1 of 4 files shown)


ldelossa:
@mirskifa - will this "just work" for individuals who aren't utilizing this feature and have already configured and deployed ClairV3?

mirskifa:
The intent was certainly to make this as backwards compatible as possible:
* If the backend doesn't have the concept of readonly, it shouldn't implement the interface (and given https://github.com/quay/clair/pull/884/files#diff-f3f12965291788a213e857a9f535f219R134 
you would run the migration as normal)
* If you're using pgsql and don't specify the flag, the unmarshal would default to `false` for `ReadOnly` and again you run the migration as normal

ldelossa:
Cool, looks OK to me as long as its opt-in

ldelossa (approved):

PR #888, opened 2019-12-06
Image post-processor plugin

Allda:
The new type of plugins can post-process image layers after an image is scanned. Registered plugins can share information between layers.
After all post-processors process image, the content of an the image is stored in a database.

Currently there is no plugin turned on, but once Red Hat releases public plugins for gathering features and security data I am going to add new post-porcessor plugin.

### DIFF
--- a/api/v3/rpc.go
+++ b/api/v3/rpc.go
@@ -26,6 +26,7 @@ import (
 	pb "github.com/quay/clair/v3/api/v3/clairpb"
 	"github.com/quay/clair/v3/database"
 	"github.com/quay/clair/v3/ext/imagefmt"
+	"github.com/quay/clair/v3/ext/imgpostprocessor"
 	"github.com/quay/clair/v3/pkg/pagination"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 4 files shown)


Allda:
@ldelossa Here is the PR which we have been talking about last week. This PR can be merged and once we release publicly available data I am going to submit new PR with specific post-processor plugin.

You can see the example of a plugin here: 
https://github.com/quay/clair/pull/864/files#diff-e689e8e675327982224ae6633a2dc773

api/v3/rpc.go
```diff
@@ -124,9 +126,28 @@ func (s *AncestryServer) PostAncestry(ctx context.Context, req *pb.PostAncestryR
 	if err = g.Wait(); err != nil {
 		return nil, newRPCErrorWithClairError(codes.Internal, err)
 	}
+	var scannedLayers []*database.LayerScanResult
+	for _, layerRequest := range req.Layers {
+		scannedLayer := layerMap[layerRequest.Hash]
+		scannedLayers = append(scannedLayers, scannedLayer)
+	}
+	scannedLayers, err = imgpostprocessor.PostProcessImage(scannedLayers)
```
ldelossa, line 134: @Allda 
Hey,
I see you're passing an array of layers so that PostProcessImage implementors can work with all of them at once. 
What is the use case here? 
What will your planned implementation do with the array of layers?

Thanks

ldelossa (commented):
@Allda 

This fundamentally looks good. Trying to get a bit more idea on the planned implementation of a PostProcessor.

api/v3/rpc.go
Allda: @ldelossa I am passing all layers to post-processor because I need to share some information (potential namespaces) in between layers. You can see an example how it is done in this PR:

https://github.com/quay/clair/pull/864/files#diff-e689e8e675327982224ae6633a2dc773

ldelossa (approved):
Looks good to me.

PR #891, opened 2019-12-11
Add new plugin for fetching modular rpms

Allda:
Modular rpm is a special type of RPMs which requires different handler.
Modular rpms have been introduced in RHEL8 and Fedora 28.

Modular rpms are stored in rpm database, but to be able to tell if given
modular rpm is vulnerable we need to know which module contains given
rpm. Module name and streams are used as an rpm namespace.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/installing_managing_and_removing_user_space_components/installing-rhel-8-content_using-appstream
https://fedoramagazine.org/modularity-fedora-28-server-edition/


hdonnay (approved):
SGTM

Allda:
@hdonnay Thanks. Can you please also merge it?

PR #892, opened 2019-12-12
indexer: add State method

hdonnay:
This commit also adds openapi documentation for it.


hdonnay:
This need some additional commits, my open finger got itchy.

hdonnay:
OK, ready

indexer/httpclient.go
```diff
@@ -47,19 +47,19 @@ func (s *httpClient) Index(ctx context.Context, manifest *claircore.Manifest) (*
 		return nil, &clairerror.ErrBadManifest{err}
 	}
 
-	url := url.URL{
-		Scheme: s.addr.Scheme,
-		Host:   s.addr.Hostname(),
-		Path:   IndexAPIPath,
+	u, err := s.addr.Parse(IndexAPIPath)
```
ldelossa, line 50: I don't see anything in GoDoc about the Parse method adding a Path field to the existing URL. This is how it works here?

ldelossa (approved):

indexer/httpclient.go
hdonnay: Yeah, it parses it in the context of the *net.URL object: [link](https://pkg.go.dev/net/url?tab=doc#URL.Parse). The previous code would strip port numbers, for example.

ldelossa:
Nice! Didnt see that before. Will make use of that.

PR #902, opened 2020-01-08
indexer: add Accept-Encoding awareness to endpoints returning json

hdonnay:
This commit has the HTTP handlers examine the incoming "accept-encoding"
header, and attempt to compress the payload accordingly.

### DIFF
--- a/cmd/clair/httptransport.go
+++ b/cmd/clair/httptransport.go
@@ -6,11 +6,12 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/quay/claircore/libindex"
+	"github.com/quay/claircore/libvuln"
+
 	"github.com/quay/clair/v4/config"
 	"github.com/quay/clair/v4/indexer"
 	"github.com/quay/clair/v4/matcher"
-	"github.com/quay/claircore/libindex"
-	"github.com/quay/claircore/libvuln"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


indexer/httptransport.go
```diff
@@ -116,8 +212,10 @@ func (h *HTTP) IndexHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	out := enc(w, r)
```
ldelossa: @hdonnay 
Instead of calling this directly in a handler, can it become a middleware chain ?
If it becomes a middleware chain then we can also easily toggle Accept-Encoding handling via a flag and this struct register's method.

indexer/httptransport.go
```diff
@@ -25,6 +32,93 @@ type HTTP struct {
...
+		},
+	}
+)
+
+// Enc picks a suitable content encoding, sets the correct header, and returns
+// an io.WriteCloser configured to write the chosen format to the
+// http.ResponseWriter.
+//
+// The caller must call Close to ensure all data is flushed.
+func enc(w http.ResponseWriter, r *http.Request) io.WriteCloser {
```
ldelossa: Nit:
Move this stuff to it's own file.

cmd/clair/httptransport.go
```diff
@@ -60,7 +61,7 @@ func devMode(ctx context.Context, conf config.Config) (*http.Server, error) {
 	matcher.Register(mux)
 	return &http.Server{
 		Addr:    conf.HTTPListenAddr,
-		Handler: mux,
+		Handler: Compress(mux),
```
ldelossa, line 64: @hdonnay 

Looks good. One question, do we want this to be configurable ? I.e. do not chain the middleware if a flag is not present ?

hdonnay: I don't think so, my thinking that it's client-side opt-in already.

ldelossa (approved):

PR #907, opened 2020-01-16
Remove misleading sentence from README

MVrachev:
In the README it says that in order to read the rest of Clair's documentation 
we should go to the CoreOs website.
On the CoreOs website there is a message which says:
"These docs are deprecated while they are being migrated to Red Hat. 
For the most up to date docs, please see the corresponding GitHub repository."

So, it doesn't make sense to point the people to the CoreOs website.
Also, it doesn't make sense to point them to the documentation folder on GitHub when
every documentation page has its own link on the README.

Signed-off-by: Martin Vrachev <mvrachev@vmware.com>


ldelossa (approved):
Thanks.

PR #908, opened 2020-01-16
Remove "Up and running" link duplication in README

MVrachev:
There is the " Follow instructions to get Clair [up and running]" link on the README
and the "Get up and running with [Local Development]" link on the README.
The second is a subset of the first and it doesn't make sense to include it on
the README.

Signed-off-by: Martin Vrachev <mvrachev@vmware.com>


ldelossa (approved):
Thanks, again.

PR #909, opened 2020-01-21
Support CPE namespaces for rpm packages

Allda:
This PR brings support of rpm packages with CPE namespaces. CPE namespaces come from public Red Hat's container API which was recently introduced. This change supports all Red Hat's products and it doesn't influence classic rpms which are installed outside of Red Hat's containers.


ldelossa (approved):
LGTM

PR #911, opened 2020-01-23
Metrics

hdonnay:
This PR adds in the beginnings of opentelemetry integration.

Additional work needs to be done in the config package to support different exporters.


indexer/httptransport.go
```diff
@@ -84,7 +102,14 @@ func (h *HTTP) IndexReportHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+const (
+	linkIndex  = `<%s>; rel="https://projectquay.io/clair/v1/index_report"`
```
ldelossa: hardcoded urls? Clair will be used outside of quay.

ldelossa (approved):

indexer/httptransport.go
ldelossa, line 139: hardcoded urls? Clair will be used outside of quay.

ldelossa (changes requested):

indexer/httptransport.go
hdonnay: The URI describes the relation, and per the RFC must be an absolute URI.
The link target (bit inside the angles) is completely relative.

So, in effect this is just denoting what is found at the link target.

cmd/clair/introspection.go
```diff
@@ -0,0 +1,184 @@
...
+				lev = lev.Str("username", *u)
+				opt = append(opt, jaeger.WithUsername(*u))
+			}
+			if p != nil {
+				lev = lev.Str("password", strings.Repeat("*", len(*p)))
+				opt = append(opt, jaeger.WithPassword(*p))
+			}
+			e = jaeger.WithCollectorEndpoint(jcfg.Collector.Endpoint, opt...)
+		default:
+			lev.Msg("neither jaeger collector nor agent specified")
```
jpkrohling, line 129: Typically, Jaeger Clients default to `localhost:6831` (UDP) when no explicit option is set. You probably want something like this (although, you'll probably want to store the default value somewhere else):

```go
switch {
case jcfg.Collector.Endpoint != "":
	var opt []jaeger.CollectorEndpointOption
	u, p := jcfg.Collector.Username, jcfg.Collector.Password
	if u != nil {
		lev = lev.Str("username", *u)
		opt = append(opt, jaeger.WithUsername(*u))
	}
	if p != nil {
		lev = lev.Str("password", strings.Repeat("*", len(*p)))
		opt = append(opt, jaeger.WithPassword(*p))
	}
	e = jaeger.WithCollectorEndpoint(jcfg.Collector.Endpoint, opt...)
default:
	endpoint := jcfg.Agent.Endpoint
	if len(endpoint) == 0 {
		endpoint = "localhost:6831"
	}
	lev = lev.Str("agent", jcfg.Agent.Endpoint)
	e = jaeger.WithAgentEndpoint(jcfg.Agent.Endpoint)
}
```

cmd/clair/introspection.go
```diff
@@ -0,0 +1,184 @@
...
+		opts := []jaeger.Option{
+			jaeger.WithOnError(func(err error) {
+				jaegerlog.Error().Err(err).Msg("jaeger error")
+			}),
+		}
+		if jcfg.BufferMax != 0 {
+			lev = lev.Int("buffer_max", jcfg.BufferMax)
+			opts = append(opts, jaeger.WithBufferMaxCount(jcfg.BufferMax))
+		}
+		if jcfg.ServiceName != "" || len(jcfg.Tags) != 0 {
```
jpkrohling: The service name is pretty much the only thing that is typically required when using Jaeger. As such, it would probably be a good idea to have a default service name in case it's not explicitly set.

matcher/httptransport.go
```diff
@@ -104,5 +105,6 @@ func (h *HTTP) VulnerabilityReportHandler(w http.ResponseWriter, r *http.Request
 }
 
 func (h *HTTP) Register(mux *http.ServeMux) {
-	mux.HandleFunc(VulnerabilityReportAPIPath, h.VulnerabilityReportHandler)
+	mux.Handle(VulnerabilityReportAPIPath, othttp.WithRouteTag(VulnerabilityReportAPIPath+":manifest",
```
jpkrohling: I would personally set an explicit tracer here, like:

```go
mux.Handle(VulnerabilityReportAPIPath, 
  othttp.WithRouteTag(VulnerabilityReportAPIPath+":manifest", http.HandlerFunc(h.VulnerabilityReportHandler)), 
  othttp.WithTracer(global.TraceProvider().Tracer("clair")),
)
```

jpkrohling (commented):
I had a couple of comments on the usage of Jaeger, but looks good in general. Do you have a sample trace? I suspect you'll have only traces with a single span, as you seem to be creating spans only for the http layer (which is already a great start).

ldelossa (approved):

PR #912, opened 2020-01-24
Improve the documentation on running Clair

hrist0stoichev:
This PR contains:
- Modification of the documentation on running Clair for non-linux users.
    - Since Docker containers run on a virtual machine and not on the host machine itself for Windows and Mac users, it is especially hard to connect to Clair from the host network. In other words, `curl localhost:6060/...` does not work on Windows and Mac. More information can be found [here](https://docs.docker.com/network/host/) - "The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows".
- Adding docker-compose.yaml.sample and explaining how it can be used to start Clair in the documentation.
    - Related to [this](https://github.com/quay/clair/issues/815) issue.

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -20,7 +20,7 @@ clair:
     options:
       # PostgreSQL Connection string
       # https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
-      source: host=localhost port=5432 user=postgres sslmode=disable statement_timeout=60000
+      source: host=clairdb port=5432 user=postgres sslmode=disable statement_timeout=60000
 
       # Number of elements kept in the cache
       # Values unlikely to change (e.g. namespaces) are cached in order to save prevent needless roundtrips to the database.
--- a/docker-compose.yaml.sample
+++ b/docker-compose.yaml.sample
@@ -1,0 +1,26 @@
+version: '3.7'
+
+services:
+  clair:
+    image: quay.io/coreos/clair:latest
+    command: -config=/config/config.yaml
+    ports:
+      - "6060:6060"
+      - "6061:6061"
+    depends_on:
+      - clairdb
+    volumes:
+      - type: bind
+        source: $PWD/clair_config
+        target: /config
+    networks:
+      - clairnet
+    restart: on-failure
+  clairdb:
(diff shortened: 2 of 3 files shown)


docker-compose.yaml.sample
```diff
@@ -0,0 +1,26 @@
+version: '3.7'
+
+services:
+  clair:
+    image: quay.io/coreos/clair:latest
```
ldelossa, line 5: The issue i see here is by pulling :latest the user will not always run the Clair version they may think.

hrist0stoichev: I totally agree. My reason for using `:latest` is to be consistent with the `Docker` part of the Documentation where `:latest` is already used. Do you think it would be better to change to `:v2.1.2` or even `:v3.0.0-pre1`?

ThunderSon: In general using `latest` is not the best thing, unless it's for development purposes. This is a sample file as well, comments can be added to explain prod vs dev environments?
Removing the latest and setting a version would put some more works on the documentations maintainers, unless some placeholder is put? `<version needed>`

ldelossa:
@hrist0stoichev @ThunderSon 

I think the question here is whether this docker-compose file wants to be used for "local-development" or a docker-compose example for production use cases. 

If it's for production use cases then the documentation can reference the build associated with the commit.

If it's for local-dev, docker-compose should build the containers on the local system and run the software locally.

config.yaml.sample
```diff
@@ -20,7 +20,7 @@ clair:
     options:
       # PostgreSQL Connection string
       # https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
-      source: host=localhost port=5432 user=postgres sslmode=disable statement_timeout=60000
+      source: host=clairdb port=5432 user=postgres sslmode=disable statement_timeout=60000
```
ianfixes, line 23: Changing this to `host=host.docker.internal` fixed this for me on OSX without using docker-compose to set a hostname.

https://docs.docker.com/docker-for-mac/networking/

ldelossa (approved):

PR #913, opened 2020-01-31
Use Handler from promhttp

### LINKED ISSUE #880: undefined : prometheus.Handler
Hi guys,

From the error I'm having and what I've found around, it seems like prometheus.Handler became promhttp.Handler. For now, I can't even do a "go get github.com/quay/clair/cmd/clair" without having the error you can see in the title.

So if someone can do something or have an idea so I could fix the issue I'm having, it would be appreciated

alexandrasandulescu:
Fixes #880

### DIFF
--- a/api/v3/server.go
+++ b/api/v3/server.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


ldelossa:
Thanks.

ldelossa (approved):

PR #918, opened 2020-02-18
Update alpine base image to 3.11

Allda:
The new version contains newer version of rpm package which support
modularity.


ldelossa (approved):
LGTM.

PR #922, opened 2020-02-24
clair: add authorization checking

hdonnay:
This commit adds authorization via two methods: PSK and a Quay
keyserver.

The PSK simply uses a key specified in the configuration file.

The keyserver authorization checking relies on the extra "kid" JWT
header to look up a public key and validate the claims on the request.

The tests for the keyserver integration are opportunistic and require
passing a "keyserver" flag to the cmd/clair test for initial setup. Once
initial setup is done, the key and server information will be cached and
used opportunistically on future test runs.

### DIFF
--- a/cmd/clair/httpauth.go
+++ b/cmd/clair/httpauth.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"context"
+	"net/http"
+	"strings"
+)
+
+// AuthCheck is an interface that reports whether the passed request should be
+// allowed to continue.
+type AuthCheck interface {
+	Check(context.Context, *http.Request) bool
+}
+
+type authHandler struct {
+	auth AuthCheck
+	next http.Handler
+}
+
+func (h *authHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if !h.auth.Check(r.Context(), r) {
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	h.next.ServeHTTP(w, r)
+}
+
+// AuthHandler returns a Handler that gates access to the passed Handler behind
+// the passed AuthCheck.
+func AuthHandler(h http.Handler, f AuthCheck) http.Handler {
+	return &authHandler{
+		auth: f,
+		next: h,
+	}
+}
+
+func fromHeader(r *http.Request) (string, bool) {
+	hs, ok := r.Header["Authorization"]
+	if !ok {
+		return "", false
+	}
+	for _, h := range hs {
+		if strings.HasPrefix(h, "Bearer ") {
+			return strings.TrimPrefix(h, "Bearer "), true
+		}
+	}
+	return "", false
+}
--- a/cmd/clair/httpauth_keyserver.go
+++ b/cmd/clair/httpauth_keyserver.go
@@ -1,0 +1,135 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"path"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/gregjones/httpcache"
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+type ks struct {
+	root   *url.URL
+	client *http.Client
+	mu     sync.RWMutex
+	cache  map[string]*jose.JSONWebKey
+}
+
+// Check implements AuthCheck.
+func (s *ks) Check(ctx context.Context, r *http.Request) bool {
+	wt, ok := fromHeader(r)
+	if !ok {
+		return false
+	}
+	tok, err := jwt.ParseSigned(wt)
+	if err != nil {
+		return false
+	}
+	aud, err := r.URL.Parse("/")
+	if err != nil {
+		return false
+	}
+	// Need to find the key id.
+	ok = false
+	var kid string
+	for _, h := range tok.Headers {
+		if h.Algorithm == string(jose.RS256) {
+			ok = true
+			kid = h.KeyID
+			break
+		}
+	}
+	if !ok {
+		return false
+	}
+	// Need to pull out the issuer to fetch the key. We cannot return "true"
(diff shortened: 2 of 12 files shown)


cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+	if !h.auth.Check(r.Context(), r) {
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	h.next.ServeHTTP(w, r)
+}
+
+// Authorize returns a Handler that gates access to the passed Handler behind
+// the passed AuthCheck.
+func Authorize(h http.Handler, f AuthCheck) http.Handler {
```
ldelossa: I think naming this "AuthHandler" would convey it's a constructor for the authHandler middleware.

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+		if strings.HasPrefix(h, "Bearer ") {
+			return strings.TrimPrefix(h, "Bearer "), true
+		}
+	}
+	return "", false
+}
+
+// PSKAuth returns an AuthCheck that validates a JWT with the supplied key and
+// ensures the issuer claim matches.
+func PSKAuth(key []byte, issuer string) (AuthCheck, error) {
```
ldelossa: "PSKAuthCheck"

ldelossa: Do you want to return interface here?

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+	return &psk{
+		key: key,
+		iss: issuer,
+	}, nil
+}
+
+type psk struct {
```
ldelossa: nit: define this above where you use it.

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+func (p *psk) Check(_ context.Context, r *http.Request) bool {
+	wt, ok := fromHeader(r)
+	if !ok {
+		return false
+	}
+	tok, err := jwt.ParseSigned(wt)
+	if err != nil {
+		return false
+	}
+	if tok.Headers[0].Algorithm != string(jose.HS256) {
```
ldelossa: So clair will only handle HS256 signed JWT's ? Maybe a little limiting for other types of clients unless this is the most secure algo JWT spec allows.

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+	if err := cl.ValidateWithLeeway(jwt.Expected{
+		Issuer: p.iss,
+		Time:   time.Now(),
+	}, 15*time.Second); err != nil {
+		return false
+	}
+	return true
+}
+
+type ks struct {
```
ldelossa: nit: breakout the keyserver stuff to it's own file.

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	tok, err := jwt.ParseSigned(wt)
+	if err != nil {
+		return false
+	}
+	aud, err := r.URL.Parse("/")
+	if err != nil {
+		return false
+	}
+	// Need to find a the key id.
```
ldelossa: s/a//g I think.

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+func (s *ks) Check(ctx context.Context, r *http.Request) bool {
+	wt, ok := fromHeader(r)
+	if !ok {
+		return false
+	}
+	tok, err := jwt.ParseSigned(wt)
+	if err != nil {
+		return false
+	}
+	aud, err := r.URL.Parse("/")
```
ldelossa: whats this little check do ?

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+			ok = true
+			kid = h.KeyID
+			break
+		}
+	}
+	if !ok {
+		return false
+	}
+	// Need to pull out the issuer to fetch the key. We cannot return "true"
+	// until *after* a keyed Claims call succeeds.
```
ldelossa: what is "keyed Claims call" ?

cmd/clair/httpauth.go
```diff
@@ -0,0 +1,215 @@
...
+		if _, ok := s.cache[ck]; !ok {
+			s.cache[ck] = jwk
+		}
+		s.mu.Unlock()
+	}
+
+	if err := tok.Claims(jwk.Key, &cl); err != nil {
+		return false
+	}
+	// Can now *possibly* return true.
```
ldelossa: nit: comment is a little confusing, think the reader can infer the error case without *possibly*

ldelossa (approved):
Looks good, all comments are mostly cosmetic or me asking questions. Very cool

cmd/clair/httpauth.go
hdonnay: I was trying to point out, along with the comment on line 132, that there can't be *any* codepath between those two points that returns true.

hdonnay: I was trying to point out, along with the comment on line 132, that there can't be *any* codepath between those two points that returns true.

cmd/clair/httpauth.go
hdonnay: The call under this is unsafe, so we can't trust it until we fetch the key and make the normal Claims call with it.

cmd/clair/httpauth.go
hdonnay: I think this may not be useful as I thought, yeah.

cmd/clair/httpauth.go
hdonnay: gets the root from the request URL.

e.g. `http://server/any/path` -> `http://server/`

cmd/clair/httpauth.go
hdonnay: yeah, I think so. It makes the return values for all these functions the same.

hdonnay:
Got the LGTM out-of-band

PR #937, opened 2020-03-06
clair: add "mode" argument

hdonnay:
This change promotes "mode" out of the config file and makes it settable
via the command line and the environment.

Additional changes are introduced that make sure a single configuration
file can work in multiple "modes" and there's some clean-up changes.


ldelossa (changes requested):
Update sample conf and local dev confs plz!

ldelossa (approved):
niceee

PR #942, opened 2020-03-12
clair: use Etag header to communicate indexer state change

hdonnay:
In the course of this change, the Etag validation logic is also
corrected.


alecmerdler (approved):
Excited to see how this integrates with Quay!

PR #943, opened 2020-03-13
Louis/init

ldelossa:
This PR introduces several new packages.
`initialize` - a package where all ClairV4 initialization occurs. 
`httptransport` - the http transport layer exposing clair's main functionality over http
`introspection` - initialize and registration of metrics and tracing 

Initialization occurs in one place with well defined methods explaining what our initialization process entails. 

The httptransport package provides a strategy of avoiding circular dependencies in a micro-service mono-repo. httptransport imports specific services (indexer, matcher, etc...), services will never import each other (causing a circular dependency), and services will never import httptransport package (another circular dependency) 

Introspection configuration has a lot of moving pieces therefore it became it's own high level package. ~~Introspection can be completely disabled by not providing a "IntrospectionAddr" field in the configuration (no exporters setup, not http latency recording,  othttp handlers no-op when called)~~

After discussion with @hdonnay - we landed on the diagnostic portion of introspection (health check, pproff) always being enabled. A default introspection server will always run at *:8089. Prometheus and Jaeger tracing is disabled if no config for these exporters are provided.


hdonnay:
I disagree with having to opt-in to the introspection server starting, it should use a random port if there's not one specified in the config.

ldelossa:
@hdonnay there are times where your application errors out, and you want to stop bombarding the downstream collector with a known issue. It's important to be able to shut off introspection at those times. 

Also as a general "is my metrics system causing this issue" you'd want to turn introspection off for a sanity check.

hdonnay:
Having a server running in-process is different from emitting traces. The debugging, profiling, health/ready, and pull-metrics endpoints should all be enabled by default.

ldelossa:
@hdonnay some of that I can agree with, mainly the health and debugging. But I do not agree with making a distinction between "push" and "pull" metrics. I don't think our code should "always turn on" prometheus by default. If I am trouble shooting a strange bug I want to be able to turn these exporters OFF.

middleware/introspection/latency.go
```diff
@@ -0,0 +1,48 @@
...
+// handler implements the http.Handler interface
+type handler struct {
+	meter   metric.Meter
+	latency metric.Int64Measure
+	next    http.Handler
+}
+
+func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	labels := h.meter.Labels(pathKey.String(r.URL.Path))
```
hdonnay: We can't use the request path as a label, the cardinality will be too high.

ldelossa:
@hdonnay 

```
diff --git a/introspection/introspection.go b/introspection/introspection.go
index ade2bcbf..e6653408 100644
--- a/introspection/introspection.go
+++ b/introspection/introspection.go
@@ -20,13 +20,14 @@ import (
 )
 
 const (
-	Prom                  = "prometheus"
-	DefaultPromEndpoint   = "/metrics"
-	DogStatsD             = "dogstatsd"
-	STDOut                = "stdout"
-	Jaeger                = "jaeger"
-	DefaultJaegerEndpoint = "localhost:6831"
-	HealthEndpoint        = "/healthz"
+	Prom                     = "prometheus"
+	DefaultPromEndpoint      = "/metrics"
+	DogStatsD                = "dogstatsd"
+	STDOut                   = "stdout"
+	Jaeger                   = "jaeger"
+	DefaultJaegerEndpoint    = "localhost:6831"
+	HealthEndpoint           = "/healthz"
+	DefaultIntrospectionAddr = "localhost:8089"
 )
 
 // Introspection provides an http server
@@ -48,15 +49,18 @@ type Introspection struct {
 func New(ctx context.Context, conf config.Config, health func() bool) (*Introspection, error) {
 	logger := zerolog.Ctx(ctx).With().Str("component", "introspection").Logger()
 
+	var addr string
 	if conf.IntrospectionAddr == "" {
-		logger.Info().Msg("no introspection address provied. introspection will not be enabled")
-		return nil, nil
+		addr = DefaultIntrospectionAddr
+		logger.Info().Str("address", addr).Msg("no introspection address provied. using default addr")
+	} else {
+		addr = conf.IntrospectionAddr
 	}
 
 	i := &Introspection{
 		conf: conf,
 		Server: &http.Server{
-			Addr:        conf.IntrospectionAddr,
+			Addr:        addr,
 			BaseContext: func(_ net.Listener) context.Context { return ctx },
 		},
 		ServeMux: http.NewServeMux(),
@@ -73,8 +77,7 @@ func New(ctx context.Context, conf config.Config, health func() bool) (*Introspe
 	logger.Info().Str("sink", conf.Metrics.Name).Msg("configuring")
 	switch conf.Metrics.Name {
 	case "", "default":
-		logger.Info().Str("sink", "default").Msg("configuring prometheus by default")
-		fallthrough
+		logger.Info().Str("sink", "default").Msg("no metrics sync enabled")
 	case Prom:
 		err := i.withPrometheus()
 		if err != nil {
@@ -118,6 +121,8 @@ func New(ctx context.Context, conf config.Config, health func() bool) (*Introspe
 		if err != nil {
 			return nil, fmt.Errorf("error configuring jaeger tracing: %v", err)
 		}
+	default:
+		logger.Info().Msg("no distributed tracing enabled")
 	}
 
 	// configure diagnostics
```

diff will always use a default IntrospectionAddr with health and diagnostics enabled. Will omit exporters if they are not present in the config specifically.

hdonnay:
I suppose having a default port works, too. It shouldn't be bound to localhost, though.

I guess I can see the logic of wanting to turn the metrics completely off in exceptional cases, and having tracing be none-by-default but metrics none-by-config is weird.

ldelossa:
@hdonnay, agreed. 
Default port bound to "*" then ?

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+
+	i := &Introspection{
+		conf: conf,
+		Server: &http.Server{
+			Addr:        conf.IntrospectionAddr,
+			BaseContext: func(_ net.Listener) context.Context { return ctx },
+		},
+		ServeMux: http.NewServeMux(),
+	}
+	i.logger = logger
```
hdonnay: Don't capture the logger, it can be passed everywhere here.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		if p != nil {
+			opt = append(opt, jaeger.WithPassword(*p))
+		}
+		e = jaeger.WithCollectorEndpoint(endpoint, opt...)
+	}
+
+	// configure the exporter
+	component := fmt.Sprintf("jaeger-exporter:%v", endpoint)
+	jaegerLog := log.With().Str("component", component).Logger()
```
hdonnay: Don't use the global logger.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+	"go.opentelemetry.io/otel/exporter/trace/jaeger"
+	"go.opentelemetry.io/otel/exporter/trace/stdout"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+const (
+	Prom                  = "prometheus"
+	DefaultPromEndpoint   = "/metrics"
+	DogStatsD             = "dogstatsd"
+	STDOut                = "stdout"
```
hdonnay: Should just be `Stdout`.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+	return nil
+}
+
+// withDogStatsD configures a dogstatsd open telemetry
+// pipeline.
+func (i *Introspection) withDogStatsD() error {
+	if i.conf.Metrics.Dogstatsd.URL == "" {
+		return fmt.Errorf("dogstatsd metrics were specified but no url was configured")
+	}
+	log.Info().
```
hdonnay: Don't use the global logger.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+	endpoint := DefaultPromEndpoint
+	if i.conf.Metrics.Prometheus.Endpoint != nil {
+		endpoint = *i.conf.Metrics.Prometheus.Endpoint
+	}
+
+	i.logger.Info().Str("endpoint", endpoint).
+		Str("server", i.Addr).
+		Msg("configuring prometheus with endpoint")
+
+	promlog := log.With().Str("component", "promtheus-metrics-exporter").Logger()
```
hdonnay: Don't use the global logger.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+			Addr:        conf.IntrospectionAddr,
+			BaseContext: func(_ net.Listener) context.Context { return ctx },
+		},
+		ServeMux: http.NewServeMux(),
+	}
+	i.logger = logger
+
+	// check for health
+	if health == nil {
+		i.logger.Warn().Msg("no health check configured. a default one will be used which simply returns OK")
```
hdonnay: `"no health check configured; unconditionally reporting OK"`

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+// endpoint to i's servemux.
+func (i *Introspection) withPrometheus() error {
+	endpoint := DefaultPromEndpoint
+	if i.conf.Metrics.Prometheus.Endpoint != nil {
+		endpoint = *i.conf.Metrics.Prometheus.Endpoint
+	}
+
+	i.logger.Info().Str("endpoint", endpoint).
+		Str("server", i.Addr).
+		Msg("configuring prometheus with endpoint")
```
hdonnay: `"configuring prometheus"`

"With endpoint" doesn't mean anything, there can't not be an endpoint.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+		},
+		ServeMux: http.NewServeMux(),
+	}
+	i.logger = logger
+
+	// check for health
+	if health == nil {
+		i.logger.Warn().Msg("no health check configured. a default one will be used which simply returns OK")
+		health = func() bool { return true }
+	}
```
hdonnay: `Health` isn't used or assigned, it should just be passed to `withDiagnostics` and the `health` struct member removed.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+	Jaeger                = "jaeger"
+	DefaultJaegerEndpoint = "localhost:6831"
+	HealthEndpoint        = "/healthz"
+)
+
+// Introspection provides an http server
+// exposing Clair metrics and traces
+type Introspection struct {
+	// configuration provided when starting Clair
+	conf config.Config
```
hdonnay: Why is a config copied onto this struct? Nothing uses it once `New` returns.

introspection/introspection.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/quay/clair/v4/config"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/key"
+	"go.opentelemetry.io/otel/exporter/metric/dogstatsd"
+	"go.opentelemetry.io/otel/exporter/metric/prometheus"
+	"go.opentelemetry.io/otel/exporter/trace/jaeger"
+	"go.opentelemetry.io/otel/exporter/trace/stdout"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
```
hdonnay: Mixed imports

initialize/init.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/quay/clair/v4/matcher"
+	"github.com/rs/zerolog"
+)
+
+type Init struct {
+	// configuration provided
+	conf config.Config
+	// a global ctx with an embedded logger
+	// enables downstream logging and global application cancelation
+	GlobalCTX context.Context
```
hdonnay, line 18: There shouldn't be a Context captured here, it should just be passed in. The calling code can construct a CancelFunc on its own.

initialize/init.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/quay/clair/v4/httptransport"
+	"github.com/quay/clair/v4/indexer"
+	"github.com/quay/clair/v4/introspection"
+	"github.com/quay/clair/v4/matcher"
+	"github.com/rs/zerolog"
+)
+
+type Init struct {
+	// configuration provided
+	conf config.Config
```
hdonnay, line 15: This shouldn't be captured, as it needs to get passed to the `httptransport` and `introspection` constructors, anyway.

initialize/logging.go
```diff
@@ -0,0 +1,49 @@
...
+	zerolog.SetGlobalLevel(level)
+
+	// attach global logger to ctx
+	i.GlobalCTX, i.GlobalCancel = context.WithCancel(context.Background())
+	globalLogger := log.With().Timestamp().Logger()
+	i.GlobalCTX = globalLogger.WithContext(i.GlobalCTX)
+
+	globalLogger.Info().Str("component", "init/Init.Logging").Msg("logging initialized")
+	return nil
+}
```
hdonnay, line 27: This shouldn't use the global logger, it should construct a logger and embed it in a passed-in context.

initialize/services.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/quay/claircore/libindex"
+	"github.com/quay/claircore/libvuln"
+	"github.com/rs/zerolog"
+)
+
+// Services will initialize the correct ClairCore services
+// dependent on operation mode.
+//
+// Services maybe local or remote (over a network).
+func (i *Init) Services() error {
```
hdonnay, line 20: This function should just construct all the options and then only use the ones it needs in the switch.

introspection/introspection.go
```diff
@@ -0,0 +1,294 @@
...
+	health func() bool
+}
+
+func New(ctx context.Context, conf config.Config, health func() bool) (*Introspection, error) {
+	logger := zerolog.Ctx(ctx).With().Str("component", "introspection").Logger()
+
+	var addr string
+	if conf.IntrospectionAddr == "" {
+		addr = DefaultIntrospectionAddr
+		logger.Info().Str("address", addr).Msg("no introspection address provied. using default addr")
```
hdonnay: s/addr//

httptransport/indexhandler.go
```diff
@@ -0,0 +1,105 @@
...
+				Code:    "bad-request",
+				Message: "bogus manifest",
+			}
+			je.Error(w, resp, http.StatusBadRequest)
+			return
+		}
+		next := path.Join(IndexReportAPIPath, m.Hash.String())
+
+		w.Header().Add("link", fmt.Sprintf(linkIndex, next))
+		w.Header().Add("link", fmt.Sprintf(linkReport, path.Join(apiRoot, "vulnerabilty_report", m.Hash.String())))
```
hdonnay:
```
w.Header().Add("link", fmt.Sprintf(linkReport, path.Join(VulnerabilityReportPath, m.Hash.String()))) 
```

httptransport/httptransport.go
```diff
@@ -0,0 +1,201 @@
...
+		api, ok := t.conf.Auth.Params[param]
+		if !ok {
+			return fmt.Errorf("missing needed config key: %q", param)
+		}
+		ks, err := auth.NewQuayKeyserver(api)
+		if err != nil {
+			return fmt.Errorf("failed to initialize quay keyserver: %v", err)
+		}
+		t.Server.Handler = auth.Handler(t.Server.Handler, ks)
+	case "pks":
```
hdonnay: `psk`

httptransport/httptransport.go
```diff
@@ -0,0 +1,201 @@
...
+type HttpTransport struct {
+	// HttpTransport embeds a http.Server and http.ServeMux.
+	// The http.Server will be configured with the ServeMux on successful
+	// initialization.
+	conf config.Config
+	*http.Server
+	*http.ServeMux
+	indexer  indexer.Service
+	matcher  matcher.Service
+	traceOpt othttp.Option
```
hdonnay: the Config, indexer, matcher, and traceOpt should just be passed around inside the constructor, as they're not used after the constructor returns.

cmd/clair/values.go
```diff
@@ -52,11 +54,11 @@ func (v *ConfMode) String() string {
...
-		return "combo"
+		return config.ComboMode
 	case ModeIndexer:
-		return "indexer"
+		return config.IndexerMode
 	case ModeMatcher:
-		return "matcher"
+		return config.MatcherMode
 	case ModeNotifier:
 		return "notifier"
```
hdonnay: This should also return a constant if the rest of these are.

cmd/clair/main.go
```diff
@@ -3,19 +3,16 @@ package main
...
 	"os"
 	"os/signal"
-	"strings"
 	"time"
 
 	"github.com/rs/zerolog"
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/quay/clair/v4/config"
+	"github.com/quay/clair/v4/initialize"
```
hdonnay, line 15: "Initialize" is a weird name for this, as it does *all* the application setup except for calling Listen.

httptransport/httptransport.go
```diff
@@ -0,0 +1,201 @@
...
+	apiRoot                 = "/api/v1/"
+	VulnerabilityReportPath = apiRoot + "vulnerability_report/"
+	IndexAPIPath            = apiRoot + "index_report"
+	IndexReportAPIPath      = apiRoot + "index_report/"
+	StateAPIPath            = apiRoot + "state"
+)
+
+// HttpTransport is the primary http server
+// Clair exposes it's functionality on.
+type HttpTransport struct {
```
hdonnay: `Server` would be a better name for this. `httptransport.HttpTransport` vs `httptransport.Server`

cmd/clair/main.go
```diff
@@ -53,129 +50,54 @@ func main() {
...
+	if err != nil {
+		golog.Fatalf("initialized failed: %v", err)
 	}
-	log := zerolog.New(out).With().
-		Timestamp().
-		Logger().
-		Level(ll)
+	logger := zerolog.Ctx(init.GlobalCTX).With().Str("component", "main").Logger()
 
-	// create global application context
```
hdonnay, line 69: The `main` function should still create and control the global context.

hdonnay (changes requested):
In addition to specific comments, most of the imports are mixed. Here's a script to fix them:

```
go list -f '{{$d := .Dir}}{{range .GoFiles}}{{printf "%s/%s\n" $d .}}{{end}}' ./... |
    xargs sed -i'' '/import (/,/)/{ /^$/d }'
go list -f '{{.Dir}}' ./... |
    xargs goimports -local "$(go list -m)" -w
```

cmd/clair/main.go
ldelossa: Initialization is "get your app ready" main calls listen and serve after init confirms all things are ready to go.

cmd/clair/main.go
ldelossa: Why? A single ctx that drives the application globally is created. Theres no ambiguity.

ldelossa: Why? A single ctx that drives the application globally is created. Theres no ambiguity.

httptransport/httptransport.go
ldelossa: Id prefer the shorter method signatures obtained by putting the common bits in the struct.

initialize/logging.go
ldelossa: Main is not initializing the global ctx and logger, the initialization package is

introspection/introspection.go
ldelossa: @hdonnay is it an issue to capture the logger? I do not want to pass a logger to every method, and I do not want every method signature taking a logger, it's verbose.

middleware/introspection/latency.go
ldelossa: fixed this by making introspection middleware wrap a handler, not a mux, and takes the path pattern string as an argument to construction. The sig now looks similar to othttp sig. I also increased cardinality of the tag by adding the request method

```
# HELP clair_http_latency latency of http requests
# TYPE clair_http_latency summary
clair_http_latency_sum{http_method="POST",http_path="/api/v1/index_report"} 8523
clair_http_latency_count{http_method="POST",http_path="/api/v1/index_report"} 1
```

middleware/introspection/latency.go
```diff
@@ -0,0 +1,63 @@
...
+// For example a resulting prometheus latency metric will take the form
+// "clair_http_latence{http_path={path}, http.method={method}}"
+func Handler(next http.Handler, path string) http.Handler {
+	meter := global.MeterProvider().Meter("clair")
+	late := meter.NewInt64Measure("clair.http.latency",
+		metric.WithDescription("latency of http requests"),
+		metric.WithUnit(unit.Milliseconds),
+		metric.WithAbsolute(true),
+		metric.WithKeys(pathKey),
+		metric.WithKeys(methodKey),
```
ldelossa, line 37: @hdonnay this api is whack, why must I put my keys here if I also provide them to h.latency.Record.

introspection/introspection.go
ldelossa: After dealing with "don't use global logger" comments, I see your point here. fixing.

httptransport/httptransport.go
ldelossa: @hdonnay  cool, I'm doing the same for introspection package.
httptransport.Server
introspection.Server

httptransport/httptransport.go
hdonnay: Missed this

httptransport/indexhandler.go
hdonnay: Missed this

middleware/introspection/latency.go
hdonnay: You set the keys here and record the values at Record time. You can use the Bind method to bind the labels early, but can't in this case because they're not known beforehand.

hdonnay (changes requested):

introspection/introspection.go
ldelossa: copied so that code cannot alter Config. 
IMO it's OK to optimize for writing less code here. these are init structures which are instantiated at most 1 time in our app. If we have memory pressure over copying the config twice, we have much larger problems.

hdonnay (approved):

PR #949, opened 2020-03-22
Required Properties for Clair V4 OpenAPI

alecmerdler:
### Description

Adds `required` properties to the OpenAPI schema definitions so that clients can verify responses from the Clair server.

Addresses https://issues.redhat.com/browse/PROJQUAY-494


ldelossa (approved):
:pray:

PR #956, opened 2020-03-27
Clairctl report subcommand

hdonnay:


ldelossa:
@hdonnay would using the http client for the transport layer be applicable here instead of writing your own client in the cli tool ?

https://github.com/quay/clair/blob/development-4.0/httptransport/client/httpclient.go

ldelossa (approved):

hdonnay:
I talked over IM, but to record here, also:

I forgot this client was in there :S An end-user flow is also different from the intra-service flow, so it might be useful to keep around, at least for a while, to make sure that flow is easy.

PR #961, opened 2020-04-01
App-SRE Integration

ldelossa:
This PR will remain in draft until all RedHat App-SRE integrations are successfully working.

### DIFF
--- a/contrib/openshift/build_and_deploy.sh
+++ b/contrib/openshift/build_and_deploy.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+set -exv
+REPOSITORY="quay.io/app-sre"
+IMAGE="${REPOSITORY}/clair"
+
+docker run --user="$(id -u):$(id -g)" -v $(pwd):/go/clair:z -e GOCACHE=/go/clair golang:1.14.1-buster "bash" "-c" "cd /go/clair; ls -la; go mod vendor"
+make container-build
+
+GIT_HASH=`git rev-parse --short=7 HEAD`
+skopeo copy --dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
+    "docker-daemon:clair-local:latest" \
+    "docker://${IMAGE}:latest"
+
+skopeo copy --dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
+    "docker-daemon:clair-local:latest" \
+    "docker://${IMAGE}:${GIT_HASH}"
--- a/contrib/openshift/manifests/manifests.yaml
+++ b/contrib/openshift/manifests/manifests.yaml
@@ -1,0 +1,314 @@
+---
+apiVersion: v1
+kind: Template
+metadata:
+  name: clair
+  labels:
+    app: clair
+objects:
+  #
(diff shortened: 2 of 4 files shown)


contrib/openshift/build_and_deploy.sh
```diff
@@ -0,0 +1,42 @@
...
+    echo "could not set root to clair repository"
+    exit 1
+fi
+
+go mod vendor
+make container-build
+
+skopeo copy --dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
+    "docker-daemon:clair-local:latest" \
+    "docker://clairv4:latest"
```
ldelossa: @tparikh fixed.

tparikh: @ldelossa it needs to be `docker://quay.io/app-sre/clairv4` and not `docker://clairv4:latest` I think.

contrib/openshift/build_and_deploy.sh
```diff
@@ -0,0 +1,42 @@
...
+
+skopeo copy --dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
+    "docker-daemon:clair-local:latest" \
+    "docker://clairv4:${GIT_HASH}"
```
tparikh: I don't see `GIT_HASH` set in the script. You can add GIT_HASH=`git rev-parse --short=7 HEAD` to the script. Note: ` around the command are not visible due to markdown formatting

tparikh (changes requested):

contrib/openshift/build_and_deploy.sh
ldelossa: good catch! thought that was coming from the env that runs the script

tparikh:
LGTM

tparikh (commented):
@ldelossa Are Clair logs shipped to CloudWatch?

ldelossa:
@tparikh not currently. I will be doing a full metrics pass on ClairV4 and ClairCore before we move further into production.

contrib/openshift/manifests/manifests.yaml
```diff
@@ -0,0 +1,314 @@
...
+    value: "4096Mi"
+    displayName: the matcher's memory requests in vCPUs
+  #
+  # shared params
+  #
+  - name: CLAIR_SECRET
+    value: "clair"
+    displayName: the vault secret containing clair's config
+  - name: CLAIR_IMAGE
+    value: ""
```
rrati: Wouldn't you want this to point to a default image?  Same for tag

ldelossa: Let me take a look, i was under the impression these values do not get used.

rrati: These should be default values.  Deploments can override, but it doesn't make sense imo for the image to be blank as a default.

ldelossa: Got it, I'll fix

ldelossa:
@rrati updated the image tags.

hdonnay (approved):

PR #962, opened 2020-04-01
Update diffs Endpoints

hdonnay:
This PR plumbs update diffs through to the clair API, and introduces an "internal" API namespace.

### DIFF
--- a/httptransport/client/differ.go
+++ b/httptransport/client/differ.go
@@ -1,0 +1,176 @@
+package client
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"path"
+	"strings"
+	"sync"
+
+	"github.com/google/uuid"
+	"github.com/quay/claircore/libvuln/driver"
+
+	"github.com/quay/clair/v4/httptransport"
+	"github.com/quay/clair/v4/matcher"
+)
+
+var _ matcher.Differ = (HTTP)(nil)
+
+// DeleteUpdateOperations attempts to delete the referenced update operations.
+func (c HTTP) DeleteUpdateOperations(ctx context.Context, ref ...uuid.UUID) error {
+	u, err := c.addr.Parse(httptransport.UpdatesAPIPath)
+	if err != nil {
+		return err
+	}
+
+	// Spawn a few requests that will write their result into "errs".
+	//
+	// These'll most likely be multiplexed and to the same host, so pick a nice
+	// lowish number like 4.
+	//
+	// Don't use an errgroup because we want to actually issue all the DELETEs,
+	// not stop all requests on the first error.
+	var wg sync.WaitGroup
+	item := make(chan int)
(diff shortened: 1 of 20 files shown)


httptransport/updatediffhandler.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/quay/claircore/pkg/jsonerr"
+
+	"github.com/quay/clair/v4/matcher"
+)
+
+type updateDiffHandler struct {
+	matcher.Differ
+}
+
+func (h *updateDiffHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
ldelossa: When I look at this I am not sure what your handler is doing at first glance. IMO it's much better to have a simple function which directly correlates with what's happening in the handler, not a handler thats doing some muxing itself.

httptransport/client/differ.go
```diff
@@ -0,0 +1,162 @@
...
+func (c HTTP) DeleteUpdateOperations(ctx context.Context, ref ...uuid.UUID) error {
+	var wg sync.WaitGroup
+	errs := make([]error, len(ref))
+	wg.Add(len(ref))
+	u, err := c.addr.Parse(httptransport.UpdatesAPIPath)
+	if err != nil {
+		return err
+	}
+
+	for i, lim := 0, len(ref); i < lim; i++ {
```
ldelossa: Since this is HTTP, and it'll directly use up conns, we probably want either a MAX_REFs which returns an error, or a way to bound that concurrency.

ldelossa (changes requested):

httptransport/client/differ.go
```diff
@@ -0,0 +1,176 @@
...
+	//
+	// These'll most likely be multiplexed and to the same host, so pick a nice
+	// lowish number like 4.
+	//
+	// Don't use an errgroup because we want to actually issue all the DELETEs,
+	// not stop all requests on the first error.
+	var wg sync.WaitGroup
+	item := make(chan int)
+	errs := make([]error, len(ref))
+	for i := 0; i < 4; i++ {
```
ldelossa, line 39: I think I'm a little confused about how this method works. You limit the amount of deletes that can be performed by 4, but do not return any indication to the client if they send more then four refs into the method, that any refs >= 5 will not be handled.

if that is the case, I think we either want to do one of two things:
- place the wait operations inside a for loop, doing "batches" of deletes with a safe concurrency bound
- return an error to the client to attempt delete again with less refs. 

lmk if I'm off here.

ldelossa (changes requested):
Everything looks good, I am just confused about usage of delete method.

httptransport/client/differ.go
hdonnay: This for loop spawns for 4 concurrent workers that pull off the `item` channel, then the for loop below sends all the indexes down that channel.

ldelossa (approved):
LGTM

PR #968, opened 2020-04-08
config: rework auth config

hdonnay:
Over the course of writing this documentation, I realized that the
current authentication configuration was going to present problems when
running in non-combo modes and may even prevent running with more
automatic/secure configurations, like mutual TLS.

The code changes here are geared toward making those possible while
keeping a single-config setup.

### DIFF
--- a/Documentation/operation.md
+++ b/Documentation/operation.md
@@ -1,0 +1,106 @@
+# Operation
+
+## Releases
+
+All of the source code needed to build clair is packaged as an archive and
+attached to the release. Releases are tracked at the [github releases].
+
+[github releases]: https://github.com/quay/clair/releases
+
+## Official Containers
+
+Clair is officially packaged and released as a container at
+[quay.io/projectquay/clair]. The `latest` tag tracks the git development branch,
+and version tags are built from the corresponding release.
+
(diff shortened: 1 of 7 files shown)


Documentation/operation.md
```diff
@@ -0,0 +1,106 @@
...
+[Notifier]) or run as a single monolith. Each process talks to separate tables
+in the database and is responsible for disparate API endpoints.
+
+[Indexer]: #indexer
+[Matcher]: #matcher
+[Notifier]: #notifier
+
+### Indexer
+
+Responsible for ...
```
ldelossa, line 31: should we just wrap these up in this pr?

config/config.go
```diff
@@ -47,9 +47,37 @@ type Config struct {
 	Metrics Metrics `yaml:"metrics"`
 }
 
+// Auth holds the specific configs for different authentication methods.
+//
+// These should be pointers to structs, so that it's possible to distinguish
+// between "absent" and "present and misconfigured."
```
ldelossa, line 53: :+1:

ldelossa (dismissed):
dope.

ldelossa (approved):

PR #973, opened 2020-04-16
Path Routes, Bug Stomping

ldelossa:
Traefik is now included in the local dev to confirm
path routing works and to provide a single url for local development.

Cliarctl is configured to point to the traefik instance doing the path
routing by default.

Found a few bugs along the way:
http.Transport in config.HttpClient was not checked for nil and did not provide
a default so we were getting nil deref panics on test.

HttpTransport client code checked bool before err and was masking an
issue. Actual error was bad local dev config value for indexer_addr in
matcher config

### DIFF
--- a/config/httpclient.go
+++ b/config/httpclient.go
@@ -14,6 +14,9 @@ import (
 // It returns an *http.Client and a boolean indicating whether the client is
 // configured for authentication, or an error that occurred during construction.
 func (cfg *Config) Client(next *http.Transport) (c *http.Client, authed bool, err error) {
+	if next == nil {
+		next = &http.Transport{}
+	}
 	authed = false
 	sk := jose.SigningKey{Algorithm: jose.HS256}
 
--- a/config/httpclient.go
+++ b/config/httpclient.go
@@ -15,7 +15,7 @@ import (
 // configured for authentication, or an error that occurred during construction.
 func (cfg *Config) Client(next *http.Transport) (c *http.Client, authed bool, err error) {
 	if next == nil {
-		next = &http.Transport{}
+		next = http.DefaultTransport.(*http.Transport).Clone()
 	}
 	authed = false
 	sk := jose.SigningKey{Algorithm: jose.HS256}
--- a/httptransport/server.go
+++ b/httptransport/server.go
@@ -64,15 +64,31 @@ func New(ctx context.Context, conf config.Config, indexer indexer.Service, match
 		traceOpt: othttp.WithTracer(global.TraceProvider().Tracer("clair")),
 	}
 
+	var e error
 	switch conf.Mode {
 	case config.ComboMode:
-		t.configureComboMode()
-		t.configureUpdateEndpoints()
+		e = t.configureComboMode()
+		if e != nil {
+			return nil, e
+		}
+		e = t.configureUpdateEndpoints()
+		if e != nil {
+			return nil, e
+		}
 	case config.IndexerMode:
-		t.configureIndexerMode()
+		e = t.configureIndexerMode()
+		if e != nil {
+			return nil, e
+		}
 	case config.MatcherMode:
-		t.configureMatcherMode()
-		t.configureUpdateEndpoints()
+		e = t.configureMatcherMode()
+		if e != nil {
+			return nil, e
+		}
+		e = t.configureUpdateEndpoints()
+		if e != nil {
+			return nil, e
+		}
 	}
 
 	// attach HttpTransport to server, this works because we embed http.ServeMux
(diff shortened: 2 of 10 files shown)


httptransport/server.go
```diff
@@ -20,12 +20,12 @@ import (
 
 const (
 	apiRoot                 = "/api/v1/"
-	VulnerabilityReportPath = apiRoot + "vulnerability_report/"
-	IndexAPIPath            = apiRoot + "index_report"
-	IndexReportAPIPath      = apiRoot + "index_report/"
-	StateAPIPath            = apiRoot + "state"
+	VulnerabilityReportPath = apiRoot + "matcher/vulnerability_report/"
```
alecmerdler: I understand why you are doing this, but keep these things in mind:
- The API implementation is bending to the limitations of OpenShift router
- The API resources are now explicitly tied to different services under the prefix
- The API is moving further away from being RESTful, which may not matter, but helps clair-ify (pun intended) future API decisions

When there is full service discovery, the paths don't _actually_ matter, because the clients don't have to hardcode them. What they care about are _operation IDs_ (like `GetIndexReport`), which if they can be discovered from a single OpenAPI endpoint, solves all your problems and you are free to route your services however you want. That seems like a much more scalable and supportable long-term solution, which because the API is small right now, is feasible without a lot of effort. And I would be happy to help.

alecmerdler (changes requested):
Blocking so we can discuss further

httptransport/server.go
ldelossa: "The API implementation is bending to the limitations of OpenShift router"

That is not true. This change makes placing any path router in front of Clair easier. Think about what an ops person needs to do. They can simply route "/indexer/" prefix to the backend service with this change and little configuration.

The alternative is having the ops person configure an arbitrary prefix, having their load balancer strip that prefix, and sending the stripped suffix to the services. Their clients  will need to change, and in a way dependent on their infra when deployed on prem. THIS example requires clients talking to the path router to bend to the infra. 

The other alternative is specifying a path route per api endpoint in a path router config. This is not great either as every time the services add an endpoint the ops engineer needs to update their infra, they must provide larger configs to their path router, and the margin for error increases meaning more tickets. 

"The API resources are now explicitly tied to different services under the prefix"

But... this is how we "do" micro-services.? Each service owns a resource, each resource has a route which points to particular service. What goals do you accomplish by decoupling http routes from their services? Do we expect index_reports to be suddenly handled by the matcher service?

"The API is moving further away from being RESTful, which may not matter, but helps clair-ify (pun intended) future API decisions"

I dont see how adding a "root" is breaking RESTFUL. Just as "api/v1" is a common root. "api/v1/indexer" is a common route for indexer resources. 


Also, i havent seen OpenApi used as a mechanism to route container and server traffic before. I am under the impression most ppl will be using ingress path routing, nginx, or traefik.

config/httpclient.go
```diff
@@ -14,6 +14,9 @@ import (
 // It returns an *http.Client and a boolean indicating whether the client is
 // configured for authentication, or an error that occurred during construction.
 func (cfg *Config) Client(next *http.Transport) (c *http.Client, authed bool, err error) {
+	if next == nil {
+		next = &http.Transport{}
```
hdonnay: This must have slipped during a rebase, sorry. It should be:

```suggestion
		next = http.DefaultTransport.(*http.Transport).Clone()
```

httptransport/server.go
```diff
@@ -67,16 +67,12 @@ func New(ctx context.Context, conf config.Config, indexer indexer.Service, match
 	switch conf.Mode {
 	case config.ComboMode:
 		t.configureComboMode()
-		if err := t.configureUpdateEndpoints(); err != nil {
```
hdonnay, line 70: The returned error should be checked.

hdonnay (changes requested):

hdonnay (dismissed):
LGTM

hdonnay (dismissed):

httptransport/discoveryhandler.go
```diff
@@ -0,0 +1,48 @@
...
+
+	je "github.com/quay/claircore/pkg/jsonerr"
+)
+
+var once sync.Once
+var buf []byte
+
+// DiscoveryHandler reads in an openAPI file, creates a global buffer holding
+// the contents, and serves this buffer on request.
+func DiscoveryHandler(openAPI string) (http.HandlerFunc, error) {
```
alecmerdler: Fantastic!

alecmerdler (dismissed):

hdonnay (approved):

PR #975, opened 2020-04-17
httptransport: serve OpenAPI definition

hdonnay:
This commit adds a new endpoint and associated helpers and tests.

Co-authored-by: ldelossa <ldelossa@redhat.com>


ldelossa (dismissed):

httptransport/discoveryhandler.go
```diff
@@ -0,0 +1,50 @@
...
+			if bail {
+				resp := &je.Response{
+					Code:    "unknown accept type",
+					Message: "endpoint only allows application/json or application/vnd.oai.openapi+json",
+				}
+				je.Error(w, resp, http.StatusBadRequest)
+				return
+			}
+		}
+		w.Header().Add("content-type", "application/vnd.oai.openapi+json")
```
alecmerdler: I don't think you can technically have more than one `Content-Type` header. Use content negotiation to simply return whatever the request's `Accept` header was set to.

httptransport/discoveryhandler_gen.go
```diff
@@ -0,0 +1,8 @@
+ // Code generated by embed_openapi.go, DO NOT EDIT.
```
alecmerdler: How much does this bloat your binary by now?

ldelossa (approved):

PR #984, opened 2020-04-21
config: implement base64 -> []byte conversion

hdonnay:
I mistakenly assumed that because the json package implemented this, the
package implementing a json "superset" would also.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"encoding/base64"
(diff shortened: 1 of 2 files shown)


ldelossa (commented):
Tested this with a config?

hdonnay:
yeah, added tests to make sure the yaml nodes unmarshal correctly

ldelossa (approved):

PR #985, opened 2020-04-23
database: add mapping for Ubuntu Focal (20.04)

aviadatsnyk:
Ubuntu 20.04 is [released](https://twitter.com/ubuntu/status/1253378475295232003). This is needed for it to be supported by clair.


ldelossa (approved):
Thanks!

PR #990, opened 2020-04-28
Fixed Clair-pg dependencies to work with Kubernetes 1.16+ + Improved Makefile

bertrand-benoit:
In [Kubernetes 1.16 Changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals) we can see various endpoint changed from various v1beta[1|2] to v1; breaking Clair's dependency with Postgresql (whose Chart version 5.1.2 still uses the old Kubernetes endpoints).

It is fixed since [Postgresql Chart version 6.3.9](https://github.com/helm/charts/pull/17281).

The fix on Makefile allows to be really idempotent without having to purge the Chart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,9 +42,8 @@ deploy-local:
 	# Both minikube and docker desktop is supported.
 	./local-dev/build.sh
 	-helm dependency update ./local-dev/helm/clair-pg
-	-helm install --name clair-pg ./local-dev/helm/clair-pg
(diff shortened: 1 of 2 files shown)


bertrand-benoit:
Any news?

ldelossa (approved):
Thanks!

PR #1009, opened 2020-06-02
remove type alias infavor of pointer to struct

ldelossa:
PR removes the need for a private and hidden implementation for a shareable HTTP client. 
This was self-documented as "weird" and I believe the original thought was it *needed* to remain private which is not the case.


hdonnay (approved):
This seems fine. I think my use of the export trick was to keep the unexported-ness that was there prior to my changes.

PR #1010, opened 2020-06-11
Louis/update ops api

ldelossa:
This PR bumps claircore to v0.0.25, plumbs the updated Updates api into ClairV4, and addresses some usability issues with the client.

### DIFF
--- a/httptransport/client/differ.go
+++ b/httptransport/client/differ.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"net/http"
 	"strings"
 	"sync"
@@ -152,25 +151,11 @@ func (c *HTTP) updateOperations(ctx context.Context, req *http.Request, cache *u
 		// check for etag, if exists store the value and add returned map
 		// to cache
 		if v := res.Header.Get("etag"); v != "" && !strings.HasPrefix(v, "W/") {
-			cache.Lock()
-			cache.uo = m
-			cache.validator = v
-			cache.Unlock()
+			cache.Set(m, v)
 		}
-		return m, nil
+		return cache.Copy(), nil
 	case http.StatusNotModified:
-		// received conditional response, serve map from cache
-		cache.RLock()
-		// make a copy
-		m := map[string][]driver.UpdateOperation{}
-		for u, ops := range cache.uo {
-			o := make([]driver.UpdateOperation, len(ops), len(ops))
-			copy(o, ops)
-			log.Printf("%v %v", o, ops)
-			m[u] = o
-		}
-		cache.RUnlock()
-		return m, nil
+		return cache.Copy(), nil
 	default:
 	}
 	return nil, fmt.Errorf("%v: unexpected status: %s", req.URL.Path, res.Status)
--- a/httptransport/client/differ.go
+++ b/httptransport/client/differ.go
@@ -5,8 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
 	"net/http"
-	"path"
 	"strings"
 	"sync"
 
@@ -21,7 +21,7 @@ var _ matcher.Differ = (*HTTP)(nil)
 
 // DeleteUpdateOperations attempts to delete the referenced update operations.
 func (c *HTTP) DeleteUpdateOperations(ctx context.Context, ref ...uuid.UUID) error {
-	u, err := c.addr.Parse(httptransport.UpdatesAPIPath)
+	u, err := c.addr.Parse(httptransport.UpdateOperationAPIPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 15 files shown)


httptransport/indexreporthandler.go
```diff
@@ -12,18 +12,6 @@ import (
 	"github.com/quay/clair/v4/indexer"
 )
 
-// unmodified determines whether to return a conditonal response
```
ldelossa, line 15: not deleted, just moved to a place indicating its shared.

httptransport/client/differ.go
```diff
@@ -122,19 +145,35 @@ func (c *HTTP) LatestUpdateOperations(ctx context.Context) (map[string]uuid.UUID
...
 	case http.StatusNotModified:
-		return nil, ErrUnchanged
+		// received conditional response, serve map from cache
+		cache.RLock()
+		// make a copy
+		m := map[string][]driver.UpdateOperation{}
+		for u, ops := range cache.uo {
+			o := make([]driver.UpdateOperation, len(ops), len(ops))
+			copy(o, ops)
+			log.Printf("%v %v", o, ops)
```
hdonnay:
```suggestion
```

httptransport/client/differ.go
```diff
@@ -5,8 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
```
hdonnay:
```suggestion
```

httptransport/updatediffhandler.go
```diff
@@ -4,158 +4,72 @@ import (
...
-	cur, err = uuid.Parse(curQ)
-	if err != nil {
-		res := &jsonerr.Response{
-			Code:    "update-error",
-			Message: fmt.Sprintf("malformed ref: %q", curQ),
+		// prev param is optional.
+		var prev uuid.UUID
+		var err error
+		if param, ok := r.URL.Query()["prev"]; ok {
+			if len(param) != 0 {
```
hdonnay: I don't think this param can exist and not have any values. `param := r.URL.Query().Get("prev"); param != ""` would be simpler than the double-if.

httptransport/updateoperationhandler.go
```diff
@@ -0,0 +1,126 @@
...
+// an etag.
+
+// Clients may provide an 'If-None-Match' header with the etag value to receive
+// a StatusNotModified when no new UpdateOperations have been created.
+func (h *UOHandler) Get(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	// handle conditional request. this is an optimization
+	if ref, err := h.serv.LatestUpdateOperation(ctx); err == nil {
+		var validator string
+		validator = `"` + ref.String() + `"`
```
hdonnay:
```suggestion
		validator := `"` + ref.String() + `"`
```

httptransport/updateoperationhandler.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		w.Header().Set("etag", validator)
+	}
+
+	var latest string
+	if param, ok := r.URL.Query()["latest"]; ok {
+		if len(param) != 0 {
+			latest = param[0]
+		}
+	}
```
hdonnay:
```suggestion
	latest := r.URL.Query().Get("latest")
```

httptransport/client/differ.go
```diff
@@ -122,19 +145,35 @@ func (c *HTTP) LatestUpdateOperations(ctx context.Context) (map[string]uuid.UUID
...
+		return m, nil
 	default:
-		return nil, fmt.Errorf("%v: unexpected status: %s", u.Path, res.Status)
-	}
-	m := make(map[string]uuid.UUID)
-	if err := json.NewDecoder(res.Body).Decode(&m); err != nil {
-		return nil, err
 	}
-	return m, nil
+	return nil, fmt.Errorf("%v: unexpected status: %s", req.URL.Path, res.Status)
```
hdonnay, line 161: In the above switch, the first case returns the map directly, and the second case makes a copy. This might be better expressed by adding a `Copy` method to the cache object, having the cases acquire and defer the lock, move this error return into the default case, and have this line return `cache.Copy(), nil`

hdonnay (changes requested):

httptransport/client/differ.go
ldelossa: yup, left over debug log :+1:

httptransport/client/differ.go
ldelossa: good call, missed that we store the returned value. Copy method sounds like a good idea.

httptransport/updatediffhandler.go
ldelossa: think you're right, the Get() is a nicer api for this.

httptransport/client/differ.go
ldelossa: I added a `cache.Set` to remove any lock usage from the caller, as it got kinda weird that the caller needed to use the lock only sometimes.

hdonnay (approved):

PR #1015, opened 2020-07-06
Louis/notifier

ldelossa:
PR implementing ClairV4 notifications.


config.yaml.sample
```diff
@@ -36,6 +36,27 @@ matcher:
...
+  webhook:
+    target: "http://webhook/"
+    callback: "http://clair-notifier/api/v1/notifications"
+  amqp:
+    exchange:
+        name: "" 
+        type: "direct"
+        durable: true
+        auto_delete: false
+    uri: "amqp://user:pass@host:10000/vhost"
```
Allda: Does this support multiple brokers with failover?

local-dev/clair/config.yaml
```diff
@@ -12,6 +12,26 @@ matcher:
...
+  indexer_addr: http://clair-indexer:8080/
+  matcher_addr: http://clair-matcher:8080/
+  connstring: host=clair-db port=5432 user=clair dbname=clair sslmode=disable
+  migrations: true
+  delivery_interval: 5s
+  poll_interval: 5s
+  # webhook:
+  #   target: "http://webhook/"
+  #   callback: "http://clair-notifier/api/v1/notifications"
+  amqp:
```
Allda, line 25: amqp plugin should support tls auth

notifier/amqp/deliverer.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return &clairerror.ErrDeliveryFailed{err}
+	}
+	msg := samqp.Publishing{
+		ContentType: "application/json",
+		AppId:       "clairV4-notifier",
+		Body:        b,
+	}
+	err = ch.Publish(
+		d.conf.Exchange.Name,
```
Allda, line 73: Is this a queue/topic where message is sent?

notifier/amqp/deliverer.go
```diff
@@ -0,0 +1,77 @@
...
+		d.conf.RoutingKey,
```
Allda, line 74: What is routing key?

notifier/postgres/deletenotifications.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+	"github.com/rs/zerolog"
+)
+
+// deleteNotifications garbage collects notifications and their associated
+// id and receipt rows
+func deleteNotifications(ctx context.Context, pool *pgxpool.Pool, id uuid.UUID) error {
```
Allda, line 14: Do you really need to delete notifications from the database? Isn't it better to mark it as processed and keep it in DB for future debugging in case something went wrong?

Allda (commented):
I made a very quick review of changes mainly focused on amqp part. I added a few comments there because I don't think it supports what we agree on in: https://projects.engineering.redhat.com/browse/CLOUDWF-136

I'll try to run it locally to see a real notification traffic.

config.yaml.sample
ldelossa: Wouldnt failover be handles by a proxy service ? If multiple urls is provided, what would be the correct one to connect to?

notifier/amqp/deliverer.go
ldelossa: These are terms used in AMQP spec. https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf

Reviewing this will clear things up. An exchange routes messages to a queue given some rules.

notifier/amqp/deliverer.go
ldelossa: This is a term in the AMQP spec: https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf

A routing key tells an exchange which queue the message is destined for in the usual case.

notifier/postgres/deletenotifications.go
ldelossa: If a notification is not delivered, it is marked undelivered and notifier will retry. 

After we successfully deliver however, we need to GC at some point, or we will always grow the DB unbounded.

local-dev/clair/config.yaml
ldelossa: Ok i think i can add this in.

notifier/postgres/setdeleted.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+)
+
+// setDeleted will update the receipt's status to "deleted" for the provied
+// notification id
+func setDeleted(ctx context.Context, pool *pgxpool.Pool, id uuid.UUID) error {
+	const (
+		query = `UPDATE receipt SET status = 'deleted', ts = $1 WHERE notification_id = $2`
```
hdonnay: If you use the database's clock when creating the receipt row, you should use it when updating the row:
```suggestion
		query = `UPDATE receipt SET status = 'deleted', ts = CURRENT_TIMESTAMP WHERE notification_id = $1`
```

notifier/postgres/setdelivered.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+)
+
+// setDelivered will update the receipt's status to "delivered" for the provided
+// notification id
+func setDelivered(ctx context.Context, pool *pgxpool.Pool, id uuid.UUID) error {
+	const (
+		query = `UPDATE receipt SET status = 'delivered', ts = $1 WHERE notification_id = $2`
```
hdonnay: Same comment about using the database's clock.

notifier/postgres/setdeliveryfailed.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+)
+
+// setDeliveryFailed will update the receipt's status to "delivery_failed" for the provied
+// notification id
+func setDeliveryFailed(ctx context.Context, pool *pgxpool.Pool, id uuid.UUID) error {
+	const (
+		query = `UPDATE receipt SET status = 'delivery_failed', ts = $1 WHERE notification_id = $2`
```
hdonnay: database clock

notifier/postgres/setdeliveryfailed.go
```diff
@@ -0,0 +1,28 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+)
+
+// setDeliveryFailed will update the receipt's status to "delivery_failed" for the provied
+// notification id
+func setDeliveryFailed(ctx context.Context, pool *pgxpool.Pool, id uuid.UUID) error {
```
hdonnay, line 13: All these functions that update the receipt table should just go in one file. They all take the same parameter(s) with a variable query and handle the results the same way.

notifier/postgres/teststore.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/quay/claircore/test/integration"
+	"github.com/remind101/migrate"
+)
+
+const (
+	// connection string for our local development. see docker-compose.yaml at root
+	DefaultDSN = `host=localhost port=5432 user=clair dbname=clair sslmode=disable`
+)
+
+func TestStore(ctx context.Context, t testing.TB) (*sqlx.DB, *Store, func()) {
```
hdonnay: I think this would be better if written to just return `(*pgxpool.Pool, func())`, because the caller can then either make a store using the pool or use it to examine the database directly or, as is usual in tests, both. It worked well for claircore's vulnstore with minimal changes.

pkg/pager/pager.go
```diff
@@ -0,0 +1,60 @@
...
+
+// SQLPager implements a bare-minimum sql paging algorithm.
+//
+// The input Page's Next field represents the page to calculate.
+// The output Page's Next field tells the client the next page to fetch.
+// The output Page's Next field will be -1 when returning the final page.
+//
+// The offset and limit output arguments are to be used in a sql query to
+// retrieve the correct subset of elements for the requested page.
+func SQLPager(inPage Page, count uint64) (outPage Page, offset uint64, limit uint64) {
```
hdonnay: `pager.SQLPager` is a clunky name

pkg/pager/pager.go
```diff
@@ -0,0 +1,60 @@
+package pager
```
hdonnay: Why is this a separate package? If this is part of the notifier API, it should just live in that package.

pkg/pager/pager.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// calculate the total number of pages necessary when
+	// the desired page size is provided and take the ceiling.
+	// if pages <= 0 bump it to one, we always deliver at least
+	// one page, even if empty.
+	var pages float64 = float64(count) / float64(inPage.Size)
+	pages = math.Ceil(pages)
+	if pages <= 0 {
+		pages++
+	}
```
hdonnay: Isn't this just:

```
pages := count/size
if count % size != 0 {
    pages++
}
if pages == 0 {
    pages++
}
```

pkg/pager/pager.go
```diff
@@ -0,0 +1,60 @@
+package pager
+
+import "math"
+
+// Page communicates a bare-minimum paging procotol with clients
+type Page struct {
+	// the max number of elements returned in a page
+	Size uint64
+	Next int64
+}
```
hdonnay: having `Next` instead be something like `Cursor uuid.UUID` would allow non-offset based paging.

notifier/postgres/notifications.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// notifications returns a slice of notifications associated with the provided
+// notification id
+//
+// this functions treats in page.Next argument as the page to fetch and return.
+func notifications(ctx context.Context, pool *pgxpool.Pool, id uuid.UUID, page *pager.Page) ([]notifier.Notification, pager.Page, error) {
+	const (
+		query      = "SELECT id, body FROM notification_body WHERE notification_id = $1"
+		pagedQuery = "SELECT id, body FROM notification_body WHERE notification_id = $1 OFFSET $2 LIMIT $3"
```
hdonnay: This query's response time will get worse the further "back" it pages. Both `pagedQuery` and `query` *always* return the same number of rows from the execution engine. `PagedQuery` just then has the server do a linear read to the offset row, then return a limited number of them.

```suggestion
		pagedQuery = "SELECT id, body FROM notification_body WHERE notification_id = $1 AND id < $2 ORDER BY id desc LIMIT $3"
```
This, along with making the Page structure record the id of the last item in the page should solve that. Solution untested, though.

notifier/postgres/failed.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+)
+
+// created will return all notification ids in "created" status
```
hdonnay: Doc copy-paste. This logic should probably be combined with the `created` and `deleted` functions and just pass in the status to look for.

notifier/postgres/deleted.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+)
+
+// deleted will return all notification ids in "deleted" status
+func deleted(ctx context.Context, pool *pgxpool.Pool) ([]uuid.UUID, error) {
```
hdonnay, line 12: See `failed` comment.

notifier/postgres/created.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	clairerror "github.com/quay/clair/v4/clair-error"
+)
+
+// created will return all notification ids in "created" status
+func created(ctx context.Context, pool *pgxpool.Pool) ([]uuid.UUID, error) {
```
hdonnay, line 12: See `failed` comment.

notifier/postgres/deletenotifications.go
```diff
@@ -0,0 +1,59 @@
...
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "notifier/postgres/deleteNotification").
+		Logger()
+
+	const (
+		deleteNotificationID = `DELETE FROM notification WHERE id = $1`
+		deleteNotifications  = `DELETE FROM notification_body WHERE notification_id = $1`
+		deleteReceipt        = `DELETE FROM receipt WHERE notification_id = $1`
+	)
```
hdonnay, line 23: Can't these cascade off the notification row?

notifier/store.go
```diff
@@ -0,0 +1,85 @@
...
+	// Created returns a slice of notification ids in created status
+	Created(ctx context.Context) ([]uuid.UUID, error)
+	// Failed returns a slice of notification ids to in delivery failed status
+	Failed(ctx context.Context) ([]uuid.UUID, error)
+	// Deleted returns a slice of notification ids in deleted status
+	Deleted(ctx context.Context) ([]uuid.UUID, error)
+	// SetDelivered marks the provided notification id as delivered
+	SetDelivered(ctx context.Context, id uuid.UUID) error
+	// SetDeliveryFailed marks the provided notification id failed to be delivered
+	SetDeliveryFailed(ctx context.Context, id uuid.UUID) error
```
hdonnay, line 81:
```suggestion
	// SetFailed marks the provided notification id failed to be delivered.
	SetFailed(ctx context.Context, id uuid.UUID) error
```

notifier/store.go
```diff
@@ -0,0 +1,85 @@
...
+	// them with the provided notification id
+	//
+	// PutNotifications must update the latest update operation for the provided
+	// updater in such a way that UpdateOperation returns the the provided update
+	// operation id when queried with the updater name
+	//
+	// PutNotifications must create a Receipt with status created status on
+	// successful persistence of notifications in such a way that Receipter.Created()
+	// returns the persisted notification id.
+	PutNotifications(ctx context.Context, opts PutOpts) error
```
hdonnay, line 55: This struct should just be arguments.

Put may be a weird verb to use here, it suggests HTTP `PUT` (like, updatable) to me, but this is a create/store method.

notifier/postgres/putnotifications.go
```diff
@@ -0,0 +1,82 @@
...
+			return clairerror.ErrPutNotifications{opts.NotificationID, err}
+		}
+	}
+	err = mBatch.Done(ctx)
+	if err != nil {
+		return clairerror.ErrPutNotifications{opts.NotificationID, err}
+	}
+
+	// update known update operations
+	_, err = tx.Exec(ctx, insertUpdateOperation, opts.Updater, opts.UpdateID, time.Now())
```
hdonnay, line 67: Why is the tag only sometimes checked?

notifier/vulnsummary.go
```diff
@@ -0,0 +1,25 @@
...
+type VulnSummary struct {
+	Name         string                  `json:"name"`
+	Description  string                  `json:"description"`
+	Package      *claircore.Package      `json:"package,omitempty"`
+	Distribution *claircore.Distribution `json:"distribution,omitempty"`
+	Repo         *claircore.Repository   `json:"repo,omitempty"`
+	Severity     string                  `json:"severity"`
+}
+
+func (vs *VulnSummary) FromVulnerability(v claircore.Vulnerability) {
```
hdonnay, line 18: This should take a pointer, because we pass Vulnerabilities as pointers everywhere.

notifier/migrations/migration1.go
```diff
@@ -0,0 +1,39 @@
...
+	CREATE TABLE IF NOT EXISTS notifier_update_operation (
+		uo_id uuid PRIMARY KEY,
+		updater text,
+		ts timestamptz
+	);
+
+	--- a relation mapping notifications to their serialized json bodies
+	CREATE TABLE IF NOT EXISTS notification_body (
+		id uuid PRIMARY KEY,
+		notification_id uuid REFERENCES notification,
```
hdonnay, line 25: Cascade on delete?

notifier/migrations/migration1.go
```diff
@@ -0,0 +1,39 @@
...
+		body jsonb NOT NULL -- serialized json body of notification
+	);
+
+	--- an enumeration identifying the possible status a receipt may be in
+	CREATE TYPE receiptstatus AS ENUM ('created', 'delivered', 'delivery_failed', 'deleted');
+
+	--- a relation expressing the current status of a notification
+	--- this acts as a trigger for application business logic
+	CREATE TABLE IF NOT EXISTS receipt (
+		notification_id uuid PRIMARY KEY REFERENCES notification,
```
hdonnay, line 37: Cascade on delete?

notifier/migrations/migration1.go
```diff
@@ -0,0 +1,39 @@
...
+		uo_id uuid REFERENCES notifier_update_operation(uo_id),
```
hdonnay: If the notifications could be made to cascade on deletion, this could be handled by a trigger that deletes the update operation row if nothing in this table points references it.

notifier/processor.go
```diff
@@ -0,0 +1,231 @@
...
+//
+// Canceling the ctx will end the processing.
+func (p *Processor) Process(ctx context.Context, c <-chan Event) {
+	go p.process(ctx, c)
+}
+
+// process is intended to be ran as a go routine.
+//
+// implements a blocking event loop via a time.Ticker
+func (p *Processor) process(ctx context.Context, c <-chan Event) {
```
hdonnay: Don't have a function immediately just `go anotherfuction()`. The caller can just put `go` in front of this method call if they want, but this removes the choice.

notifier/processor.go
```diff
@@ -0,0 +1,231 @@
...
+// and updates the notifier system with the "latest" seen UOID.
+//
+// Canceling the ctx will end the processing.
+func (p *Processor) Process(ctx context.Context, c <-chan Event) {
+	go p.process(ctx, c)
+}
+
+// process is intended to be ran as a go routine.
+//
+// implements a blocking event loop via a time.Ticker
```
hdonnay: `time.Ticker`?

notifier/processor.go
```diff
@@ -0,0 +1,231 @@
...
+		log.Info().Msg("encountered first update operation. will not process notifications")
+		return false
+	}
+
+	current, prev := uos[0], uos[1]
+	if current.Ref.String() != e.uo.Ref.String() {
+		log.Info().Str("new", current.Ref.String()).Msg("newer update operation is present, will not process notifications")
+		return false
+	}
+	p.previous = prev.Ref
```
hdonnay: Doing this return via member assignment is needlessly complicated.

notifier/processor.go
```diff
@@ -0,0 +1,231 @@
...
+			vs.FromVulnerability(*vuln)
+
+			digest, err := claircore.ParseDigest(manifest)
+			if err != nil {
+				return err
+			}
+			n := Notification{
+				Manifest:      digest,
+				Reason:        r,
+				Vulnerability: vs,
```
hdonnay: Not creating a separate VulnSummary struct would avoid a copy here.

notifier/poller.go
```diff
@@ -0,0 +1,130 @@
...
+// a new UpdateOperation is discovered.
+type Event struct {
+	updater string
+	uo      driver.UpdateOperation
+}
+
+// Poll is a non blocking call which begins
+// polling the Matcher for UpdateOperations.
+//
+// Returned channel can be listened too for events.
```
hdonnay:
```suggestion
// Returned channel can be listened to for events.
```

notifier/poller.go
```diff
@@ -0,0 +1,130 @@
...
+
+	for updater, uo := range latest {
+		if len(uo) == 0 {
+			log.Debug().Msg("received 0 update operations after polling Matcher")
+			return
+		}
+		latest := uo[0]
+		// confirm notifications were never created for this UOID.
+		var errNoReceipt clairerror.ErrNoReceipt
+		_, err := p.store.ReceiptByUOID(ctx, latest.Ref)
```
hdonnay, line 111: Isn't this check also done by the `Processor`?

notifier/delivery.go
```diff
@@ -0,0 +1,169 @@
...
+		Str("deliverer", d.Deliverer.Name()).
+		Str("component", "notifier/delivery/Delivery.Deliver").Logger()
+	log.Info().Msg("delivering notifications")
+	go d.deliver(ctx)
+}
+
+// deliver is intended to be ran as a go routine.
+//
+// implements a blocking event loop via a time.Ticker
+func (d *Delivery) deliver(ctx context.Context) error {
```
hdonnay, line 54: Don't immediately just call `go` on another function. The caller can just add the go statement, but can't take it away.

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,88 @@
...
+	clairerror "github.com/quay/clair/v4/clair-error"
+	"github.com/quay/clair/v4/notifier"
+	"github.com/rs/zerolog"
+)
+
+type Deliverer struct {
+	conf Config
+	// a client to use for POSTing webhooks
+	c   *http.Client
+	log *zerolog.Logger
```
hdonnay: `Log` member seems unused.

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Deliver implements the notifier.Deliverer interface.
+//
+// Deliver POSTS a webhook data structure to the configured target.
+func (d *Deliverer) Deliver(ctx context.Context, nID uuid.UUID) error {
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "notifier/webhook/deliverer.Deliver").
+		Logger()
+
+	callback := d.conf.callback
```
hdonnay, line 86: This isn't accessed under the lock.

notifier/webhook/config.go
```diff
@@ -0,0 +1,41 @@
...
+	// the notification will be appended to this url
+	Callback string `yaml:"callback"`
+	callback *url.URL
+	// any htp headers necessary for the request to Target
+	Headers http.Header `yaml:"headers"`
+}
+
+// Validate confirms configuration is valid and fills in private members
+// with parsed values on success.
+func (c *Config) Validate() error {
```
hdonnay: Why are only some the of the private members updated under lock? The private members also don't seem to be accessed under the lock.

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,88 @@
...
+
+	wh := notifier.Callback{
+		NotificationID: nID,
+		Callback:       *callback,
+	}
+	b, err := json.Marshal(&wh)
+	if err != nil {
+		return err
+	}
+	buf := bytes.NewBuffer(b)
```
hdonnay:
```suggestion
	buf := bytes.NewReader(b)
```

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,88 @@
...
+		Method: http.MethodPost,
+	}
+
+	log.Info().Str("notification_id", nID.String()).
+		Str("callback", callback.String()).
+		Str("target", d.conf.Target).
+		Msg("dispatching webhook")
+
+	resp, err := d.c.Do(req)
+	if err != nil {
```
hdonnay, line 128:
```suggestion
	if resp != nil {
		defer resp.Body.Close()
	}
	if err != nil {
```

notifier/amqp/config.go
```diff
@@ -0,0 +1,125 @@
...
+				return fmt.Errorf("amqps:// broker requires tls_cert")
+			}
+			if c.TLSKey == "" {
+				return fmt.Errorf("amqps:// broker requires tls_key")
+			}
+		}
+	}
+	if c.TLSRootCA != "" {
+		TLS := tls.Config{}
+		TLS.RootCAs = x509.NewCertPool()
```
hdonnay:
```suggestion
		TLS.RootCAs = x509.SystemCertPool()
```

This would make adding a CA to the system store work as expected, and make the CA argument optional.

notifier/amqp/failover.go
```diff
@@ -0,0 +1,94 @@
...
+
+	err := f.Config.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	f.RLock()
+	if f.conn != nil && !f.conn.IsClosed() {
+		log.Debug().Msg("existing connection exist and is not closed. returning this connection")
+		return f.conn, nil
```
hdonnay, line 37: This return leaves a read lock dangling.

notifier/amqp/config.go
```diff
@@ -0,0 +1,125 @@
...
+	Type string `yaml:"type"`
+	// Whether the exchange survives server restarts
+	Durable bool `yaml:"durability"`
+	// Whether bound consumers define the lifecycle of the Exchange.
+	AutoDelete bool `yaml:"auto_delete"`
+}
+
+// Config provides configuration for an AMQP deliverer.
+type Config struct {
+	sync.Mutex
```
hdonnay: A `FailOver` and a `Deliverer` can't share a config because of this lock, is that desirable?

notifier/amqp/deliverer.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return &clairerror.ErrDeliveryFailed{err}
+	}
+
+	ch, err := conn.Channel()
+	if err != nil {
+		return &clairerror.ErrDeliveryFailed{err}
+	}
+
+	callback := d.conf.callback
```
hdonnay, line 56: Not accessed under lock

notifier/amqp/failover.go
```diff
@@ -0,0 +1,94 @@
...
+		// 'DialTLS will use the provided tls.Config when it encounters an amqps:// scheme and will dial a plain connection when it encounters an amqp:// scheme.'
+		conn, err := samqp.DialTLS(uri, &f.tls)
+		if err != nil {
+			if conn != nil {
+				conn.Close()
+			}
+			log.Info().Str("broker", uri).Msg("failed to connect to AMQP broker. attempting next broker")
+			continue
+		}
+		ch, err := conn.Channel()
```
hdonnay, line 52: The `ch.Close()` method is only called on the success path. Is that OK?

config/config.go
```diff
@@ -51,190 +48,17 @@ type Config struct {
...
-}
-
-type Trace struct {
-	Name        string   `yaml:"name" json:"name"`
-	Probability *float64 `yaml:"probability" json:"probability"`
-	Jaeger      Jaeger   `yaml:"jaeger" json:"jaeger"`
+	LogLevel string  `yaml:"log_level" json:"log_level"`
+	Indexer  Indexer `yaml:"indexer" json:"indexer"`
+	Matcher  Matcher `yaml:"matcher"`
+	Notifier Notifier
```
hdonnay: Missing tag

config/introspection.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	Collector struct {
+		Endpoint string  `yaml:"collector_endpoint" json:"endpoint"`
+		Username *string `yaml:"username" json:"username"`
+		Password *string `yaml:"password" json:"password"`
+	}
+	ServiceName string            `yaml:"service_name" json:"service_name"`
+	Tags        map[string]string `yaml:"tags" json:"tags"`
+	BufferMax   int               `yaml:"buffer_max" json:"buffer_max"`
+}
```
hdonnay: This changes the config by removing the `inline` annotation to the struct tags.

The newly un-inlined keys are now `Capped` instead of `lower`.

The configuration options now stutter: `Agent: {agent_endpoint: ""}`.

These changes shouldn't be made in this commit, if at all.

notifier/service/service.go
```diff
@@ -0,0 +1,223 @@
...
+func (s *service) Notifications(ctx context.Context, id uuid.UUID, page *pager.Page) ([]notifier.Notification, pager.Page, error) {
+	return s.store.Notifications(ctx, id, page)
+}
+
+func (s *service) DeleteNotifications(ctx context.Context, id uuid.UUID) error {
+	return s.store.DeleteNotifications(ctx, id)
+}
+
+// Opts configures the notifier service
+type Opts struct {
```
hdonnay, line 70: This is just a re-statement of `config.Notifier`, why is it needed?

notifier/service/service.go
```diff
@@ -0,0 +1,223 @@
...
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/jmoiron/sqlx"
+	"github.com/quay/clair/v4/indexer"
+	"github.com/quay/clair/v4/matcher"
+	"github.com/quay/clair/v4/notifier"
+	namqp "github.com/quay/clair/v4/notifier/amqp"
```
hdonnay, line 14: Why is this renamed?

notifier/service/service.go
```diff
@@ -0,0 +1,223 @@
...
+		return fmt.Errorf("amqp validation failed: %v", err)
+	}
+
+	if len(opts.AMQP.URIs) == 0 {
+		log.Warn().Msg("amqp delivery was configured with no broker URIs to connect to. delivery of notifications will not occur.")
+		return nil
+	}
+
+	fo := &namqp.FailOver{
+		Config: *opts.AMQP,
```
hdonnay: This has an embedded lock, it's not okay to copy around.

notifier/service/service.go
```diff
@@ -0,0 +1,223 @@
...
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "notifier/service/webhookInit").
+		Logger()
+	ctx = log.WithContext(ctx)
+	log.Info().Int("count", deliveries).Msg("intializing webhook deliverers")
+
+	ds := make([]*notifier.Delivery, 0, deliveries)
+	for i := 0; i < deliveries; i++ {
+		distLock := pgdl.NewLock(lockPool, 0)
+		wh, err := webhook.New(*opts.Webhook, nil)
```
hdonnay: This has an embedded lock, it's not OK to copy.

httptransport/client/matcher.go
```diff
@@ -10,13 +11,39 @@ import (
...
+	b, err := json.Marshal(&ir)
+	if err != nil {
+		return nil, err
+	}
+	buf := bytes.NewBuffer(b)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), buf)
+	if err != nil {
+		return nil, err
+	}
```
hdonnay, line 38: `Req.Body` isn't closed, status code isn't checked.

httptransport/notificationshandler.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	// optional page parameter
+	var page int64
+	if param := r.URL.Query().Get("page"); param != "" {
+		page, err = strconv.ParseInt(param, 10, 64)
+		if err != nil {
+			resp := &je.Response{
+				Code:    "bad-request",
+				Message: "could not parse \"page\" query param into uuid",
```
hdonnay: uuid?

httptransport/notificationshandler.go
```diff
@@ -0,0 +1,160 @@
...
+	// pages are not 0 indexed.
+	if page == 0 {
+		page = 1
+	}
+
+	inP := &pager.Page{
+		Size: pageSize,
+		Next: page,
+	}
+	notifications, outP, err := h.serv.Notifications(ctx, notificationID, inP)
```
hdonnay, line 137: Does this mean it's impossible to call without paging?

httptransport/server.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 
+	notifier "github.com/quay/clair/v4/notifier/service"
```
hdonnay, line 9: group with the local imports

httptransport/server.go
```diff
@@ -300,7 +336,7 @@ func unmodified(r *http.Request, v string) bool {
 	return false
 }
 
-// WriterError is a helper that closes over an error that may be returned after
+// writerError is a helper that closes over an error that may be returned after
```
hdonnay, line 368: Why un-capitalize the start of a sentence?

initialize/services.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"time"
 
+	notifier "github.com/quay/clair/v4/notifier/service"
```
hdonnay, line 7: Group with the local imports

initialize/services.go
```diff
@@ -42,8 +43,39 @@ func (i *Init) Services() error {
...
+				Msg: "notifier: failed to parse delivery interval: " + err.Error(),
+			}
+		}
+		pInterval, err := time.ParseDuration(i.conf.Notifier.PollInterval)
+		if err != nil {
+			return &clairerror.ErrNotInitialized{
+				Msg: "notifier: failed to parse poll interval: " + err.Error(),
+			}
+		}
+
```
hdonnay, line 60: Ah, I see, the config and the notifier opts have some differing types.

Why doesn't the config just have the needed types? The yaml package handles Durations correctly: https://play.golang.org/p/MyhYUozoQBo

hdonnay (changes requested):

notifier/postgres/setdeliveryfailed.go
ldelossa: I'm usually a proponent of a function per file, makes finding things easier.

httptransport/notificationshandler.go
ldelossa: Yes, the API assumes paging.

httptransport/server.go
ldelossa: the comment is referring to a private function, caps have meaning.

notifier/postgres/teststore.go
ldelossa: I think I need this for distlocks

notifier/poller.go
ldelossa: It's the same idea of "check, then check under lock". There's a fast pass check here, then the processor confirms the work still needs to be done and another worker did not complete it while it was parked or waiting.

notifier/delivery.go
ldelossa: Calling go here removes the littering of go routine calls in the calling code. the usage is explained, and this provides a cleaner api to the caller.

notifier/webhook/config.go
ldelossa: Parse is called by goroutines sharing a config, the fields which are written to are locked.

initialize/services.go
ldelossa: Just following the current pattern of things.

pkg/pager/pager.go
ldelossa: The idea was to eventually move this to ClairCore and share it, however this may go away completely with a seeking pager method

pkg/pager/pager.go
ldelossa: again, probably gonna go away with seek pager implementation

notifier/store.go
ldelossa: No need to bloat the function argument list, a struct is common way to hold a grouping of arguments to a method of function.

notifier/vulnsummary.go
ldelossa: I don't want this function to modify the incoming vuln, why pass a pointer? If the calling code has a pointer, It can deref it.

notifier/processor.go
ldelossa: If the purpose of the function is to be asynchronous, and the usage is explained, removing the need for the caller to write more code is not a bad thing, the intention is for these structs to be asynchronous processors.

notifier/service/service.go
ldelossa: I've been using samqp for steadway's and namqp for notifier's

pkg/pager/pager.go
ldelossa: Think this may go away with seeking pagination.

notifier/webhook/deliverer.go
ldelossa: Deliverer now gets a copy of a Validated config. no locks necessary.

notifier/amqp/config.go
ldelossa: this will use the default cert repo on a linux system with common root ca's?

notifier/amqp/config.go
ldelossa: All configs have locks removed now, Validate() method returns a copy.

notifier/amqp/deliverer.go
ldelossa: Remedied by Validate() methods returning a copy, locks are gone.

config/introspection.go
ldelossa: A subset of these changes are already in master, these changes came in to remedy a rebase.

notifier/service/service.go
ldelossa: Locks are gone.

notifier/service/service.go
ldelossa: locks are gone.

httptransport/server.go
```diff
@@ -25,6 +26,7 @@ const (
 	IndexAPIPath            = apiRoot + "index_report"
 	IndexReportAPIPath      = apiRoot + "index_report/"
 	IndexStateAPIPath       = apiRoot + "index_state"
+	NotificationAPIPath     = apiRoot + "notification/"
```
Allda, line 29: OpenAPI spec defines it as plural ``/api/v1/notifications/

initialize/services.go
```diff
@@ -42,8 +43,39 @@ func (i *Init) Services() error {
...
+		n, err := notifier.New(i.GlobalCTX, notifier.Opts{
+			DeliveryInterval: dInterval,
+			ConnString:       i.conf.Notifier.ConnString,
+			Indexer:          libI,
+			Matcher:          libV,
+			Migrations:       i.conf.Notifier.Migrations,
+			PollInterval:     pInterval,
+			Webhook:          i.conf.Notifier.Webbook,
```
Allda, line 68: AMQP is missing in Combo mode

pkg/pager/pager.go
ldelossa: this is obsolete now.

pkg/pager/pager.go
ldelossa: this is obsolete now.

pkg/pager/pager.go
ldelossa: done.

notifier/postgres/notifications.go
ldelossa: fixed.

notifier/postgres/failed.go
ldelossa: fixed

notifier/processor.go
ldelossa: fixed

notifier/processor.go
ldelossa: the safe method is designed to just return a bool on whether or not the processor should continue processing.

notifier/webhook/deliverer.go
ldelossa: done.

notifier/amqp/failover.go
ldelossa: Think so, since every error path calls "conn.Close()" and their docs say:

"After returning from this call, all resources associated with this connection,
including the underlying io, Channels, Notify listeners and Channel consumers
will also be closed."

notifier/postgres/putnotifications.go
ldelossa: fixed

notifier/processor.go
ldelossa: good catch, updating to:

```
			// summarize most severe vuln affecting manifest
			// the vulns array will be sorted by most severe
			vuln := affected.Vulnerabilities[vulns[0]]

			digest, err := claircore.ParseDigest(manifest)
			if err != nil {
				return err
			}
			n := Notification{
				Manifest: digest,
				Reason:   r,
			}
			n.Vulnerability.FromVulnerability(*vuln)

			notifications = append(notifications, n)
```

config/config.go
ldelossa: fixed

httptransport/client/matcher.go
ldelossa: fixed

httptransport/notificationshandler.go
ldelossa: now valid, since "next" should be a uuid.

```
		n, err := uuid.Parse(param)
		if err != nil {
			resp := &je.Response{
				Code:    "bad-request",
				Message: "could not parse \"next\" query param into uuid",
			}
			je.Error(w, resp, http.StatusBadRequest)
			return
		}
```

httptransport/server.go
ldelossa: fixed

initialize/services.go
ldelossa: fixed, great catch.

notifier/postgres/setdeleted.go
ldelossa: using db timestamps now.

notifier/store.go
ldelossa: If you look a little closer, these method names are an echo of the enum names in the db. This method literally means "set the delivery_failed enumeration".

config/config.go
```diff
@@ -247,6 +71,9 @@ func Validate(conf Config) error {
 		if conf.Matcher.ConnString == "" {
 			return fmt.Errorf("matcher mode requires a database connection string")
 		}
+		if conf.Notifier.ConnString == "" {
+			return fmt.Errorf("notifier mode requires a database connection string")
+		}
```
hdonnay, line 76:
```suggestion
		if conf.Notifier.ConnString == "" {
			conf.Notifier.ConnString = conf.Matcher.ConnString
		}
```

config/introspection.go
```diff
@@ -0,0 +1,35 @@
...
+type Trace struct {
+	Name        string   `yaml:"name" json:"name"`
+	Probability *float64 `yaml:"probability" json:"probability"`
+	Jaeger      Jaeger   `yaml:"jaeger" json:"jaeger"`
+}
+
+type Jaeger struct {
+	Agent struct {
+		Endpoint string `yaml:"endpoint" json:"endpoint"`
+	} `yaml:"agent"`
```
hdonnay: Needs json tags if we're changing this because of json.

httptransport/client/matcher.go
```diff
@@ -10,13 +11,50 @@ import (
...
+
+	resp, err := c.c.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
```
hdonnay:
```suggestion
	if resp.Body != nil {
		defer resp.Body.Close()
	}
	if err != nil {
		return nil, err
	}
```

httptransport/notificationshandler.go
```diff
@@ -0,0 +1,158 @@
...
+	"path/filepath"
+	"strconv"
+
+	"github.com/google/uuid"
+
+	"github.com/quay/clair/v4/notifier"
+	"github.com/quay/clair/v4/notifier/service"
+	"github.com/quay/clair/v4/pkg/pager"
+	je "github.com/quay/claircore/pkg/jsonerr"
+	"github.com/rs/zerolog"
```
hdonnay:
```suggestion
	"github.com/google/uuid"
	je "github.com/quay/claircore/pkg/jsonerr"
	"github.com/rs/zerolog"

	"github.com/quay/clair/v4/notifier"
	"github.com/quay/clair/v4/notifier/service"
	"github.com/quay/clair/v4/pkg/pager"
```

notifier/amqp/deliverer_integration_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/google/uuid"
+	"github.com/quay/claircore/test/integration"
+	samqp "github.com/streadway/amqp"
```
hdonnay, line 13:
```suggestion
	"golang.org/x/sync/errgroup"
	"github.com/google/uuid"
	"github.com/quay/claircore/test/integration"
	samqp "github.com/streadway/amqp"
```

notifier/processor.go
hdonnay: OK, but it actually returns `(ok, previousRef)`, it just does the second one via assigning it to shared state instead of returning it.

Having this method return it and the `create` method take it would make the sequence dependency explicit.

notifier/service/service.go
```diff
@@ -0,0 +1,276 @@
...
+
+const (
+	processors = 4
+	deliveries = 4
+)
+
+// Service is an interface wrapping ClairV4's notifier functionality.
+type Service interface {
+	// Retrieves an optional paginated set of notifications given an notification id
+	Notifications(ctx context.Context, id uuid.UUID, page *pager.Page) ([]notifier.Notification, pager.Page, error)
```
hdonnay: `Pager.Page` should just be a part of the `notifier` package.

notifier/stomp/config.go
```diff
@@ -0,0 +1,97 @@
...
+	// The filesystem path where a tls private key can be read.
+	Key string `yaml:"key"`
+}
+
+type Login struct {
+	Login    string `yaml:"login"`
+	Passcode string `yaml:"passcode"`
+}
+
+type Config struct {
```
hdonnay, line 25: Yaml tags are inconsistent on this struct.

pkg/pager/pager.go
hdonnay: Yeah, there's only a single struct in here, it should just get put in the `notifier` package.

hdonnay (changes requested):

config/config.go
ldelossa: This would be odd one-off thing to do for notifier. Right now the config requires a db string be present for each service so that it works the same if you decide not to run in combo-mode.

I do see what you're getting at here, we only need to "find" one db string for combo mode, but in combo mode can also be pointed to different databases, its probably better to be explicit in regards to database strings since it works the same for any mode.

notifier/service/service.go
ldelossa: That seems not right, this struct will be used for *any* paging api we use moving forward. the chances of it being shared and being moved to another package is high.

notifier/processor.go
ldelossa: I changed it.

notifier/amqp/config.go
ldelossa: fixed

config/config.go
```diff
@@ -51,190 +48,17 @@ type Config struct {
...
-}
-
-type Trace struct {
-	Name        string   `yaml:"name" json:"name"`
-	Probability *float64 `yaml:"probability" json:"probability"`
-	Jaeger      Jaeger   `yaml:"jaeger" json:"jaeger"`
+	LogLevel string   `yaml:"log_level" json:"log_level"`
+	Indexer  Indexer  `yaml:"indexer" json:"indexer"`
+	Matcher  Matcher  `yaml:"matcher"`
+	Notifier Notifier `yaml:"notifer"`
```
Allda: s/notifer/notifier/

notifier/processor.go
```diff
@@ -0,0 +1,231 @@
...
+		return false, uuid.Nil
+	}
+	if _, ok := all[e.updater]; !ok {
+		log.Warn().Msg("updater missing from update operations returned from matcher. matcher may have garbage collected")
+		return false, uuid.Nil
+	}
+
+	uos := all[e.updater]
+	n := len(uos)
+	if n < 2 {
```
Allda, line 220: Are you going to change it to produce notification for dynamic updaters?

notifier/stomp/deliverer.go
```diff
@@ -0,0 +1,62 @@
...
+	fo   *FailOver
+}
+
+func New(conf Config, fo *FailOver) (*Deliverer, error) {
+	var c Config
+	var err error
+	if c, err = conf.Validate(); err != nil {
+		return nil, err
+	}
+	return &Deliverer{
```
Allda, line 31: Configuration is stored twice in Deliver and in Failover and both have separate TLS configurations. Why is that?

notifier/stomp/failover.go
```diff
@@ -0,0 +1,83 @@
...
+//
+// The caller MUST call conn.Disconnect() to close the underlying tcp connection
+// when finished.
+func (f *FailOver) Connection(ctx context.Context) (*gostomp.Conn, error) {
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "notifier/stomp/failover").
+		Logger()
+	ctx = log.WithContext(ctx)
+
+	_, err := f.Config.Validate()
```
Allda: This doesn't create a tls object in a config since the config is not a pointer

notifier/stomp/failover.go
```diff
@@ -0,0 +1,83 @@
...
+func (f *FailOver) Dial(uri string) (*gostomp.Conn, error) {
+	opts := []func(*gostomp.Conn) error{}
+
+	if f.Login != nil {
+		opts = append(opts, gostomp.ConnOpt.Login(f.Login.Login, f.Login.Passcode))
+	}
+
+	var conn io.ReadWriteCloser
+	var err error
+	if f.tls != nil {
```
Allda, line 37: In integration tests this is always nil, because `Validate` doesn't return updated config with tls setup

notifier/stomp/deliverer.go
```diff
@@ -0,0 +1,62 @@
...
+// the broker.
+type Deliverer struct {
+	conf Config
+	fo   *FailOver
+}
+
+func New(conf Config, fo *FailOver) (*Deliverer, error) {
+	var c Config
+	var err error
+	if c, err = conf.Validate(); err != nil {
```
Allda, line 25: `Validate()` is actually called twice and tls is also configured twice in `Deliverer` and in `Failover`. Is that really needed?

notifier/stomp/deliverer.go
ldelossa: They both use the same config. Validate() returns a copy of a Config with all private members parsed

notifier/stomp/failover.go
ldelossa: what do you mean?

notifier/stomp/deliverer.go
ldelossa: I think you're right, I'll make failOver private and have the Deliverer constructor create it.

notifier/stomp/failover.go
ldelossa: Validate() returns a copy of the config, in this case we were only confirming that the config is still valid.

config/config.go
hdonnay: I forgot the notifier has its own DB; this suggestion sucks.

httptransport/keybyidhandler.go
```diff
@@ -0,0 +1,77 @@
...
+	"errors"
+	"net/http"
+	"path"
+
+	jose "gopkg.in/square/go-jose.v2"
+
+	"github.com/google/uuid"
+	clairerror "github.com/quay/clair/v4/clair-error"
+	"github.com/quay/clair/v4/notifier"
+	je "github.com/quay/claircore/pkg/jsonerr"
```
hdonnay, line 14:
```suggestion
	jose "gopkg.in/square/go-jose.v2"
	je "github.com/quay/claircore/pkg/jsonerr"
	"github.com/google/uuid"

	clairerror "github.com/quay/clair/v4/clair-error"
	"github.com/quay/clair/v4/notifier"
```

httptransport/keybyidhandler.go
```diff
@@ -0,0 +1,77 @@
...
+		case err == nil:
+			// hop out
+		default:
+			resp := &je.Response{
+				Code:    "internal-server-error",
+				Message: err.Error(),
+			}
+			je.Error(w, resp, http.StatusInternalServerError)
+			return
+
```
hdonnay, line 66:
```suggestion
```

notifier/keymanager/manager.go
```diff
@@ -0,0 +1,210 @@
...
+// Ensure cancelation of ctx to avoid go routine leakage.
+func NewManager(ctx context.Context, store notifier.KeyStore) (*Manager, error) {
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "notifier/keymanager/NewManager").
+		Logger()
+
+	var m *Manager = &Manager{
+		kp:    atomic.Value{},
+		store: store,
+	}
```
hdonnay, line 69:
```suggestion
	m := &Manager{
		kp:    atomic.Value{},
		store: store,
	}
```

notifier/keymanager/manager.go
```diff
@@ -0,0 +1,210 @@
...
+	newKP, err := m.genKeyPair(ctx)
```
hdonnay, line 70: This shouldn't be a method because it doesn't mutate or reference anything from the Manager.

notifier/keymanager/manager.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (m *Manager) gc(ctx context.Context) {
+	log := zerolog.Ctx(ctx).With().
+		Str("component", "notifier/keymanager/Manager.gc").
+		Logger()
+
+	log.Info().Msg("gc starting")
+	var total int64
+	var err error
+	var n int64 = math.MaxInt64
```
hdonnay: Why is this initialized to MaxInt instead of just -1?

notifier/keymanager/manager.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		if err := m.store.PutKey(ctx, newKP.ID, newKP.Public); err != nil {
+			return err
+		}
+		if err := m.store.BumpExpiration(ctx, delta, kp.ID); err != nil {
+			return err
+		}
+		// atomic store
+		m.kp.Store(&newKP)
+		return nil
```
hdonnay: Why does this create a new key instead of just re-using the existing one?

notifier/migrations/migration1.go
```diff
@@ -0,0 +1,50 @@
...
+		details jsonb -- any additional details specific to the delivery mechanism
+	);
+	CREATE INDEX receipt_idx ON receipt (notification_id, uo_id);
+
+    --- a relation holding a pub_key in PKIX, ASN.1 DER form
+	--- expiration is application defined and not associated with the public key
+	CREATE TABLE IF NOT EXISTS key (
+	    id uuid PRIMARY KEY,
+		expiration timestamptz,
+		pub_key bytea
```
hdonnay: indentation messed up

notifier/migrations/migration1.go
```diff
@@ -0,0 +1,50 @@
...
+	);
+	CREATE INDEX key_idx ON key (id);
```
hdonnay: I think this index is redundant because the primary key always has an index.

notifier/processor.go
hdonnay: Yeah, this needs to return `true, uuid.NIl` and the decision to not send needs to be made once the diff size is seen.

notifier/keymanager/manager.go
```diff
@@ -0,0 +1,204 @@
...
+		newKP, err := m.genKeyPair(ctx)
+		if err != nil {
+			return err
+		}
+		if err := m.store.PutKey(ctx, newKP.ID, newKP.Public, delta); err != nil {
+			return err
+		}
+		// atomic store
+		m.kp.Store(&newKP)
+		return nil
```
hdonnay, line 195: Why does this generate a new keypair, instead of just using the existing keypair with a new expiration?

notifier/postgres/keystore.go
```diff
@@ -0,0 +1,188 @@
...
+
+	pub, err := derToRSAPublic(tmp.der)
+	if err != nil {
+		return notifier.Key{}, err
+	}
+	key := notifier.Key{
+		ID:         tmp.id,
+		Expiration: tmp.exp,
+		Public:     pub,
+	}
```
hdonnay: Can't you just Scan directly into a `Key` struct and a byte slice? I don't think the extra struct buys anything for readability.

notifier/stomp/config.go
```diff
@@ -0,0 +1,97 @@
...
+	// Specifies the number of notifications delivered in single STOMP message
+	// when Direct is true.
+	//
+	// Ignored if Direct is not true
+	// If 0 or 1 is provided no rollup occurs and each notification is delivered
+	// separately.
+	Rollup int `yaml:"rollup"`
+	// The callback url where notifications are retrieved.
+	Callback string `yaml:"callback"`
+	callback url.URL
```
hdonnay, line 42: Why isn't this a pointer? I think it might be a holdover from the old locked version of this struct. Turning this into a pointer makes the code in `Validate` simpler.

notifier/stomp/directdeliverer.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	for bs, be := 0, rollup; bs < len(d.n); bs, be = be, be+rollup {
+		// if block-end exceeds array bounds, slice block underflow.
+		// next block-start will cause loop to exit.
+		if be > len(d.n) {
+			be = len(d.n)
+		}
+
+		n := d.n[bs:be]
+		b, err := json.Marshal(&n)
```
hdonnay: This will generate a lot of separate slices. I think it will be easier on the gc to reuse a bytes.Buffer and json.Encoder for each iteration of the loop, e.g.:

```
buf := &bytes.Buffer{}
enc := json.NewEncoder(buf)
for ... {
    ...
    buf.Reset()
    if err := enc.Encode(&n); err != nil {
        ...
    }
    ...
    tx.Send(..., buf.Bytes(), ...)
}
```

This would result in just two big buffers getting used, rather than 1 + number of iterations.

Its a real shame there's not an API for the stdlib json that doesn't involve buffering the whole data structure internally.

notifier/vulnsummary.go
hdonnay: Because this forces a copy of the struct, just to copy everything out of it.

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,144 @@
...
+func (d *Deliverer) sign(ctx context.Context, req *http.Request, kp keymanager.KeyPair) error {
+	signer, err := jose.NewSigner(jose.SigningKey{Algorithm: jose.PS512, Key: kp.Private}, (&jose.SignerOptions{}).WithType("JWT"))
+	if err != nil {
+		return fmt.Errorf("failed to create jwt signer: %v", err)
+	}
+	now := jwt.NumericDate(time.Now().Unix())
+	expire := jwt.NumericDate(time.Now().Add(1 * time.Hour).Unix())
+	cl := jwt.Claims{
+		Subject:  "clair-notifier",
+		Issuer:   "clair-notifier",
```
hdonnay: I think this clashes with the hardcoded path in the notifier. This would result in requests for `/services/clair-notifier/keys/{id}`

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,144 @@
...
+		Expiry:   &expire,
+		IssuedAt: &now,
+	}
+	kIDClaim := struct {
```
hdonnay: This needs to go into the `(jwt.JsonWebToken).Headers`, not the claims. It's very possible I implemented the auth incorrectly over in that part of the code.

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,144 @@
...
+		Msg("dispatching webhook")
+
+	resp, err := d.c.Do(req)
+	if resp != nil {
+		defer resp.Body.Close()
+	}
+	if err != nil {
+		return &clairerror.ErrDeliveryFailed{E: err}
+	}
+	if resp.StatusCode != http.StatusOK {
```
hdonnay, line 131: The server will return OK and not Created or Accepted?

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,144 @@
...
+		Kid string `json:"kid"`
+	}{
+		Kid: kp.ID.String(),
+	}
+	token, err := jwt.Signed(signer).Claims(cl).Claims(kIDClaim).FullSerialize()
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("kid", kp.ID.String())
```
hdonnay: The keyserver protocol doesn't mandate this as an HTTP header.

notifier/webhook/deliverer_test.go
```diff
@@ -0,0 +1,163 @@
...
+		t.Fatalf("failed to create request: %v", err)
+	}
+
+	ctx := log.TestLogger(context.Background(), t)
+	err = d.sign(ctx, req, kp)
+	if err != nil {
+		t.Fatalf("failed to sign request: %v", err)
+	}
+
+	kid := req.Header.Get("kid")
```
hdonnay: The spec says nothing about this HTTP header

hdonnay (changes requested):

notifier/webhook/deliverer.go
```diff
@@ -0,0 +1,144 @@
...
+func (d *Deliverer) Name() string {
+	return "webhook"
+}
+
+// sign will use the provided private key to sign and attach a jwt to the provided
+// request.
+//
+// a kid header specifying the private key which signed the jwt is attached as well.
+func (d *Deliverer) sign(ctx context.Context, req *http.Request, kp keymanager.KeyPair) error {
+	signer, err := jose.NewSigner(jose.SigningKey{Algorithm: jose.PS512, Key: kp.Private}, (&jose.SignerOptions{}).WithType("JWT"))
```
hdonnay: The keyid needs to be set here

notifier/keymanager/manager.go
ldelossa: I do not want this method at the package level, this is being used simply as a namespace.

notifier/keymanager/manager.go
ldelossa: fair. I think at first I had <=.

notifier/processor.go
ldelossa: @hdonnay @Allda can we talk about this change a bit?

Ideally, we do not want to perform the request for a very large diff, as we'd be asking the matcher to do all the work and return a large response just to say we won't process the notification.

Ideally this change happens at the matcher, and the matcher can return a `413 Payload Too Large` saying the diff is too large to process. However this still has the matcher do a least a "count" query. Maybe this is not a bad thing? 

You see that the safest thing to do is simply not process an update diff if it's the only one. I'm not held doing this, but lets discuss best way to do it moving forward.

httptransport/keybyidhandler.go
ldelossa: @hdonnay this is the default gofmt ordering, if I change the ordering, my editor snaps it right back. Also any external contributions will not perform this ordering. 

if gofmt support a config file that we can drop at root, I would say we can enforce this, but for now I won't enforce.

notifier/keymanager/manager.go
ldelossa: any reason?

notifier/keymanager/manager.go
ldelossa: The key has expired, if we want to promote the rotation of keys after expiration, we should create a new one.

notifier/postgres/keystore.go
ldelossa: Yup you are right, this was a stay-over from when the Key structure did not match what the tmp structure was being used for.

notifier/stomp/config.go
ldelossa: the url struct is rather small, why not remove the possibility of it winding up on the GC sweep by copying it around?

notifier/webhook/deliverer.go
ldelossa: @hdonnay I have not worked with a webhook system where the server receiving a webhook does not respond with 200OK.

notifier/webhook/deliverer.go
ldelossa: I'm not seeing how issuer field clashes with anything else in the app? Isn't the only thing concerned with this value is the receiver of the JWT. Does "issuer" get used for anything other then the receiving code's verification logic.

notifier/keymanager/manager.go
ldelossa: The key has expired, if we want to promote the rotation of keys after expiration, we should create a new one.

notifier/keymanager/manager.go
ldelossa: fixed

httptransport/keybyidhandler.go
ldelossa: my source file does not have this space, may have been formatted away. lets see if it's still there after this next round of fixes get pushed.

notifier/migrations/migration1.go
ldelossa: ran SQL thru a formatter.

notifier/migrations/migration1.go
ldelossa: fixed

notifier/postgres/keystore.go
ldelossa: Oh wait, I see why this is now. I want to process the tmp values into keys after the rows are closed, and not in the rows.Next() loop. Either way I'd be creating some structure to hold both the Key and the der bytes until I finished the full rows.Next() loop.

notifier/stomp/directdeliverer.go
ldelossa: fixed

notifier/postgres/keystore.go
ldelossa: fixed.

notifier/webhook/deliverer.go
ldelossa: fixed

notifier/webhook/deliverer.go
ldelossa: fixed

notifier/webhook/deliverer_test.go
ldelossa: fixed

notifier/webhook/deliverer.go
ldelossa: fixed

notifier/keymanager/manager.go
hdonnay: It's the idiom. Writing the declaration out the long way makes me think something weird is happening, but I'm pretty sure there's not.

ldelossa: I think this is fine, its valid syntax, is explicit about what it's doing, and not ambiguous.

notifier/webhook/deliverer.go
hdonnay: because the service name is hardcoded to just "notifier" in the path, but this would cause the verifier to look up "clair-notifier"

hdonnay (changes requested):
lgtm with the webhook jwt issuer changed to match the service name exported on the keyserver

hdonnay (approved):
LGTM

PR #1026, opened 2020-08-06
Fix advisory date check in Red Hat oval plugin

Allda:
There was a bug that ignores advisories released in the same day.

### DIFF
--- a/ext/vulnsrc/redhat/redhat.go
+++ b/ext/vulnsrc/redhat/redhat.go
@@ -164,9 +164,9 @@ func (u *updater) Update(datastore database.Datastore) (resp vulnsrc.UpdateRespo
 
 			// since the current document was determined to be new/updated, log a summary of what was collected
 			log.WithFields(log.Fields{
-				"collectedVulns":     len(currentCollectedVulnerabilities),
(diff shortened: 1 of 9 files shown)


ext/vulnsrc/redhat/redhat.go
```diff
@@ -702,11 +702,11 @@ func DbLookupIsAdvisoryProcessed(definition ParsedAdvisory, datastore database.D
 		// error during db lookup, treat advisory as unprocessed
 		return false
 	}
-	if len(foundVulns) > 0 {
-		// found a record, so this has already been processed
-		return true
+	for _, vuln := range foundVulns {
+		if vuln.Valid {
+			return true
```
johnbe11, line 707: @Allda - based on our conversation in chat - should we only return true if the entire set of fetched vulns matches (as opposed to returning true if any element matches as is the case here), since a given vuln can appear in multiple docs?

ldelossa (approved):
LGTM

ext/vulnsrc/redhat/redhat.go
Allda: This was already answered by PST folks internally.

PR #1030, opened 2020-08-21
Notifier test mode

ldelossa:
This PR provides several commits to implement an easy way to test the notifier subsystem.

### DIFF
--- a/notifier/service/service.go
+++ b/notifier/service/service.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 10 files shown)


ldelossa:
cc @josephschorr @Allda when this merges a much simpler way to test notifier will be available.

notifier/service/testmode.go
```diff
@@ -0,0 +1,113 @@
...
+			return claircore.NewAffectedManifests(), nil
+		}
+
+		data := make([]byte, 10)
+		_, err := rand.Read(data)
+		if err != nil {
+			return claircore.AffectedManifests{}, err
+		}
+		manifestHash := "sha256:" + fmt.Sprintf("%x", sha256.Sum256(data))
+		claircore.MustParseDigest(manifestHash) // panic if we can't parse
```
hdonnay: This would be simpler by just reading `sha256.Size` worth of bytes and then calling `claircore.NewDigest`

notifier/service/service.go
```diff
@@ -101,6 +103,13 @@ func New(ctx context.Context, opts Opts) (*service, error) {
 		return nil, fmt.Errorf("failed to initialize key manager: %v", err)
 	}
 
+	// check for test mode
+	tm := os.Getenv("NOTIFIER_TEST_MODE")
+	if strings.ToLower(tm) == "true" {
```
hdonnay:
```suggestion
	if os.Getenv("NOTIFIER_TEST_MODE") != "" {
```

If this is just a bool it shouldn't care what the value is, just that it's set.

hdonnay (dismissed):
Just some simplicity nits

hdonnay (approved):
LGTM

PR #1031, opened 2020-08-24
Airgap

hdonnay:


Documentation/operation.md
```diff
@@ -104,3 +104,127 @@ auth:
...
+### Configuration
+
+#### Airgap
+
+```yaml
+indexer:
+  airgap: true
+```
+
+#### Specific Scanners
```
ldelossa, line 216: what is this block for?

config/config.go
```diff
@@ -55,6 +58,51 @@ type Config struct {
 	Auth     Auth     `yaml:"auth" json:"auth"`
 	Trace    Trace    `yaml:"trace" json:"trace"`
 	Metrics  Metrics  `yaml:"metrics" json:"metrics"`
+	Updaters Updaters `yaml:"updaters" json:"updaters"`
+}
+
+// Updaters configures updater behavior.
+type Updaters struct {
```
ldelossa, line 65: do we need to update the configs used in our repos now?

There is a sample config at the root of the repo,
./local-dev/clair/config.yaml is used in local dev.

ldelossa (changes requested):

ldelossa:
only blocking on whether sample and local dev configs need updating now, if not looks good to me.

Documentation/operation.md
hdonnay: To have some documentation of how configuring scanners works.

I didn't sit down to write prose to flesh out these headings, though.

hdonnay:
Will double-check that and ping you

Documentation/operation.md
ldelossa: didn't realize config for scanners was being implemented. 

do we use that yet?

ldelossa (approved):

Documentation/operation.md
hdonnay: It's wired up, now. In theory rhel uses it, but it's not end-to-end tested.

PR #1032, opened 2020-08-24
Fix JSON Struct Tags for Notifier Config

alecmerdler:
### Description

Adds missing `json` struct tags and fixes existing ones. Needed for Quay Operator to deploy Clair v4 notifications.


ldelossa (approved):
LGTM

PR #1034, opened 2020-09-02
Default updaters

Allda:
Init default updaters

By importing the module default updaters are registered.


hdonnay (approved):
LGTM, but I added a commit. So @ldelossa can I get a second approval?

ldelossa (approved):

PR #1041, opened 2020-09-15
logging: add similar logging to v3

mcoops:
Add back in HTTP logging to what was present in v3, to each of the handlers.

Matched the existing format/fields from v3 here:
https://github.com/quay/clair/blob/1d1417737209cfe2e1d0b870658c8c6767832905/api/v3/server.go#L62 

Didn't add tests as there isn't really much to test.

### DIFF
--- a/httptransport/logginghandler.go
+++ b/httptransport/logginghandler.go
@@ -1,0 +1,34 @@
+package httptransport
+
+import (
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+type httpStatusWriter struct {
+	http.ResponseWriter
+
+	StatusCode int
+}
+
+// LoggingHandler will log HTTP requests using the pre initialized zerolog.
+func LoggingHandler(next http.Handler) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+
+		log := zerolog.Ctx(r.Context())
+
+		next.ServeHTTP(w, r)
+
+		log.Info().
+			Str("remote addr", r.RemoteAddr).
+			Str("method", r.Method).
+			Str("request uri", r.RequestURI).
+			Str("status", strconv.Itoa(http.StatusOK)).
+			Float64("elapsed time (md)", float64(time.Since(start).Nanoseconds())*1e-6).
+			Msg("handled HTTP request")
+	}
+}
--- a/httptransport/logginghandler.go
+++ b/httptransport/logginghandler.go
@@ -2,7 +2,6 @@ package httptransport
 
 import (
 	"net/http"
-	"strconv"
 	"time"
 
 	"github.com/rs/zerolog"
@@ -11,6 +10,11 @@ import (
 type httpStatusWriter struct {
 	http.ResponseWriter
 	StatusCode int
+}
+
+func (lrw *httpStatusWriter) WriteHeader(code int) {
+	lrw.StatusCode = code
+	lrw.ResponseWriter.WriteHeader(code)
 }
 
 // LoggingHandler will log HTTP requests using the pre initialized zerolog.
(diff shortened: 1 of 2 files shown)


httptransport/logginghandler.go
```diff
@@ -0,0 +1,34 @@
...
+		log := zerolog.Ctx(r.Context())
+
+		next.ServeHTTP(w, r)
+
+		log.Info().
+			Str("remote addr", r.RemoteAddr).
+			Str("method", r.Method).
+			Str("request uri", r.RequestURI).
+			Str("status", strconv.Itoa(http.StatusOK)).
+			Float64("elapsed time (md)", float64(time.Since(start).Nanoseconds())*1e-6).
```
ldelossa, line 37: small nit; dividing by the constants in time package maybe a little cleaner.

```
(time.Since(start).Nanoseconds()/int64(time.Millisecond))
```

ldelossa (dismissed):
Looks good, small nit.

httptransport/logginghandler.go
ldelossa: Also, does this need to be a float64 at all? Looks like you get an Int with Nanoseconds.

httptransport/logginghandler.go
```diff
@@ -0,0 +1,34 @@
...
+
+		log := zerolog.Ctx(r.Context())
+
+		next.ServeHTTP(w, r)
+
+		log.Info().
+			Str("remote addr", r.RemoteAddr).
+			Str("method", r.Method).
+			Str("request uri", r.RequestURI).
+			Str("status", strconv.Itoa(http.StatusOK)).
```
Allda: Does this always log 200?

httptransport/logginghandler.go
mcoops: Makes sense, only thing here is you might loose that visibility moving to an int64, i.e. you wont see 0.013 instead will be just 0 milliseconds. But def cleaner and easier to follow.

httptransport/logginghandler.go
mcoops: That's embarrassing *runs* I hate how that works. Readding

httptransport/logginghandler.go
ldelossa: Ah, actually missed the second float64 wrap. I see now. Your call.

mcoops:
I assume we squash on merge? Otherwise can rebase.

httptransport/logginghandler.go
```diff
@@ -0,0 +1,36 @@
...
+		// default HTTP StatusOK
+		lrw := &httpStatusWriter{ResponseWriter: w, StatusCode: http.StatusOK}
+
+		next.ServeHTTP(w, r)
+
+		log.Info().
+			Str("remote addr", r.RemoteAddr).
+			Str("method", r.Method).
+			Str("request uri", r.RequestURI).
+			Str("status", strconv.Itoa(lrw.StatusCode)).
```
ldelossa: Why the need to call Itoa here? can you avoid the method call with just 

Int("status", lrw.StatusCode) ?

httptransport/logginghandler.go
```diff
@@ -0,0 +1,36 @@
...
+
+// LoggingHandler will log HTTP requests using the pre initialized zerolog.
+func LoggingHandler(next http.Handler) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+
+		log := zerolog.Ctx(r.Context())
+
+		// default HTTP StatusOK
+		lrw := &httpStatusWriter{ResponseWriter: w, StatusCode: http.StatusOK}
```
ldelossa, line 28: I don't think I understand how this works, what updates lrw.StatusCode with a non http.StatusOK ?

ldelossa (changes requested):

httptransport/logginghandler.go
mcoops: Not sure why v3 ever did this tbh, removed.

ldelossa (approved):
Looks good to me. I suggest running Clair in local dev or on prem to ensure the logs work correctly.

PR #1042, opened 2020-09-17
auth: psk fixup

ldelossa:
this commit no longer requires a base64 PSK in the config, decoding it
became an issue when interoping with Quay.

commit also simplifies the PSK check to look at multiple issuers. This
is mostly done for more accurate logging.

### DIFF
--- a/Documentation/reference/config.md
+++ b/Documentation/reference/config.md
@@ -513,11 +513,11 @@ a string value
 A shared key distributed between all parties signing and verifying JWTs.
 ```
 
-#### &emsp;&emsp;issuer: ""
+#### &emsp;&emsp;issuer: []string
 ```
-a string value
+a list of string value
 
-The "Issuer" key is what the service expects to verify as the "issuer" claim.
+A list of issuers to verify. An empty list will accept any issuer in a jwt claim.
 ```
 
 ### &emsp;keyserver: \<object\>
--- a/middleware/auth/httpauth_psk.go
+++ b/middleware/auth/httpauth_psk.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/rs/zerolog"
 	"gopkg.in/square/go-jose.v2/jwt"
 )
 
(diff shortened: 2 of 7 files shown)


middleware/auth/httpauth_psk.go
```diff
@@ -27,22 +28,42 @@ func NewPSK(key []byte, issuer string) (*PSK, error) {
...
+
+	for i, iss := range p.iss {
+		if iss == cl.Issuer {
+			break
+		}
+		if i == len(p.iss)-1 {
+			log.Debug().Err(err).Str("iss", cl.Issuer).Msg("could not verify issuer")
+			return false
+		}
+	}
```
hdonnay, line 68: This should go after the Validate call. As a general security principal, the payload shouldn't be used before it's trusted (in this case, validated).

I also think this should fail-closed, e.g.
```suggestion
	ok := false
	for _, iss := range p.iss {
		if iss == cl.Issuer {
			ok = true
			break
		}
	}
	if !ok {
		log.Debug().Err(err).Str("iss", cl.Issuer).Msg("could not verify issuer")
		return false
	}
```

Documentation/reference/config.md
```diff
@@ -513,11 +513,11 @@ a string value
...
 ```
 
-#### &emsp;&emsp;issuer: ""
+#### &emsp;&emsp;issuer: []string
 ```
-a string value
+a list of string value
 
-The "Issuer" key is what the service expects to verify as the "issuer" claim.
+A list of issuers to verify. An empty list will accept any issuer in a jwt claim.
```
hdonnay, line 520: I think it should fail-closed. The relevant section of config can be commented out if jwt auth isn't desired.

hdonnay (changes requested):

Documentation/reference/config.md
ldelossa: Per the spec, iss optional:

```
  The "iss" (issuer) claim identifies the principal that issued the
   JWT.  The processing of this claim is generally application specific.
   The "iss" value is a case-sensitive string containing a StringOrURI
   value.  Use of this claim is OPTIONAL.
```

there can be use cases where a client may have ephemeral services coming up and down, if they create their own issuer with a hostname or some other ephemeral node value, forcing an issuer would break this situation.

hdonnay (approved):
sgtm

PR #1043, opened 2020-09-18
deployment: use service prefix for simplified path routing

ldelossa:
This commit adds a service prefix to api paths. Prior to this commit
adding a new end point to ClairV4 would require operations to
reconfigure their infrastructure.


ldelossa:
cc @alecmerdler - will need api changes in Quay.

hdonnay (approved):
Still not happy about this change, but it does make operators' configs much simpler.

ldelossa:
@hdonnay continue to brain storm on ways we can achieve both goals - nothing is set in stone just yet.

PR #1047, opened 2020-09-23
client: fix nil check

hdonnay:
These nil checks were checking an http.Response's Body member instead of
the Response pointer, causing panics whenever the Response pointer was
nil.


ldelossa (approved):
LGTM.

PR #1050, opened 2020-09-23
cicd: change log generation and releases

ldelossa:
this commit adds change log generation and a set of gihub workflows to
support it.


hdonnay (approved):
sgtm

PR #1051, opened 2020-09-23
notifier: remove first update constraint

ldelossa:
Claircore will limit the number of returned update diffs now. The check
to determine if the first update operation is encountered is no longer
needed.


Allda (approved):
LGTM

hdonnay (approved):

PR #1055, opened 2020-09-28
Use NVD 1.1 feed

byteborg:
NVD 1.0 feed is no longer available which seems to break classification of finding criticality.
Cf. https://github.com/quay/clair/pull/1021 which this is a small subset of.


byteborg:
Will post again with signoff.

PR #1073, opened 2020-09-29
config: allow disabling notifier

hdonnay:
This commit allows for the notifier to be disabled in combo mode by
removing all configuration stanzas. A log line is printed to notify the
operator, although this shouldn't affect operation, as nothing was
configured to happen in this case.

The behavior of notifier mode is slightly altered: it will now refuse to
start if unconfigured.


ldelossa (commented):
Can you confirm this works in both combo and non combo modes

ldelossa (commented):
Why do we treat notifier special here in combo mode? Should we make other modes optional as well?

alecmerdler:
What about having incremental backoff of delivery failures instead?

ldelossa:
@hdonnay  Whats that status of this?

PR #1074, opened 2020-09-30
cicd: api reference check

ldelossa:
commit adds a check to confirm api reference does not go stale


hdonnay (approved):
LGTM

PR #1086, opened 2020-10-07
notifier: ensure Content-Type header present in webhook notification

alecmerdler:
When receiving the notification webhook in Quay, we call Flask's `.json()` method. This returns `None` if the `Content-Type` header is not set to `application/json`.

### DIFF
--- a/notifier/webhook/config.go
+++ b/notifier/webhook/config.go
@@ -45,5 +45,11 @@ func (c *Config) Validate() (Config, error) {
 		return conf, fmt.Errorf("failed to parse callback url")
 	}
 	conf.callback = callback
+
+	if conf.Headers == nil {
+		conf.Headers = map[string][]string{}
+	}
+	conf.Headers.Set("Content-Type", "application/json")
+
 	return conf, nil
 }


notifier/webhook/deliverer.go
```diff
@@ -96,9 +96,15 @@ func (d *Deliverer) Deliver(ctx context.Context, nID uuid.UUID) error {
 	}
 	buf := bytes.NewReader(b)
 
+	var headers http.Header = map[string][]string{}
+	headers["Content-Type"] = []string{"application/json"}
+	for key, val := range d.conf.Headers {
+		headers[key] = val
+	}
+
```
hdonnay: I think this would be better written as:

```suggestion
	headers := d.conf.Headers.Clone()
	headers.Set("Content-Type", "application/json")

```

hdonnay (changes requested):

notifier/webhook/deliverer.go
alecmerdler: What if `Headers` is `nil`?

hdonnay: Hm yeah, that eats it because Clone returns nil. We could make Validate ensure it's non-nil, or not do it my suggested way. @ldelossa thoughts?

ldelossa: I think doing this in Validate method makes sense. 

The webhook deliverer will *always* be sending json data. So in the "Validate()" method, merge any incoming headers with the "Content-type" header and construct the deliverer with the the merged http.Header map.

ldelossa (approved):

hdonnay (approved):

PR #1088, opened 2020-10-08
local-dev: make quay container ignore validations

jan-zmeskal:
Without this value being set, clair container fails to start with the following in the logs:
```
+------------------------+-------------------------------------------+--------+
| SigningEngine          | GPG2_PRIVATE_KEY_NAME is required         | 🔴     |
+                        +-------------------------------------------+--------+
|                        | Cannot access the file signing-public.gpg | 🔴     |
+------------------------+-------------------------------------------+--------+
```
It expects that GPG keys stored at variables `GPG2_PRIVATE_KEY_FILENAME` and `GPG2_PUBLIC_KEY_FILENAME` are available, which they aren't.

Furthermore I fixed a broken link in docs that is supposed to point at a sample config file.

### DIFF
--- a/Documentation/howto/getting_started.md
+++ b/Documentation/howto/getting_started.md
@@ -35,7 +35,7 @@ This how-to will demonstrate combo mode and introduce some further reading on a 
 
 Clair uses PostgreSQL for its data persistence. Migrations are supported so you should only need to point Clair to a fresh database and have it do the setup for you.
 
-We will assume you have setup a postgres database it's reachable with the following connection string:
+We will assume you have setup a postgres database and it's reachable with the following connection string:
 `host=clair-db port=5432 user=clair dbname=clair sslmode=disable`. Adjust for your environment accordingly. 
 
 ## Starting Clair In Combo Mode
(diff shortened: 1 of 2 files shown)


Documentation/howto/getting_started.md
```diff
@@ -44,14 +44,14 @@ At this point, you should either have built Clair from source or have pulled the
 
 *You may need to configure [docker](https://docs.docker.com/network/) or [podman](https://podman.io/getting-started/network.html) networking if you are utilizing containers. This is out of scope for this how-to.*
 
-A basic config for combo mode can be found [here](https://github.com/quay/clair/blob/development-4.0/config.yaml.sample). Make sure to edit this config with your database settings and set "migrations" to `true` for all mode stanzas. In this basic combo mode, all "connstring" fields should point to the same database and any *_addr fields are simply ignored. For more details see the [config reference](../reference/config.md) and [deployment models](./deployment.md)
+A basic config for combo mode can be found [here](../../config.yaml.sample). Make sure to edit this config with your database settings and set "migrations" to `true` for all mode stanzas. In this basic combo mode, all "connstring" fields should point to the same database and any *_addr fields are simply ignored. For more details see the [config reference](../reference/config.md) and [deployment models](./deployment.md)
```
ldelossa: @jan-zmeskal I do not think this will work. We render documentation with "mdbook" and I think any referenced links must belong within the Documentation folder. 

I could be wrong about this, you can actually test if it works by running "mdbook serve" from the root of the repository. You will need to install the mdbook tool. https://github.com/rust-lang/mdBook

ldelossa (changes requested):

Documentation/howto/getting_started.md
jan-zmeskal: @ldelossa You are right. I replaced the relative path with web link pointing to `main` branch. I tried this locally by running `mdbook serve` and it worked fine.

ldelossa (approved):
Appreciated!

PR #1089, opened 2020-10-08
notifier: copy url struct

ldelossa:
previous to this commit a reference to the callback url structure was
being mutated.

this commit ensures a copy of the url structure is made and the original
reference is no longer mutated.

### DIFF
--- a/notifier/webhook/deliverer.go
+++ b/notifier/webhook/deliverer.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"path"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


hdonnay (approved):

notifier/webhook/deliverer.go
```diff
@@ -83,12 +83,12 @@ func (d *Deliverer) Deliver(ctx context.Context, nID uuid.UUID) error {
 		Str("notification_id", nID.String()).
 		Logger()
 
-	callback := d.conf.callback
+	callback := *d.conf.callback
 	callback.Path = path.Join(callback.Path, nID.String())
```
hdonnay: This is probably better, but does require the config to end with a slash:
```suggestion
	callback, err := d.conf.callback.Parse(nID.String())
	if err != nil {
		return err
	}
```

hdonnay (changes requested):

hdonnay (approved):
LEBTM (looks even better to me)

PR #1099, opened 2020-10-12
Accept OCI manifests for indexing

hdonnay:
This _seemed_ simple enough, haven't check if there are other places where we want to accept other inputs.

Still need to update the openapi reference.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,8 @@ require (
 	github.com/klauspost/compress v1.13.6
 	github.com/ldelossa/responserecorder v1.0.2-0.20210711162258-40bec93a9325
 	github.com/mattn/go-sqlite3 v1.11.0 // indirect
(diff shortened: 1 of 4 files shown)


ldelossa:
This is cool, and I like the idea. Only question is how do we want to communicate this to user? Do we simply say "make a json post with this mime type and we'll assume you are sending an oci manifest"? whats your thoughts here.

hdonnay:
Yeah, I think we'd document the acceptable input types, and control them via the `content-type` header. The default would be the current state of things (assume it's our manifest format).

ldelossa:
Cool, as a caveat we will have to reject any oci manifests where either do not contain a content descriptor, or contain a content descriptor without urls.

mieliespoor:
Any indications on when this will see progress again? Merging this will greatly improve integrations.

PR #1102, opened 2020-10-13
local-dev: add pgadmin4 container

jan-zmeskal:
The purpose here is to make clair-db introspection easier. With this change, a developer can use [pgadmin](https://www.pgadmin.org/) browser GUI to view clair-db database. Having it run in container and pre-configured removes the need for a developer to install it manually on their system and set up clair-db connection every time. With this change, one can simply:
1. `make local-dev-up`
1. Enter `localhost:80` in the browser
1. Log-in with the default user (clair@clair.com/clair)
1. Inspect the clair-db (no need for setting up the connection, it's already pre-loaded)
image

I don't know about exposing port 80 though. It's the one that is used in [pdagmin4 docs](https://www.pgadmin.org/docs/pgadmin4/development/container_deployment.html), so I chose it for compliance. At the same time, I know it's heavily contested port. What do you think?

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -24,6 +24,21 @@ services:
       retries: 3
       start_period: 10s
 
+  pgadmin:
+    container_name: pgadmin
+    image: dpage/pgadmin4:latest
+    environment:
+      PGADMIN_DEFAULT_EMAIL: clair@clair.com
+      PGADMIN_DEFAULT_PASSWORD: clair
+      PGADMIN_SERVER_JSON_FILE: /pgadmin4/config/servers.json
+      PGADMIN_LISTEN_PORT: 8081
+    ports:
+      - "8081:8081"
+    volumes:
+      - "./local-dev/pgadmin:/pgadmin4/config"
+    depends_on:
+      - clair-db
+
   traefik:
     container_name: clair-traefik
     image: traefik:v2.2
--- a/Documentation/howto/testing.md
+++ b/Documentation/howto/testing.md
@@ -50,6 +50,11 @@ localhost:5433 --- Quay's Postgres DB
                      username: quay
                      database: quay
 
+localhost:8081 --- Postgres GUI (pgadmin4)
+                   Login:
+                     username: clair@clair.com
+                     password: clair
+
 localhost:8082 --- OpenAPI Swagger Editor.
                    You can view ClairV4's public API here.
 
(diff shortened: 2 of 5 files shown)


docker-compose.yaml
```diff
@@ -24,6 +24,20 @@ services:
       retries: 3
       start_period: 10s
 
+  pgadmin:
+    container_name: pgadmin
```
ldelossa, line 28: Safe to pin this to "latest" ? Or should we be more strict?

docker-compose.yaml
```diff
@@ -24,6 +24,20 @@ services:
...
+    image: dpage/pgadmin4:latest
+    environment:
+      PGADMIN_DEFAULT_EMAIL: clair@clair.com
+      PGADMIN_DEFAULT_PASSWORD: clair
+      PGADMIN_SERVER_JSON_FILE: /pgadmin4/config/servers.json
+    ports:
+      - "80:80"
```
ldelossa: Can we move this to a less common port? Taking port 80 seems a bit brazen and requires sudo.

ldelossa (changes requested):
Nice add. Few comments in the pull but this seems like a good idea.

docker-compose.yaml
jan-zmeskal: Hard to say. I don't have long-term experience with this image as previously I had pgadmin just installed on my system. As this is a development tool and read-only at that, I don't think it's an issue and I think we can go with latest.

docker-compose.yaml
jan-zmeskal: Yeah, I also think so. What do you think would be a good port number, @ldelossa ? I was thinking maybe 8081 since we already use 8080 and 8082.

jan-zmeskal:
@ldelossa I updated pgadmin to use port 8081. Also please note that I did two additional things:
1. I noticed the command to restart rabbitmq in `Makefile` was wrong so I've fixed it 
2. I added command `local-dev-clair-restart` to restart all the instances of clair (indexer, matcher and notifier)

ldelossa (dismissed):
LGTM.

ldelossa:
@jan-zmeskal thanks for the adds. Looks good. Needs a rebase however.

jan-zmeskal:
@ldelossa Can you pls approve again? Sorry for that, I needed to rebase

ldelossa (approved):

PR #1104, opened 2020-10-14
Pid1

hdonnay:
These should remove the need for an init in the container.

An init already isn't needed when running under k8s, only when run as a standalone container.

The whole pid1 dance is linux-specific, I don't know if we'd want build tags on it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,8 +24,7 @@ RUN go build\
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal AS final
 RUN microdnf install tar
-RUN curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 && chmod +x /usr/local/bin/dumb-init 
-ENTRYPOINT ["/usr/local/bin/dumb-init", "--", "/bin/clair"]
+ENTRYPOINT ["/bin/clair"]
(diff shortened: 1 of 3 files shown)


hdonnay:
This passes a smoke test locally:

```
% podman run --rm -it clair-local:latest
pid1: becoming a dumb init
2020/10/14 18:31:10 must provide a -conf flag or set "CLAIR_CONF" in the environment
```

hdonnay:
Added a way to force previous behavior via the environment. Here's what it looks like now:

```
% podman run --rm -it -e NO_INIT=1 clair-local:latest
2020/10/14 20:17:09 must provide a -conf flag or set "CLAIR_CONF" in the environment
% podman run --rm -it clair-local:latest
pid1: becoming a dumb init
2020/10/14 20:17:17 must provide a -conf flag or set "CLAIR_CONF" in the environment
% podman run --init --rm -it clair-local:latest
2020/10/14 20:18:57 must provide a -conf flag or set "CLAIR_CONF" in the environment
```

ldelossa:
I'm not in favour or doing this in Clair binary proper for a few reasons:

Clair should not be concerned with being an init
This narrows Clair's usage to linux, and even with that the code hardcodes a syscall number which may change between kernel builds.
The comments indicate there are implications around using systemd facilities when this code is used, Clair should not include those implications. 
The init functionality should be punted to the infrastructure running Clair.

PR #1106, opened 2020-10-19
Notifier: fix poller loop

Allda:
The return was in incorrect place and cause that only first updater was
processed. Rest of updaters were skipped.

### DIFF
--- a/notifier/poller.go
+++ b/notifier/poller.go
@@ -117,12 +117,18 @@ func (p *Poller) onTick(ctx context.Context, c chan<- Event) {
 			select {
 			case c <- e:
 			default:
-				log.Warn().Str("updater", e.updater).Str("UOID", e.uo.Ref.String()).Msg("could not deliver event to channel. backing off till next tick")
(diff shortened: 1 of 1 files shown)


notifier/poller.go
```diff
@@ -118,8 +118,8 @@ func (p *Poller) onTick(ctx context.Context, c chan<- Event) {
 			case c <- e:
 			default:
 				log.Warn().Str("updater", e.updater).Str("UOID", e.uo.Ref.String()).Msg("could not deliver event to channel. backing off till next tick")
+				return
 			}
-			return
```
ldelossa, line 122: this actually needs to be replaced with a "continue". We create an event, if we cannot place into channel just log the Warning and continue the outer for loop.

so change this return line to "continue" and remove the "return" you placed at line 121. The channel maybe able to enqueue an item on next loop iteration.

ldelossa (changes requested):

ldelossa (approved):

PR #1108, opened 2020-10-19
notifier: do less work

ldelossa:
previously the notifer would attempt to create notifications for the
same update operation if no affected manifests were discovered.

with this patch we book keep the update operations we encountered and
short circuit the processing of notifications which do not produce
affected manifests.

a receipt in "delivered" status is written to the database directly for
update operations that do not produce affected manifests. this effectively
removes them from subsequent processing and avoides any delivery
attempts.

### DIFF
--- a/notifier/postgres/putreceipt.go
+++ b/notifier/postgres/putreceipt.go
@@ -1,0 +1,59 @@
+package postgres
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/quay/clair/v4/notifier"
+)
+
+func putReceipt(ctx context.Context, pool *pgxpool.Pool, updater string, r notifier.Receipt) error {
+	const (
+		insertNotification    = `INSERT INTO notification (id) VALUES ($1);`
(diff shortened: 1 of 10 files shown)


Allda (dismissed):
LGTM

notifier/postgres/putreceipt.go
```diff
@@ -0,0 +1,58 @@
...
+func putReceipt(ctx context.Context, pool *pgxpool.Pool, updater string, r notifier.Receipt) error {
+	const (
+		insertNotification    = `INSERT INTO notification (id) VALUES ($1);`
+		insertReceipt         = `INSERT INTO receipt (notification_id, uo_id, status, ts) VALUES ($1, $2, $3, CURRENT_TIMESTAMP);`
+		insertUpdateOperation = `
+		INSERT INTO notifier_update_operation (updater, uo_id, ts)
+		VALUES ($1, $2, CURRENT_TIMESTAMP)
+		`
+	)
+	tx, err := pool.Begin(ctx)
```
hdonnay, line 20: This is missing a defer to cover all the early returns

hdonnay (changes requested):

hdonnay (approved):

PR #1111, opened 2020-10-23
Initialization reorg

hdonnay:
This reorg is to support being able to use liveness/readiness probes on k8s.

Currently, the entire init process happens behind `initialize.New`. With these changes, the configuration is first-pass checked, the logging is configured, then the two HTTP servers are started in parallel using an errgroup, with any errors in that group bringing down the process.


cmd/clair/main.go
```diff
@@ -51,54 +56,70 @@ func main() {
...
-			err := init.Introspection.ListenAndServe()
-			if err != nil {
-				logger.Err(err).Msg("introspection server failed to launch. continuing anyway")
-			}
-		}()
-	}
+		i, err := introspection.New(ctx, conf, nil)
+		if err != nil {
+			logger.Warn().
+				Err(err).Msg("introspection server configuration failed. continuing anyway")
```
ldelossa, line 78: Moving forward, do we want to ensure introspection is up and not ignore it if it fails?

cmd/clair/main.go
```diff
@@ -51,54 +56,70 @@ func main() {
...
+			return
+		}
+		down.Add("introspection", i.Server)
+		if err := i.ListenAndServe(); err != http.ErrServerClosed {
```
ldelossa, line 82: You've added the server to the Shutdown structure, but it can still fail to ListenAndServe. 

Is it okay for the Shutdown structure to call a server's "Shutdown" method if that server is not running? 

If not you - make sure it does not get into the Shutdown's structure's map. If it is okay, ignore this comment.

cmd/clair/shutdown.go
```diff
@@ -0,0 +1,52 @@
...
+// Shutdown aggregates http.Sever Shutdown methods.
+type Shutdown struct {
+	mu sync.Mutex
+	m  map[string]*http.Server
+}
+
+// Add registers a server with specified name.
+//
+// Currently, it is disallowed to add two servers with the same name.
+func (s *Shutdown) Add(name string, srv *http.Server) {
```
ldelossa: You could technically make the map key the server's pointer. You do not retrieve servers, and pointers will remain unique so naming conflicts go away.

cmd/clair/shutdown.go
```diff
@@ -0,0 +1,52 @@
...
+	s.m[name] = srv
+}
+
+// If reloading support is added, a
+//	Swap(string, *http.Server) *http.Server
+// function would be useful.
+
+// Shutdown calls Shutdown on all added Servers. If a timeout is needed, it
+// should be done via the passed Context.
+func (s *Shutdown) Shutdown(ctx context.Context) error {
```
ldelossa, line 30: nice way to handle a single timeout for multiple server shutdown.

initialize/services.go
```diff
@@ -14,233 +16,217 @@ import (
...
+// provided configuration.
+func Services(ctx context.Context, cfg *config.Config) (*Srv, error) {
+	log := zerolog.Ctx(ctx).With().Str("component", "init/Services").Logger()
 	log.Info().Msg("begin service initialization")
+	defer log.Info().Msg("end service initialization")
 
-	switch i.conf.Mode {
+	var srv Srv
+	var err error
+	switch cfg.Mode {
```
ldelossa, line 54: great refactor.

ldelossa (changes requested):
Blocking on confirmation or dismal of a concern. 

Refactor looks dope.

cmd/clair/main.go
hdonnay: I'm unsure. The application can work just fine if this server doesn't come up, is my reasoning for having it like this.

cmd/clair/main.go
hdonnay: I think this is okay, but I'm going to try to prove it to myself.

cmd/clair/main.go
ldelossa: Okay we'll leave it for now and chuck it to a community dev question when we have some ppl in there.

cmd/clair/main.go
hdonnay: https://play.golang.org/p/IbPqyX2MyY-

The playground seems broken right now, but running this locally doesn't panic.

cmd/clair/shutdown.go
hdonnay: Oh, I did it that so that hot reloading could, eventually, work, but that can just quickly be reworked if it ever happens.

ldelossa (approved):

PR #1119, opened 2020-11-03
docs: Add information about AMQP delivery compatibility

jan-zmeskal:
In clair we use this library for AMQP connection: https://github.com/streadway/amqp#goals

According to its documentation (and I also ran into that wall personally), it supports only AMQP 0.9.1 protocol. Since it may not be obvious to the end user (it was not obvious to me), I though it would be good to mention this in docs so that other users might avoid this pitfall.


hdonnay (dismissed):

jan-zmeskal:
@hdonnay I rebased on top of main, pls re-review

ldelossa (approved):

PR #1121, opened 2020-11-10
config: add 'omitempty' to 'updaters' config struct for correct marshalling

alecmerdler:
Clair treats the `updaters.sets` field differently when set to `nil` or `[]string{}`. Adds the `omitempty` struct tag to ensure when marshalling the config the empty `updaters` section is omitted rather than zeroed out.

Fixes https://issues.redhat.com/browse/PROJQUAY-1268


hdonnay (approved):
I guess this is fine to add, with the proviso that it still doesn't allow a user to create all the valid configs because of the vagaries of the gopkg.in/yaml marshaling.

If that's a desirable property, we should stop using yaml.

PR #1124, opened 2020-11-19
Notifier: Fix bugs that ignore some notification

Allda:
This PR fixes 2 bugs that caused a subset of notifications was completely ignored.

1)
Fix slices in the direct notifier
There was a bug in the direct notifier that caused only a subset of
notifications was sent to the client. The first iteration over notification
replaced the original notification slice with just the first "page". This caused
that only one "page" was sent and the rest was ignored.

This commit fixes amqp and stomp direct notification delivery.

2)
Create a notification for each vulnerability

The previous implementation sent only a subset of vulnerabilities which
can cause confusion on client side. The current implementation sends
notifications for each combination of vulnerability and affected image.

### DIFF
--- a/notifier/amqp/directdeliverer.go
+++ b/notifier/amqp/directdeliverer.go
@@ -85,6 +85,7 @@ func (d *DirectDeliverer) Deliver(ctx context.Context, _ uuid.UUID) error {
 
 	var buf bytes.Buffer
 	enc := json.NewEncoder(&buf)
+	var currentBlock []notifier.Notification
 	for bs, be := 0, rollup; bs < len(d.n); bs, be = be, be+rollup {
 		buf.Reset()
 		// if block-end exceeds array bounds, slice block underflow.
@@ -93,8 +94,8 @@ func (d *DirectDeliverer) Deliver(ctx context.Context, _ uuid.UUID) error {
 			be = len(d.n)
 		}
 
-		d.n = d.n[bs:be]
-		err := enc.Encode(&d.n)
+		currentBlock = d.n[bs:be]
+		err := enc.Encode(&currentBlock)
 		if err != nil {
 			ch.TxRollback()
 			return &clairerror.ErrDeliveryFailed{err}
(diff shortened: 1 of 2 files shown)


ldelossa (changes requested):
The changes you want to make in notifier/processor.go does not align with how Quay utilizes notifications. 

When the notifier detects a new update and affected manifests it will filter and create a notification for only the most severe advisory encountered. 

When the notification is delivered, it is expected the client requests a new vulnerability report providing all the details of the update.

notifier/amqp/directdeliverer.go
```diff
@@ -93,8 +93,8 @@ func (d *DirectDeliverer) Deliver(ctx context.Context, _ uuid.UUID) error {
 			be = len(d.n)
 		}
 
-		d.n = d.n[bs:be]
-		err := enc.Encode(&d.n)
+		notificationSlice := d.n[bs:be]
```
ldelossa: The idea behind the first, but buggy, way to do this was to avoid allocating a new slice variable over and over...

I think the same can be accomplished, without the bug, like this... 

```
	var buf bytes.Buffer
	enc := json.NewEncoder(&buf)
	var currentBlock []notifier.Notification
	for bs, be := 0, rollup; bs < len(d.n); bs, be = be, be+rollup {
		buf.Reset()
		// if block-end exceeds array bounds, slice block underflow.
		// next block-start will cause loop to exit.
		if be > len(d.n) {
			be = len(d.n)
		}

		currentBlock = d.n[bs:be]
		err := enc.Encode(&currentBlock)
		if err != nil {
			ch.TxRollback()
			return &clairerror.ErrDeliveryFailed{err}
		}

```

in this case, we make one slice var 'currentBlock' and continually reuse it.

ldelossa (changes requested):

ldelossa (dismissed):
thanks for fix!

ldelossa (approved):

PR #1126, opened 2020-11-19
Clairctl fixes

hdonnay:
These are some issues found by Red Hat QE when poking at clairctl.


ldelossa (dismissed):
this LGTM. 

I do wonder if it makes sense to serve the config to clairctl from Clair, instead of requiring the user to have the config local, but that's more of an RFE to this then a review comment.

ldelossa:
@hdonnay confirmed this working.

ldelossa (approved):

PR #1130, opened 2020-11-27
Dockerfile: Get build image from Quay instead of DockerHub

jan-zmeskal:
quay.io/repository/app-sre/golang is supposed to mirror docker.io/golang. Using it will enable us to work around DockerHub's newly introduced pull limits.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM docker.io/library/golang:1.15 AS build
+FROM quay.io/app-sre/golang:1.15 AS build
 WORKDIR /build/
 ADD . /build/
 ARG CLAIR_VERSION=dev


ldelossa:
This is not correct. We create a golang dockerfile with some dependencies. See https://github.com/quay/claircore/blob/master/etc/Dockerfile

jan-zmeskal:
Hi @ldelossa thank you for looking into that so quickly, but keep in mind that this was only a draft PR intended as a placeholder while I'm figuring this out. But it doesn't matter, at least the conversation got started. :-)

Anyway, could you help me understand what's the relationship between https://github.com/quay/claircore/blob/master/etc/Dockerfile and https://hub.docker.com/_/golang? Becase it seems to me that is is what main Dockerfile of this repository seems to be using.

ldelossa:
@jan-zmeskal 

Typically, you can open an issue for discussion on topics like this. 

The dockerfile linked to is ubi8, which has rpm tool required for Clair to work correctly. Default golang container is based on ubuntu which will not. 

We create a custom base container for a reason, and that should not be swapped out, the base in /etc has the dependencies used to run Clair, the default golang base does not. 

There is no need to change the container, hence why the pr is closed. If youd like to discuss further open an issue.

ldelossa:
@jan-zmeskal my apologies, I misread this PR. I had thought we were changing usage of our base docker image: 
https://quay.io/repository/claircore/golang?tab=tags

It appears you just want to swap out the docker builder image with one that app-sre hosts.

jan-zmeskal:
@ldelossa Yes, that's right. I believe this change should not impact anything, but maybe I'm missing something obvious. Also I tried deploying clair from my source branch to our dev env and everything seems to be working fine.

ldelossa:
@jan-zmeskal 

Is the app-sre repository a direct mirror of golang's official. Will there be any lag between golang releases and app-sre repository. 

I know little about this repo.

jan-zmeskal:
@ldelossa When checking the tags in both repositories, it seems that the Quay one is not really mirroring the DokcerHub one. Some tags have been pushed to Quay even earlier than to DockerHub, example here:
quay_docker_compare

The `app-sre` organization on Quay seems to actively maintain a wide range of images, see here: https://quay.io/organization/app-sre I'll try to contact the people behind it and ask them about the relation between their golang image and the one hosted on DockerHub.

ldelossa:
@jan-zmeskal sounds good. 

My only concern with this is moving away from the officially supported Golang containers which ensure availability as soon as new Go versions are released.

jan-zmeskal:
@ldelossa Completely understandable. Let me get in touch with folks maintaining the repo and find out what kind of SLA is in place there.

ldelossa (approved):
Conversation with @jan-zmeskal and @hdonnay out of band resulting in an approval of this change.

PR #1138, opened 2020-12-03
cicd: use CI golang image from quay.io

jan-zmeskal:
Another PR aiming to get us rid of DockerHub dependency

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,7 +66,7 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
-    container: docker.io/library/golang:${{ matrix.go }}-alpine
+    container: quay.io/app-sre/golang:${{ matrix.go }}-alpine
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -66,7 +66,7 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
-    container: docker.io/library/golang:${{ matrix.go }}-alpine
+    container: quay.io/app-sre/golang:${{ matrix.go }}-alpine
```
ldelossa, line 69: @jan-zmeskal 

just a sanity check, each image in this matrix exists in the app-sre repo correct?

jan-zmeskal: Yes, last Friday (December 4) I checked that all the tags in the matrix are present in https://quay.io/repository/app-sre/golang. As far as I can tell, all the tags are mirrored expect those containing "windows" and "nanoserver" in their name.

ldelossa (approved):

Issue #1156, opened 2021-01-06
Clair's path towards DockerHub independence


jan-zmeskal:
### Introduction

DockerHub introduced [pull rate limits](https://www.docker.com/increase-rate-limits) in November 2020, which is especially hurtful for CI workflows. If they run many times a day, they might easily hit it. Therefore I propose that we change our CI so that it does not rely on docker.io/golang image any more. 

### Implementation
After discussing this topic with Louis and Hank, we came to the conclusion that we can build our own golang image. We already do that for local development anyway. Of course, if custom golang image is to become the basis of our official clair image, it needs to be well maintained by means of automation. We discussed how to do that and settled on regularly scheduled GitHub action that checks golang version and build and build a new custom golang image if new golang version has been released since our last build.

### Tasks
- [x] As of now, our custom golang image resides in quay.io/organization/claircore, we should move it to quay.io/organization/projectquay. All code that depends on quay.io/claircore/golang must be changed accordingly.
- [x] Have an authoritative list of major golang releases that we test with (as of time of me writing this, it's 1.13, 1.14, 1.15).
- [x] Set up regular GH action that tracks golang releases, possibly by means of GH API tracking golang tags as suggested by Hank. Other option might be tracking changes in DockerHub golang repository. But that seems unnecessarily indirect.
- [x] GH action persists information about latest golang releases, possibly in artifact as suggested by Louis
- [x] The regular GH action checks if our golang image's tags are up to date with golang releases. If not, it builds and pushes tags that are behind.
- [x] All code that relies on docker.io/golang is changed so that it uses quay.io/projectquay/golang instead. I created https://github.com/quay/claircore/pull/284 and https://github.com/quay/clair/pull/1147 before this plan which might be used for reference.
- [x] We need to figure out whether we can do this with one Dockerfile (I think we can, because it can be provided build args) or we need to template several Dockerfiles
- [x] We need to figure out whether a separate directory should be set up for this or whether this automation can live in this repo

### Notes:
- In case we need to persists some information between the runs, `actions/download-artifact@v2` can't really be used for that purpose. However, we might be able to fetch previously build artifacts using [REST API](https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#artifacts).

jan-zmeskal:
@ldelossa @hdonnay Please review my proposal and check that I haven't forgotten anything major. Also, please let's treat this issue as the main discussion platform for this effort from now on.

ldelossa:
[...]

PR #1159, opened 2021-01-18
cicd: add publish-golang-image workflow

jan-zmeskal:
Intent of this PR is to deal with most of the problems outlined in #1156. Please see the issue for in-depth analysis of why we need this. 

## Implementation
I've has some really good discussions about this with Louis and Hank in #clair-collaboration and this is the outcome. The new workflow will run every day, at 5AM UTC. Here's the basic logic:
```
Find out what is the latest released minor version of golang
(Example: For go1.15 it'll be 1.15.7)
|
v
Find out when respective tag (golang major version) of quay.io/projectquay/golang was last modified
(Example: quay.io/projectquay/golang:1.15 was last modified on January 1 2021)
|
v
Find out when the released minor version of golang was last modified
(Example: https://dl.google.com/go/go1.15.7.linux-amd64.tar.gz was last modified on January 15 2021)
|
v
Check if released golang is newer than the quay.io tag
|  |
|  L> If not, no need to do anything
|
v
If yes (case of our example), continue
|
v
Build new image and tag it quay.io/projectquay/golang:1.15
|
v
Push the newly built image
```

## Missing stuff
1. There's one thing I could not figure out and that is how to get sha256 checksum of golang archive programatically. Why do we even need this? The Dockerfile used to build our golang image [uses it to verify](https://github.com/quay/claircore/blob/fc1aa30841820e309653733db660c56deae033f8/etc/Dockerfile#L12) that the downloaded archive wasn't tampered with. This workflow works fine even without providing such checksum, but we have this in the GH Actions log:
image
So couple of points here:
- Do we really need this check?
- Do you guys have any idea how to fetch the checksum in the automation?
- I mean I could parse the HTML code of the download page but that seems like a [bad idea](https://stackoverflow.com/a/1732454/6758779)
2. For this initiative to be finalized, we need to make sure that our CI here and in quay/claircore consumes quay.io/projectquay/golang instead of quay.io/claircore/golang. I suggest to do it afterwards in separate, small PRs once we are sure this is running regularly and reliably.

### DIFF
--- a/.github/scripts/get-latest-golang-minor-version.sh
+++ b/.github/scripts/get-latest-golang-minor-version.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+MAJOR_VERSION=$1
+LOGFILE=$2
+
+LATEST_MINOR_VERSION=0
+
+while true
+do
+    let CHECK_MINOR_VERSION=LATEST_MINOR_VERSION+1
+    URL=https://dl.google.com/go/go${MAJOR_VERSION}.${CHECK_MINOR_VERSION}.linux-amd64.tar.gz
+    echo -n "${URL} " >> ${LOGFILE}
+    HTTP_RESP_CODE=`curl -I -o /dev/null -s -w "%{http_code}\n" "${URL}"`
+    echo "${HTTP_RESP_CODE}" >> ${LOGFILE}
+    if [[ HTTP_RESP_CODE -ne 200 ]]
+    then
+        break
+    fi
+    let LATEST_MINOR_VERSION++
+done
+
+# Return latest Golang version available for download in format "1.15.6"
+echo "${MAJOR_VERSION}.${LATEST_MINOR_VERSION}"
--- a/.github/workflows/golang-image.yml
+++ b/.github/workflows/golang-image.yml
@@ -1,0 +1,90 @@
+---
+name: Build and publish golang image
+
+on:
+  # Run every day at 5AM UTC
+  schedule:
+    - cron: '0 5 * * *'
+
+env:
+  QUAY_ORG: projectquay
+  QUAY_REPO: golang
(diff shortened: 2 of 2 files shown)


jan-zmeskal:
@ldelossa @hdonnay Can you please review?

jan-zmeskal:
Refreshing request for review @ldelossa @hdonnay

jan-zmeskal:
@ldelossa @hdonnay Regular ping :)

ldelossa:
@jan-zmeskal appreciate the work on this. 

Can we move this work to the ClairCore repository? I do not feel that Clair is the correct repository for this.

jan-zmeskal:
@ldelossa Sure, I don't think there's a problem with that. Well, actually, one potential problem comes to mind. I'm making use of `secrets.QUAY_TOKEN`. I don't know if that secret is also available in claircore repo. I can't access the secrets and I can't see it being used in the code.

Also, before I proceed with this, I'd like to hear @hdonnay 's take on the first point in Missing stuff section. And also an opinion of both of you if you are okay with the general shape of the workflow.

hdonnay:
So there's a file in the go download bucket that has the checksum that's at `<file>.sha256`, e.g. `https://dl.google.com/go/go1.16.3.linux-amd64.tar.gz.sha256` There's some hints on how this is done in the `golang.org/dl/internal/version` package.

That being said, I'm not totally sure how much the checksum buys us if it's not copied out-of-band. So, I'm okay with it getting dropped.

jan-zmeskal:
@hdonnay Thank you for the review. Now I only need `secrets.QUAY_TOKEN` to be available in claircore repo. Could you please look into it?

jan-zmeskal:
Close in favor of https://github.com/quay/claircore/pull/383

PR #1161, opened 2021-01-22
HTTP Readiness

hdonnay:
This change (based on #1111) has the http server start serving requests immediately, and start serving the API after initialization is done.

### DIFF
--- a/httptransport/client/matcher.go
+++ b/httptransport/client/matcher.go
@@ -236,3 +236,9 @@ func (c *HTTP) UpdateDiff(ctx context.Context, prev, cur uuid.UUID) (*driver.Upd
 	}
 	return &d, nil
 }
+
+// Initialized is present to fulfill the interface, but isn't exposed as part of
+// the HTTP API. This method is stubbed out.
+func (c *HTTP) Initialized(_ context.Context) (bool, error) {
+	return true, nil
+}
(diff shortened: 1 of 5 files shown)


hdonnay:
Going to rework claircore's init and then circle back to this.

ldelossa:
how does this PR compare with https://github.com/quay/claircore/pull/300 

doesn't #300 obsolete this?

hdonnay:
We'll still need to do something in clair's http layers to make use of the information exposed in that change.

ldelossa:
Gotchya, was expecting a new PR, but doesn't matter if this one is used.

alecmerdler:
@hdonnay Let me know if you need help testing this. This is going to speed up TNG Operator's reconciliation of `QuayRegistry` objects so much faster (Clair updaters take 5+ min to complete, blocking Quay startup).

hdonnay:
This has been reworked and rebased in face of claircore changes.

ldelossa (changes requested):
Possible to write a integration test for this in some way? Ensuring you see the appropriate status codes change at the appropriate times?

alecmerdler:
@hdonnay Would love to see this merged. Is there any way I can help test?

ldelossa:
From my point of view, it just needs a test that confirms the http status flips correctly when initialized

hdonnay:
I'll jump on a test for the handler. To test it live, you should just need to run it with no updaters (`$.updaters.sets = []`) against an empty database.

ldelossa (dismissed):
@hdonnay approved, needs rebase, ping me when done for re-ack.

ldelossa (approved):

PR #1163, opened 2021-01-25
PROJQUAY-586 - Remove Defcon1 severity

syed:
Defcon1 is a manual severity only used by the quay team. It is being depreciated from Quay hence removing from Quay


openapi.yaml
```diff
@@ -659,7 +659,7 @@ components:
           description: |
             A well defined set of severity strings guaranteed to be present.
           type: string
-          enum: [Unknown, Negligible, Low, Medium, High, Critical, Defcon1]
```
ldelossa, line 662: @syed @alecmerdler 

Will Quay need any updates in it's API validator (specifically with regards to the Clair V4 client in quay) since the openapi spec is changing now?

ldelossa (changes requested):

openapi.yaml
syed: I think I removed it from Quay as well https://github.com/quay/quay/pull/664/files#diff-de511d8a0ffe2df62730e731ccb49271e73c4f276e26bd5a8e4a757e46beb26f

ldelossa: Cool I think that would do it.

ldelossa (approved):
WELCOME TO CLAIR AGAIN!!!!!!!!!!!!!!!!

PR #1170, opened 2021-02-02
updaters: plumb update retention in

ldelossa:

### DIFF
--- a/Documentation/reference/config.md
+++ b/Documentation/reference/config.md
@@ -45,7 +45,9 @@ matcher:
     max_conn_pool: 0
     indexer_addr: ""
     migrations: false
-    updater_sets: []
+    period: ""
+    disable_updaters: false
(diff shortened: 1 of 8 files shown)


hdonnay (changes requested):
If the `Validate` function now sets defaults, the signature needs to change to take a pointer and the documentation should note that it may modify the passed structure.

hdonnay:
Also, should these changes be picked onto the stable branch?

ldelossa:
@hdonnay reworked the config a bit to make "per stanza" validators. 

As far as back-porting. Let's cut v4.1.0 and then do a compare against stable `release 4.0` then decide what needs to get back ported and what does not.

ldelossa:
Note: only backport documentation fixes

hdonnay (approved):
lgtm

PR #1181, opened 2021-02-11
clairctl: Add subpath to clairctl

fhopfensperger:
Allows clairctl to connect to clair, which is located behind an http-subpath of the ingress. For example, clair might only be available at https://example.com/clair, with this PR `clairctl` now also takes the subpath into account with the host property. `clairctl -D report --host https://example.com/clair/`

Signed-off-by: Florian Hopfensperger <florian.hopfensperger@de.ibm.com>

### DIFF
--- a/cmd/clairctl/client.go
+++ b/cmd/clairctl/client.go
@@ -99,7 +99,7 @@ func (c *Client) IndexReport(ctx context.Context, id claircore.Digest, m *clairc
 		req *http.Request
 		res *http.Response
 	)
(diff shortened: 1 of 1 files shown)


hdonnay (dismissed):
LGTM

The whole point of using the `Parse` method was to make this work, but then the `httptransport.*Path` strings are absolute, so it falls apart there.

Thanks!

hdonnay (approved):
rebased onto latest head

PR #1187, opened 2021-02-16
config: handover matcher settings to claircore

arajkumar:
A root level config key `matchers` is added to customize the `Matcher` construction.

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -36,6 +36,21 @@ matcher:
   - "rhel"
   - "suse"
   - "ubuntu"
+matchers:
+  names:
+  - "alpine"
+  - "aws"
+  - "debian"
+  - "oracle"
+  - "photon"
+  - "python"
+  - "rhel"
+  - "suse"
+  - "ubuntu"
+  - "crda"
+  config:
+    crda:
+      url: https://f8a-analytics-preview-2445582058137.production.gw.apicast.io/?user_key=3e42fa66f65124e6b1266a23431e3d08
 notifier:
   indexer_addr: http://clair-indexer:8080/
   matcher_addr: http://clair-matcher:8080/
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/docker/docker v1.13.1 // indirect
 	github.com/go-stomp/stomp v2.0.6+incompatible
+	github.com/golang/mock v1.5.0 // indirect
 	github.com/google/go-cmp v0.5.4
 	github.com/google/go-containerregistry v0.0.0-20191206185556-eb7c14b719c6
 	github.com/google/uuid v1.1.2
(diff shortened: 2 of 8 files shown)


ldelossa (changes requested):
@arajkumar 

Do we need to make a distinction between "RemoteMatcher" configs and general "Matcher" configs? 

I also do not think we need "RemoteMatcher" as top level global key, when a "Matcher" stanza already exists. Can you find a way to do this config inside the "Matcher" stanza and possible make the config agnostic to whether its a "Remote" or "Local" matcher?

arajkumar:
> 
> Do we need to make a distinction between "RemoteMatcher" configs and general "Matcher" configs?

IMO, normal Matcher hardly needs configurability. Do you have any use case for that?
> 
> I also do not think we need "RemoteMatcher" as top level global key, when a "Matcher" stanza already exists. Can you find a way to do this config inside the "Matcher" stanza and possible make the config agnostic to whether its a "Remote" or "Local" matcher?

This is doable.

ldelossa:
We often get requests to block specific Vulns or packages from matching,
this is a good opportunity to allow this configuration into matchers.

arajkumar:
@ldelossa 
>We often get requests to block specific Vulns or packages from matching,

Isn't this come under policy purview? This shall be done at consumer level,.e.g registry org admin shall have a policy to ignore vulns/pkgs which is not affecting their use case.

ldelossa:
Dont think so.

When we implement proper registry all registry components will get config.
No need to special case things.

arajkumar:
@ldelossa , updated  `config.yaml.sample` which moves matcher configurability under `matcher` stanza. Let's first agree upon a config format, I will make code changes later.

arajkumar:
>> Isn't this come under policy purview? This shall be done at consumer level,.e.g registry org admin shall have a policy to ignore vulns/pkgs which is not affecting their use case.

> When we implement proper registry all registry components will get config.

I actually meant container registry.

for example, consider a container image which has a known vuln, but the admin of the container registry org aware that the vuln is not affecting their use case. Then the admin can take the decision to mute a specific vuln/pkg. I'm not sure whether this option is already present in any of the popular registry hubs(quay.io, docker, etc).

config.yaml.sample
```diff
@@ -21,6 +21,9 @@ indexer:
   layer_scan_concurrency: 5
   migrations: true
 matcher:
+  matchers:
```
ldelossa: @arajkumar 

Lets use the way we currently plug "Updater" config into clair as an example: https://quay.github.io/clair/reference/config.html#updaters-object

Can we emulate this same structure with a "Matcher" and "Config" subkey? 

It would be nice to keep things looking the same. @hdonnay please chime in asap if you do not agree.

arajkumar: @ldelossa If I understand correctly, it will look like below,

```
matcher:
    connstring: ""
    max_conn_pool: 0
    indexer_addr: ""
    migrations: false
    period: ""
    disable_updaters: false
    update_retention: 2
    sets:
        - rhel
        - ubuntu
        - python
        - ..
        - crda
    config:
       crda:
           host: https://blabla
           concurrency_limit: 10
````

Sounds good to me.

Comment (ldelossa) [reply config.yaml.sample]: not exactly. Place Matchers block at root of config like updaters is currently. This was a decision to signify multiple services may use that block, not just the matcher (for instance, the omnimatcher in the indexer might)

```
matchers:
  names: (since we don't have a "set" concept for matchers)
    - rhel
    - ubuntu
    - python
    - ...
    - crda
  config:
    crda:
        ...

arajkumar:
Changes are made in sync with https://github.com/quay/claircore/pull/337

go.mod
```diff
@@ -16,24 +17,24 @@ require (
 	github.com/mattn/go-sqlite3 v1.11.0 // indirect
 	github.com/prometheus/client_golang v1.9.0
 	github.com/prometheus/procfs v0.3.0 // indirect
-	github.com/quay/claircore v0.3.1
+	github.com/quay/claircore v0.3.2-0.20210316145942-25dd763a32d2
```
arajkumar: Just a temporary measure until we get a next semver release.

ldelossa (changes requested):
Hey Arun, 

This needs config reference doc updates: https://github.com/quay/clair/blob/main/Documentation/reference/config.md

Other then this, we need to keep in mind that CRDA should *not* wind up in the omnimatcher. What you have is fine, but make sure when you wire this all together, this list of default matcher names do not get auto-injected into the omni-matcher.

arajkumar:
@ldelossa 
>This needs config reference doc updates: https://github.com/quay/clair/blob/main/Documentation/reference/config.md

Done.

>Other then this, we need to keep in mind that CRDA should not wind up in the omnimatcher.

I'll make sure this never happens.

ldelossa (approved):
@hdonnay cc - a bit more config updates if you plan on rebaing jsonPath config commits you have. Take note on these new ones.

Issue #1186, opened 2021-02-16
Big POST request seems to be closed from indexer server side


jan-zmeskal:
### Description of Problem
Hey guys, it's me again. I recently posted #1173 and from what I can gather, it seems to have fixed the problem described in the PR (TLDR: None of the notifier's processors could acquire lock). However, it seems to have uncovered another issue that seems to me even more complicated. I'll try to describe it as best as I can and it'll hopefully give you enough clues to untie this knot, because I'm at my wits' end.

So after #1173 has been merged and deployed to our envs, I made sure that the issue doesn't reproduce anymore. And indeed it hasn't which is good. However, every once in a while, our notifier pod got OOM-killed by OpenShift. As an immediate remedy, I raised the pod's memory limit to 4GiB (from 1.5GiB) and that helped. However I also noticed that one of the processors constantly keeps processing update operation by RHEL7-rhel-7-supplementary updater, seemingly without an end. 

There was no message in the pod's log that would really expose what's going on so I decided to run notifier locally but connected it to all the other parts of our stage env where the problem reproduces. First I just ran it in development container using `make local-dev-notifier-restart` target. It ran for a couple of minutes and then crashed with this:
```
9:35AM DBG lock acquired by another processor. will not process UOID=78d96d7b-7fb5-41b8-aab4-31f6d753c5f7 component=notifier/processor/Processor.process processor_id=3 updater=RHEL7-rhel-7-supplementary
9:35AM INF notification ids in failed status component=notifier/delivery/Delivery.RunDelivery deliverer=amqp-direct- failed=0 id=1
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xcde2c3]
goroutine 13 [running]:
github.com/quay/clair/v4/httptransport/client.(*HTTP).AffectedManifests(0xc000424960, 0xffcfe0, 0xc000424a50, 0xc072380000, 0x231fc, 0x32a16, 0x0, 0x0, 0x0)
        /src/clair/httptransport/client/indexer.go:42 +0x403
github.com/quay/clair/v4/notifier.(*Processor).create(0xc00011a280, 0xffcfe0, 0xc000424a50, 0xc000251c50, 0xc, 0xe648604a4799e171, 0xcdd89dee4f2de3ab, 0xc000250170, 0xc, 0xc0004e36e0, ...)
        /src/clair/notifier/processor.go:127 +0x7aa
[...]

PR #1190, opened 2021-02-17
client: remove request body buffering

hdonnay:
This should prevent any request bodies from being buffered in memory in
their entirety before being sent. I think this will help with #1186.


hdonnay:
Superseded by #1193

PR #1193, opened 2021-02-19
Notifier efficiency

### LINKED ISSUE #1198: Notifier panics at start
When deploying notifier to OCP, it panics at the start. Full log is attached. I'm not gonna go into great detail as this problem is already addressed in #1193. I just created this issue to point out that this can be observed also on the `main` branch and is not unique for Hank's `notifier-efficiency` branch.
[clair-notifier-1-59g25.log](https://github.com/quay/clair/files/6100519/clair-notifier-1-59g25.log)

### Environment
Encountered when running clair from `main` branch, commit 102ae88dd84c1f769b8c037226d92b301d887aab.

hdonnay:
This changeset attempts to improve memory usage across clair, and especially around the `affected_manifests` call.

* It attempts to use `io.Pipe` to cross-wire JSON encoding and API requests, to avoid buffering the entire body request in memory.
* It replaces usage of `encoding/json` with `github.com/ugorji/go/codec` configured for JSON in order to allow streaming the JSON encoding.
* It chunks `affected_manifests` calls in the notifier. This should prevent large vulnerability turnovers from causing extremely large API calls.

cc @jan-zmeskal

### DIFF
--- a/httptransport/client/indexer.go
+++ b/httptransport/client/indexer.go
@@ -3,9 +3,7 @@ package client
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"path"
 
@@ -14,6 +12,7 @@ import (
 	clairerror "github.com/quay/clair/v4/clair-error"
 	"github.com/quay/clair/v4/httptransport"
 	"github.com/quay/clair/v4/indexer"
+	"github.com/quay/clair/v4/internal/codec"
 )
 
 var _ indexer.Service = (*HTTP)(nil)
@@ -23,19 +22,11 @@ func (s *HTTP) AffectedManifests(ctx context.Context, v []claircore.Vulnerabilit
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse api address: %v", err)
 	}
-	rd, wr := io.Pipe()
-	go func() {
-		defer wr.Close()
-		if err := json.NewEncoder(wr).Encode(struct {
-			V []claircore.Vulnerability `json:"vulnerabilities"`
-		}{
-			v,
-		}); err != nil {
-			wr.CloseWithError(err)
-		}
-	}()
-	defer rd.Close()
-
+	rd := codec.JSONReader(struct {
+		V []claircore.Vulnerability `json:"vulnerabilities"`
+	}{
+		v,
+	})
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), rd)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create request: %v", err)
@@ -56,7 +47,9 @@ func (s *HTTP) AffectedManifests(ctx context.Context, v []claircore.Vulnerabilit
 	var a claircore.AffectedManifests
 	switch ct := req.Header.Get("content-type"); ct {
 	case "", `application/json`:
-		if err := json.NewDecoder(resp.Body).Decode(&a); err != nil {
+		dec := codec.GetDecoder(resp.Body)
+		defer codec.PutDecoder(dec)
+		if err := dec.Decode(&a); err != nil {
 			return nil, err
 		}
 	default:
@@ -76,16 +69,8 @@ func (s *HTTP) Index(ctx context.Context, manifest *claircore.Manifest) (*clairc
 	if err != nil {
 		return nil, fmt.Errorf("failed to create request: %v", err)
 	}
-	rd, wr := io.Pipe()
-	go func() {
-		defer wr.Close()
-		if err := json.NewEncoder(wr).Encode(manifest); err != nil {
-			wr.CloseWithError(err)
-		}
-	}()
-	defer rd.Close()
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), rd)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), codec.JSONReader(manifest))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create request: %v", err)
 	}
@@ -105,7 +90,9 @@ func (s *HTTP) Index(ctx context.Context, manifest *claircore.Manifest) (*clairc
 	var ir claircore.IndexReport
 	switch ct := resp.Header.Get("content-type"); ct {
 	case "", `application/json`:
-		if err := json.NewDecoder(resp.Body).Decode(&ir); err != nil {
+		dec := codec.GetDecoder(resp.Body)
+		defer codec.PutDecoder(dec)
+		if err := dec.Decode(&ir); err != nil {
 			return nil, err
 		}
 	default:
@@ -143,8 +130,9 @@ func (s *HTTP) IndexReport(ctx context.Context, manifest claircore.Digest) (*cla
 	}
 
 	ir := &claircore.IndexReport{}
-	err = json.NewDecoder(resp.Body).Decode(ir)
-	if err != nil {
+	dec := codec.GetDecoder(resp.Body)
+	defer codec.PutDecoder(dec)
+	if err := dec.Decode(ir); err != nil {
 		return nil, false, &clairerror.ErrBadIndexReport{E: err}
 	}
 	return ir, true, nil
--- a/httptransport/client/indexer.go
+++ b/httptransport/client/indexer.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"path"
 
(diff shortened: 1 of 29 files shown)


hdonnay:
@ldelossa I'm unsure if the composed `AffectedManifest` objects need a pass to deduplicate and sort the `VulnerableManifests` member.

hdonnay:
Undrafting because I added a `Sort` call.

notifier/processor.go
```diff
@@ -126,24 +127,35 @@ func (p *Processor) create(ctx context.Context, e Event, prev uuid.UUID) error {
...
-		Vulnerabilities:     make(map[string]*claircore.Vulnerability),
-		VulnerableManifests: make(map[string][]string),
+	tab := notifTab{
+		N:      make([]Notification, 0),
+		lookup: make(map[string]int),
 	}
 	eg, wctx := errgroup.WithContext(ctx)
-	eg.Go(getAffected(wctx, p.indexer, diff.Added, added))
-	eg.Go(getAffected(wctx, p.indexer, diff.Removed, removed))
+	eg.Go(getAffected(wctx, p.indexer, !p.NoSummary, diff.Added, Added, &tab))
```
ldelossa: Could you keep the same meaning of "NoSummary" all the way down, even into the getAffected method, instead of negating the bool here?

notifier/processor.go
```diff
@@ -213,10 +191,16 @@ func min(a, b int) int {
 	return b
 }
 
+type notifTab struct {
```
ldelossa, line 197: comment the purpose of this.

httptransport/client/indexer.go
```diff
@@ -13,43 +12,50 @@ import (
...
 		return nil, err
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return nil, &clairerror.ErrRequestFail{Code: resp.StatusCode, Status: resp.Status}
-	}
-	err = json.NewDecoder(resp.Body).Decode(&affected)
-	if err != nil {
-		return nil, &clairerror.ErrBadAffectedManifests{err}
+		return nil, &clairerror.ErrRequestFail{
```
ldelossa, line 41: Why the change in error type?

httptransport/client/indexer.go
```diff
@@ -59,36 +65,40 @@ func (s *HTTP) AffectedManifests(ctx context.Context, v []claircore.Vulnerabilit
...
+	var ir claircore.IndexReport
+	switch ct := resp.Header.Get("content-type"); ct {
+	case "", `application/json`:
+		dec := codec.GetDecoder(resp.Body)
+		defer codec.PutDecoder(dec)
+		if err := dec.Decode(&ir); err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("unrecognized content-type %q", ct)
```
ldelossa, line 99: I don't know if forcing the client to provide the json content type is that useful. 

We export a JSON api, we should try to parse json, regardless of whether a client is providing a content type or not. I think we ran into this already and it was an annoyance more then a feature.

httptransport/client/matcher.go
```diff
@@ -25,31 +25,33 @@ func (c *HTTP) Scan(ctx context.Context, ir *claircore.IndexReport) (*claircore.
...
-		return nil, err
+	switch ct := req.Header.Get("content-type"); ct {
+	case "", `application/json`:
+		dec := codec.GetDecoder(resp.Body)
+		defer codec.PutDecoder(dec)
+		if err := dec.Decode(&vr); err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("unrecognized content-type %q", ct)
```
ldelossa, line 54: same, don't think this should be a requirement.

httptransport/notificationshandler.go
```diff
@@ -149,9 +146,8 @@ func (h *NotifHandler) Get(w http.ResponseWriter, r *http.Request) {
 		Notifications: notifications,
 	}
 
-	err = json.NewEncoder(w).Encode(&response)
-	if err != nil {
-		log.Error().Err(err).Msg("failed to serialize response")
-	}
-	return
+	defer writerError(w, &err)()
```
ldelossa, line 149: Still confused about whether this is too useful, how is anything going to know to check a trailer?

httptransport/updateoperationhandler.go
```diff
@@ -114,7 +115,5 @@ func (h *UOHandler) Delete(w http.ResponseWriter, r *http.Request) {
 			Message: fmt.Sprintf("could not get update operations: %v", err),
 		}
 		je.Error(w, resp, http.StatusInternalServerError)
-		return
```
ldelossa, line 117: is this a linter thing?

internal/codec/codec_test.go
```diff
@@ -0,0 +1,20 @@
...
+func Example() {
+	enc := GetEncoder(os.Stdout)
+	defer PutEncoder(enc)
+	enc.MustEncode([]string{"a", "slice", "of", "strings"})
+	fmt.Fprintln(os.Stdout)
+	enc.MustEncode(nil)
+	fmt.Fprintln(os.Stdout)
+	enc.MustEncode(map[string]string{})
+	fmt.Fprintln(os.Stdout)
+	// Output: ["a","slice","of","strings"]
```
ldelossa, line 22: probably want to test with json large then your 4096 buffer size to make sure things are OK.

notifier/callback.go
```diff
@@ -12,40 +13,39 @@ import (
...
+// MarshalJSON implements json.Marshaler.
 func (cb Callback) MarshalJSON() ([]byte, error) {
-	var m = map[string]string{
-		"notification_id": cb.NotificationID.String(),
-		"callback":        cb.Callback.String(),
-	}
-	return json.Marshal(m)
+	// This writes out a block of bytes directly, instead of allocating a
+	// cbMessage structure, serializing the URL, then copying it into a []byte.
+	b := bytes.NewBufferString(`{"callback":"`)
```
ldelossa: this looks like a rough thing to maintain if the schema changes. 
is this strictly necessary?

notifier/processor.go
```diff
@@ -124,17 +126,36 @@ func (p *Processor) create(ctx context.Context, e Event, prev uuid.UUID) error {
...
+				removed++
+			}
+		}
+		ev.
+			Int("added", added).
+			Int("removed", removed).
+			Msg("affected manifest counts")
+	}
+
+	if len(tab.N) == 0 {
```
ldelossa, line 158: why is this condition added? You're still processing all potential notifications in this method call right?

ldelossa (changes requested):
I'm not 100% sure the pipe driven JSONReader should be used across the board. There are resource considerations to initializing the channels it uses internally. It may not be necessary to use them where memory constraints are not a concern.

notifier/processor.go
hdonnay: It's the equivalent to `if len(added.VulnerableManifests) == 0 && len(removed.VulnerableManifests) == 0 {`

httptransport/client/indexer.go
hdonnay: I don't know what you mean, it's still `ErrRequestFail`.

httptransport/client/indexer.go
hdonnay: hence ```case "", `application/json`:```

httptransport/notificationshandler.go
hdonnay: It's called out in the headers.

I waffle on how useful reporting it via the trailer is, but this is at least centralizing handling the writer errors in one spot, one way.

httptransport/client/indexer.go
ldelossa: Why the switch case then? If you're catching the default case?

httptransport/updateoperationhandler.go
hdonnay: I think so. Something flagged the useless returns for me, iirc

notifier/callback.go
hdonnay: Strictly necessary, no, but these methods are the only places where we don't get the piecewise encoding of the `codec` package "for free," so I thought it'd be prudent to minimize the number of intermediary allocations here.

ldelossa: Do you think you loose the majority of this PRs effectiveness if we moved away from writing out the json directly?

httptransport/client/indexer.go
ldelossa: Misread that

httptransport/client/indexer.go
hdonnay: To return an error if other content-types are specified

notifier/callback.go
hdonnay: The majority? No, it just seemed like low-hanging fruit.

jan-zmeskal:
@hdonnay @ldelossa I obviously don't want to rush this PR as I support your drive to do it right. I was just thinking if we could get 5c1b27fba36973620b41de8bc1f3fec2d811f09e merged separately, as many people seem to run into this issue on the main branch.

notifier/callback.go
ldelossa: Id prefer to not do this, string building json json is pretty ugly, and not great to maintain. As long as we increase the majority of the efficiency, we can still keep maintainability in mind. Amdahl's law comes to mind.

httptransport/client/indexer.go
ldelossa: Right, so im saying i dont think we should restrict it at all, we have a JSON api, try to parse the JSON and return an error if invalid. Is there any harm in not being defensive here for convenience ?

hdonnay:
regarding the pipes, the JSON encoder is doing some buffering, so it's not like every object is incurring channel overhead.

Jan reported memory consumption dropping >10x in a bad instance; I'm okay with paying some cycles to get that sort of memory pressure reduction.

ldelossa (approved):

ldelossa (approved):

PR #1201, opened 2021-03-10
httptransport: use correct handler for state endpoint

hdonnay:
Looks like this was missed in the metrics pass and review.

This should resolve some part of #1199.


ldelossa (approved):
Thanks

PR #1211, opened 2021-03-22
Ratelimiter

hdonnay:
Adds a ratelimiter and more correct Client usage, and wires it up where it seems appropriate.

### DIFF
--- a/internal/httputil/ratelimiter.go
+++ b/internal/httputil/ratelimiter.go
@@ -1,0 +1,67 @@
+package httputil
+
+import (
+	"net/http"
+	"sync"
+
+	"golang.org/x/time/rate"
+)
+
+// RateLimiter wraps the provided RoundTripper with a limiter allowing 10
+// requests/second/host.
+//
+// It responds to HTTP 429 responses by automatically decreasing the rate.
+func RateLimiter(next http.RoundTripper) http.RoundTripper {
+	return &ratelimiter{
+		rt: next,
+		lm: sync.Map{},
+	}
+}
+
+// Ratelimiter implements the limiting by using a concurrent map and Limiter
+// structs.
+type ratelimiter struct {
+	rt http.RoundTripper
+	lm sync.Map
+}
+
+const rateCap = 10
+
+// RoundTrip implements http.RoundTripper.
+func (r *ratelimiter) RoundTrip(req *http.Request) (*http.Response, error) {
+	key := req.URL.Host
+	li, ok := r.lm.Load(key)
+	if !ok {
+		// Limiter allows "rateCap" per sec, one at a time.
+		l := rate.NewLimiter(rate.Limit(rateCap), 1)
+		li, _ = r.lm.LoadOrStore(key, l)
+	}
+	l := li.(*rate.Limiter)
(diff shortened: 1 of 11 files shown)


internal/httputil/ratelimiter.go
```diff
@@ -0,0 +1,62 @@
...
+	res, err := r.rt.RoundTrip(req)
+	// This seems to be the contract that http.Transport implements.
+	if err != nil {
+		return nil, err
+	}
+	// This doesn't seem entirely correct, but will at least make number go
+	// down.
+	if res.StatusCode == http.StatusTooManyRequests {
+		l.SetLimit(detune(l.Limit()))
+	}
```
crozzy, line 57: I may be reading this wrong, but when/if the downstream recovers, would the limit be "stuck" at a lower than optimal rate?

hdonnay: Yes.

I'm not sure what to do to re-ramp the limit. Using `SetLimitAt` would result in the limit getting similarly stuck, by my thinking. Perhaps a second Limiter with token representing an increase to the limit?

crozzy: I am not so familiar with the request patterns or how a certain node can become overwhelmed, but would it make sense to have a second case (or an else case) that increases the Limit (retune) when `status == <good status>` and then taking the `Min(newLimit, lim)` to ensure you don't go over the max? (In general, detune should reduce faster than retune increases it).

In the long run it seems like using a client to mitigate against server errors would lead to a different loadbalancing strategy to the `RoundTrip`, but that seems like a discussion for a different time.

hdonnay:
Here's output from `time`, dumping to `/dev/null`:

```
go run ./cmd/clairctl -c config.yaml.sample export-updaters > /dev/null  175.86s user 6.42s system 204% cpu 1:29.10 total
```

hdonnay:
NB: Failing because of dependency on unpublished `claircore` changes. Will rebase with a new version when those changes are merged.

ldelossa:
> Here's output from `time`, dumping to `/dev/null`:
> 
> ```
> go run ./cmd/clairctl -c config.yaml.sample export-updaters > /dev/null  175.86s user 6.42s system 204% cpu 1:29.10 total
> ```

in comparison to... ?

hdonnay:
on `main`:

```
go run ./cmd/clairctl -c config.yaml.sample export-updaters > /dev/null  335.48s user 9.87s system 249% cpu 2:18.41 total
```

ldelossa:
> on `main`:
> 
> ```
> go run ./cmd/clairctl -c config.yaml.sample export-updaters > /dev/null  335.48s user 9.87s system 249% cpu 2:18.41 total
> ```

You see a consecutive speed up? Interesting it got faster.

hdonnay:
What do you mean "consecutive?"

ldelossa:
> What do you mean "consecutive?"

Do you see a speedup over more then just one iteration of the comparison? Im not sure why there would be a speedup

hdonnay:
Repeat runs of this branch were around 1:40.
Repeat runs of `main` were around 2:10.

ldelossa (approved):

PR #1232, opened 2021-04-09
codec: use stdlib time.Time encoding

### LINKED ISSUE #1231: Vulnerability report can break API contract due to time.Time being null

Using the library https://github.com/ugorji/go/ to do the JSON response encoding means that all `time.Time` objects that evaluate to `IsZero() == true` will be marshaled as `"time": null`. 

This means that if, when grabbing `vuln`s from the DB, they have an `Issued` time of `0001-01-01 00:00:00+00` they will end up being send back from the API as something like:

```
	"vulnerabilities": {
		"931353": {
			"id": "931353",
			"updater": "pyupio",
			"name": "pyup.io-26132",
			"description": "setuptools before 1.3 has a security vulnerability in SSL match_hostname check as reported in Python 17997.",
			"issued": null,
			"links": "",
			"severity": "",
			"normalized_severity": "Unknown",
			"package": {
				"id": "",
				"name": "setuptools",
				"version": "\u003c1.3",
				"kind": "binary"
			},
```

This isn't a huge deal unless the consumer is validating the response as Quay does and throws this error:
```
gunicorn-web stdout | jsonschema.exceptions.ValidationError: None is not of type 'string'
gunicorn-web stdout | Failed validating 'type' in schema['properties']['vulnerabilities']['additionalProperties']['properties']['issued']:
gunicorn-web stdout |     {'description': 'The timestamp in which the vulnerability was issued\n',
gunicorn-web stdout |      'type': 'string'}
gunicorn-web stdout | On instance['vulnerabilities']['931353']['issued']:

```

### Expected Outcome

The JSON is always a string (or Quay is updated to accept null).

### Actual Outcome

The `issued` time can be null and Quay seems to expect a string.

### Environment

- Clair version/image: main
- Clair client name/version: main
- Host OS: Linux
- Kernel (e.g. `uname -a`): 4.15.0-34-generic
- Kubernetes version (use `kubectl version`): NA
- Network/Firewall setup: docker-compose

This might not be an issue, or could be an environment quirk, but it's something I ran into while running locally.

hdonnay:
The new json encoder has different behavior for encoding zero time.Time
values. This change forces the Marshal method to be called, which gets
back the previous behavior (always produce a string) at the cost of
doing an allocation.

Closes #1231

### DIFF
--- a/internal/codec/codec.go
+++ b/internal/codec/codec.go
@@ -15,6 +15,10 @@ func init() {
 	// This is documented to cause "smart buffering".
 	jsonHandle.WriterBufferSize = 4096
 	jsonHandle.ReaderBufferSize = 4096
+	// Force calling time.Time's Marshal function. This causes an allocation on
+	// every time.Time value, but is the same behavior as the stdlib json
+	// encoder. If we decide nulls are OK, this should get removed.
+	jsonHandle.TimeNotBuiltin = true
 }
 
 // Encoder and decoder pools, to reuse if possible.
--- a/internal/codec/codec_test.go
+++ b/internal/codec/codec_test.go
@@ -1,11 +1,13 @@
 package codec
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 )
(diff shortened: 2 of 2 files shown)


ldelossa (dismissed):
Follow this up with a v4.1-alpha.2 release. 

cc @thomasmckay

ldelossa (approved):

PR #1238, opened 2021-04-14
Fixing some small testing race conditions

crozzy:
These are both pretty insignificant as they are only present in the testing code, but I suppose it helps to not pollute the testing output when searching for real-world race conditions.

I added the target in the test file for unit-tests, that should probably reflect what the CI is running unit-test wise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,3 +159,8 @@ docs-build:
 	mdbook build
 	rsync --recursive --delete-after --exclude 'v4.*' --exclude .git\
 		./book/ $(DOCS_DIR)/
+
+# runs unit tests
+.PHONY: unit
+unit:
+	go test -race ./...
--- a/notifier/processor_create_test.go
+++ b/notifier/processor_create_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync/atomic"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -159,3 +159,8 @@ docs-build:
 	mdbook build
 	rsync --recursive --delete-after --exclude 'v4.*' --exclude .git\
 		./book/ $(DOCS_DIR)/
+
+.PHONY: unit-test
+unit-test: vendor
```
ldelossa: small nit, but would you mind making this similar to the makefile in claircore. 

https://github.com/quay/claircore/blob/main/Makefile#L29

just for sanity sake, using the same keyword is helpful.

crozzy: Sure, I'll just match them and the coverage stuff can be a discussion outside of this.

notifier/processor_create_test.go
```diff
@@ -216,13 +217,13 @@ func testProcessorCreate(t *testing.T) {
 			}, nil
 		},
 	}
-	count := 0
+	count := uint64(0)
```
ldelossa, line 220: This looks good, p.create was updated recently to do work inside go routines, and this must have been missed.

Makefile
crozzy: done

ldelossa (approved):

ldelossa:
Thanks for this.

PR #1254, opened 2021-04-26
httptransport: fix LatestUpdateOperations method

jan-zmeskal:
It's not that the changed method was broken, but it actually returned _all_ update operations for each updater.
Before my change: 
image

After it:
image

It's unnecessary to return all of the update operations because from [this point](https://github.com/quay/clair/blob/1f9b56577957ce28044b221af58b160328a671a2/notifier/poller.go#L118), we only work with the latest one anyway. I think with this, the actual workings of the function and its docstring are more in sync.

Signed-off-by: Jan Zmeskal <jzmeskal@redhat.com>

### DIFF
--- a/httptransport/client/matcher.go
+++ b/httptransport/client/matcher.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


ldelossa (approved):
Nice catch.

PR #1256, opened 2021-04-26
clairctl: move to zlog

hdonnay:
This makes the command more chatty by default, which I don't particularly like, but it makes `clairctl`'s output commands actually affect the logging of the rest of the system, which is almost certainly what a user expects.

I added a flag to quiet the output as a poor imitation of my beloved terse output.

This change is motivated by noticing that the `zlog` move meant that the `clairctl` output was completely uncontrollable.


ldelossa (commented):
Json logs for cli output??

ldelossa (approved):

hdonnay:
This was merged, I goofed up the order of operations :disappointed:

PR #1267, opened 2021-05-11
cc bump

hdonnay:
This pulls the latest version of claircore and makes needed calling changes.

### DIFF
--- a/httptransport/client/matcher.go
+++ b/httptransport/client/matcher.go
@@ -129,16 +129,19 @@ func (c *HTTP) DeleteUpdateOperations(ctx context.Context, ref ...uuid.UUID) (in
 
 // LatestUpdateOperation shouldn't be used by client code and is implemented
 // only to satisfy the matcher.Differ interface.
-func (c *HTTP) LatestUpdateOperation(_ context.Context) (uuid.UUID, error) {
+func (c *HTTP) LatestUpdateOperation(_ context.Context, _ driver.UpdateKind) (uuid.UUID, error) {
 	return uuid.Nil, nil
 }
 
 // UpdateOperations returns all the known UpdateOperations per updater.
-func (c *HTTP) UpdateOperations(ctx context.Context, updaters ...string) (map[string][]driver.UpdateOperation, error) {
+func (c *HTTP) UpdateOperations(ctx context.Context, k driver.UpdateKind, updaters ...string) (map[string][]driver.UpdateOperation, error) {
 	u, err := c.addr.Parse(httptransport.UpdateOperationAPIPath)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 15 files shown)


httptransport/client/matcher.go
```diff
@@ -147,13 +150,14 @@ func (c *HTTP) UpdateOperations(ctx context.Context, updaters ...string) (map[st
...
 // LatestUpdateOperations returns the most recent UpdateOperation per updater.
-func (c *HTTP) LatestUpdateOperations(ctx context.Context) (map[string][]driver.UpdateOperation, error) {
+func (c *HTTP) LatestUpdateOperations(ctx context.Context, k driver.UpdateKind) (map[string][]driver.UpdateOperation, error) {
 	u, err := c.addr.Parse(httptransport.UpdateOperationAPIPath)
 	if err != nil {
 		return nil, err
 	}
 	v := url.Values{}
 	v.Add("latest", "true")
+	v.Add("kind", string(k))
```
ldelossa, line 160: I think we talked about special casing an empty driver.UpdateKind ( value: "" ) to signal that the UpdateKind "vulnerability" should be used. 

https://github.com/quay/claircore/blob/main/internal/vulnstore/postgres/getupdateoperations.go

In these, and similar methods, I think we want to send off a "vulnerability" value of driver.UpdateKind as the "default", so that when it hits the ClairCore code, we maintain API compatibility. 

i.e. 

```
switch driver.UpdateKind {
case "" || driver.VulnerabilityKind:
    ... 
    v.Add("kind", "vulnerability")
case driver.EnrichmentKind:
   v.Add("kind", "enrichment")
default:
   return nil, fmt.Errorf("wtf")
}
```

ldelossa: scrolled down a bit and saw you handle this in the actual handler, that's fine too. Maybe worth doing it in both places for self-documentation?

initialize/services.go
```diff
@@ -133,7 +134,16 @@ func localIndexer(ctx context.Context, cfg *config.Config) (indexer.Service, err
 			opts.ScannerConfig.Repo[name] = node.Decode
 		}
 	}
-	s, err := libindex.New(ctx, &opts)
+	tr := http.DefaultTransport.(*http.Transport).Clone()
```
ldelossa, line 137: whats the context around this change?

ldelossa (approved):
This looks good, one comment which I made too early but brings up a good point about defensiveness. You can use your discretion as the driver.UpdateKind handling in the http handlers works fine.

initialize/services.go
hdonnay: New Libvuln API now requires an http.Client at construction. So this constructs a client via the config's method and then passes it in.

PR #1274, opened 2021-05-21
matcher: disable updaters creates empty updater sets

### LINKED ISSUE #1273: Matcher disable updaters config not woking in v4.1.0

In relase v4.1.0 the code that checks the matcher config value of `disable_updaters` was removed, meaning that updaters will be enabled even with this value set to true.

### Expected Outcome

Setting `matcher.disable_updaters` to `true` in the config sets the Updaters `Sets` value to an empty list, disabling updaters for this matcher instance.

### Actual Outcome

Setting `matcher.disable_updaters` has no effect. 

### Environment

- Clair version/image: v1.4.0

paulaldridge:
Setting disable_updaters to true in the config should result in an empty
updaters sets list

Fixes #1273

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -64,6 +64,9 @@ func Validate(conf *Config) error {
 	if conf.HTTPListenAddr == "" {
 		conf.HTTPListenAddr = DefaultAddress
 	}
+	if conf.Matcher.DisableUpdaters {
+		conf.Updaters.Sets = []string{}
+	}
 	switch strings.ToLower(conf.Mode) {
 	case ComboMode:
 		if err := conf.Indexer.Validate(true); err != nil {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -12,9 +12,8 @@ import (
 
 func Test_Config_Validate_Failure(t *testing.T) {
 	var table = []struct {
-		name   string
-		conf   config.Config
-		expect bool
+		name string
+		conf config.Config
 	}{
 		{
 			name: "No Mode",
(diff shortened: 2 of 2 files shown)


crozzy:
This seems like a sensible change, but I think maybe [config.Validate()](https://github.com/quay/clair/blob/45538e0ccee4d4ee12247da11f295ff32dcbe4ff/config/config.go#L63)  is better location for this change, it would also avoid setting a package global to allow the test to mock the initializing libvuln function. Thoughts?

initialize/services_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Matcher: config.Matcher{
+			DisableUpdaters: true,
+		},
+		Updaters: config.Updaters{
+			Sets: []string{"alpine", "aws"},
+		},
+	}
+
+	_, err := localMatcher(ctx, &cfg)
+	assert.NoError(t, err)
```
crozzy: It would be nice to avoid adding a new dependency for this line, nothing against testify.

(I know this is probably over-shared https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s&ab_channel=TheGoProgrammingLanguage)

paulaldridge:
Thanks for reviewing. Nice suggestion, it does work much better in `config.Validate()`. How does it look now?

initialize/services_test.go
paulaldridge: I agree with this - good thoughts. I hadn’t actually seen the video, he makes some good points!

crozzy (dismissed):
Looks great to me @paulaldridge, clean and nice tests

paulaldridge:
Thanks!

hdonnay (approved):

PR #1284, opened 2021-06-03
config: omit Authorization header for empty claims

### LINKED ISSUE #1283: Clair wrongly injects Authorization headers when fetching layers
tl;dr: When sending HTTP requests to fetch layers, Clair injects an extraneous `Authorization` header which may confuse the object store containing the layers.

---

Our internal registry uses OpenStack Swift as a backing storage, and Clair for scanning vulnerabilities. We have Clair set up with the `auth.psk` config section, and our registry issues tokens to itself to talk to the Clair API. After upgrading Clair from 4.0.2 to 4.1.0, we saw indexing in Clair fail with:

```
failed to fetch layers: encountered error while fetching a layer: fetcher: unexpected status code: 403 403 Forbidden
```

After some digging, the problem turned out to be that Clair adds an `Authorization` header to each HTTP request, containing a Bearer token derived from the `auth.psk` key:

https://github.com/quay/clair/blob/45538e0ccee4d4ee12247da11f295ff32dcbe4ff/config/httpclient.go#L75-L88

This `Authorization` header trips up Swift, since the issued token is only valid inside of Clair itself. The Swift authentication is supposed to be handled by query parameters in the layer URL as provided by in the index request, but the extraneous `Authorization` header makes Swift ignore all that.

This code has not been touched in a while, so I'm not entirely sure why this did not happen with Clair 4.0.2, but it's very reproducible on Clair 4.1.0. The correct behavior here would be to honor the index request and fetch the layer only with the headers included therein.

### Environment

- Clair version/image: 4.1.0
- Clair client name/version: N/A
- Host OS: Linux
- Kernel (e.g. `uname -a`): N/A
- Kubernetes version (use `kubectl version`): N/A
- Network/Firewall setup: N/A

hdonnay:
This change makes the HTTP client configuration method accept a Claims
pointer, and if `nil` is passed, omits the automatic signing that would
happen if a PSK was configured.

Fixes #1283.

### DIFF
--- a/cmd/clairctl/export.go
+++ b/cmd/clairctl/export.go
@@ -66,7 +66,7 @@ func exportAction(c *cli.Context) error {
 	}
 
 	tr := http.DefaultTransport.(*http.Transport).Clone()
-	cl, _, err := cfg.Client(httputil.RateLimiter(tr), commonClaim)
+	cl, _, err := cfg.Client(httputil.RateLimiter(tr), &commonClaim)
 	if err != nil {
 		return err
 	}
--- a/cmd/clairctl/import.go
+++ b/cmd/clairctl/import.go
@@ -35,7 +35,7 @@ func importAction(c *cli.Context) error {
 		return err
 	}
 
-	cl, _, err := cfg.Client(nil, commonClaim)
+	cl, _, err := cfg.Client(nil, &commonClaim)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


hdonnay:
@majewsky could you give this a try when you get a chance?

majewsky:
@hdonnay Thanks for your quick response on this issue. Unfortunately, I will be on vacation next week, and I did not get to test-driving this PR today. I'll get back to you when I'm back from vacation.

hdonnay:
Feeling confident about this, so I'm going to merge and we can deal with any edge cases that pop-up.

PR #1286, opened 2021-06-07
auto: add automatic runtime configuration

hdonnay:
This change adds an init hook that adjusts GOMAXPROCS, if not set in the
environment already, according to the runtime environment. On Linux,
this means examining CPU quotas.

Doing this will help control latencies by avoiding repeatedly blowing
the CPU budget in containers when run in a container on a big machine.
Such setups usually expose all the cores and then use quotas to constrain the CPU usage.

### DIFF
--- a/initialize/auto/auto.go
+++ b/initialize/auto/auto.go
@@ -1,0 +1,24 @@
+// Package auto does automatic detection and runtime configuration for certain
+// environments.
+//
+// All top-level functions are not safe to call concurrently.
+package auto
+
+import (
+	"context"
+)
+
+var msgs = []func(context.Context){}
+
+func init() {
+	CPU()
+}
+
+// PrintLogs uses zlog to report any messages queued up from the runs of
+// functions since the last call to PrintLogs.
+func PrintLogs(ctx context.Context) {
+	for _, f := range msgs {
+		f(ctx)
+	}
+	msgs = msgs[:0]
+}
--- a/initialize/auto/cpu_linux.go
+++ b/initialize/auto/cpu_linux.go
@@ -1,0 +1,131 @@
+package auto
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"io/fs"
+	"os"
+	"path"
+	"runtime"
+	"strconv"
+
+	"github.com/quay/zlog"
+)
+
+// CPU guesses a good number for GOMAXPROCS based on information gleaned from
+// the current process's cgroup.
+func CPU() {
+	if os.Getenv("GOMAXPROCS") != "" {
+		msgs = append(msgs, func(ctx context.Context) {
+			zlog.Info(ctx).Msg("GOMAXPROCS set in the environment, skipping auto detection")
+		})
+		return
+	}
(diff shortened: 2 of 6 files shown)


hdonnay:
Just to prove that all these build constraints work:

```
% for go in go1.15rc1 go go1.17rc1; do $go version && $go test -count=1 ./initialize/auto; done                      
go version go1.15rc1 linux/amd64
ok  	github.com/quay/clair/v4/initialize/auto	0.001s [no tests to run]
go version go1.16.6 linux/amd64
ok  	github.com/quay/clair/v4/initialize/auto	0.003s
go version go1.17rc1 linux/amd64
ok  	github.com/quay/clair/v4/initialize/auto	0.003s
```

initialize/auto/cpu_linux.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		msgs = append(msgs, func(ctx context.Context) {
+			zlog.Error(ctx).
+				Err(err).
+				Msg("unable to guess GOMAXPROCS value")
+		})
+		return
+	}
+	prev := runtime.GOMAXPROCS(gmp)
+	msgs = append(msgs, func(ctx context.Context) {
```
crozzy, line 36: I think by definition this will only be run once and in one routine so probably good concurrency wise, but might be worth flagging that it's not concurrency safe in case of future expansion.

initialize/auto/cpu_linux.go
```diff
@@ -0,0 +1,134 @@
...
+		zlog.Info(ctx).
+			Int("cur", gmp).
+			Int("prev", prev).
+			Msg("set GOMAXPROCS value")
+	})
+}
+
+func cglookup(r fs.FS) (int, error) {
```
crozzy: Nit: maybe name cgLookup

hdonnay:
Rebased and dropped version constraints now that 1.15 is out of upstream support.

initialize/auto/auto.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+var msgs = []func(context.Context){}
+
+func init() {
+	CPU()
+}
+
+// Logs uses zlog to report any messages queued up during initialization.
+func Logs(ctx context.Context) {
```
crozzy: Bit of a weird one as it's not returning anything or really changing the state but maybe a name that's more imperative. What to you think?

crozzy (commented):
LGTM, just a few nitty comments. Is the idea that anything needing the value should look it up directly from `runtime.GOMAXPROCS(0)`?

hdonnay:
> LGTM, just a few nitty comments. Is the idea that anything needing the value should look it up directly from `runtime.GOMAXPROCS(0)`?

Yeah, the idea is that we can depend on `runtime.GOMAXPROCS(0)` instead of `runtime.NumCPU()` for scaling concurrent work. In the worst case (if this package does nothing) they're the same, and in the best case, the process will be less likely to blow its quota and get throttled.

crozzy (approved):

PR #1289, opened 2021-06-09
shutdown: introduce the new NotifyContext

crozzy:
As of Go 1.16 signal.NotifyContext ties the context to
the signal handling instead of having to do it manually.
This PR also bumps the Go version.

### DIFF
--- a/cmd/clair/main.go
+++ b/cmd/clair/main.go
@@ -117,43 +117,25 @@ func main() {
 	})
 
 	// Signal handler goroutine.
-	srvs.Go(func() error {
-		c := make(chan os.Signal, 1)
-		signal.Notify(c, os.Interrupt)
+	go func() {
+		ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
 		defer func() {
-			signal.Stop(c)
-			close(c)
+			// Note that we're using a background context here, so that we get a
+			// full timeout if the signal handler has fired.
+			tctx, done := context.WithTimeout(context.Background(), 10*time.Second)
+			err := down.Shutdown(tctx)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


hdonnay:
I'd like to keep 1.15 support as long as it's supported. So, let's undraft this after the 1.17 release, which should be next month.

hdonnay:
This can probably be dusted off, now.

hdonnay (dismissed):
LGTM, but also I touched this, so @crozzy take a look also please.

cmd/clair/main.go
```diff
@@ -117,44 +117,26 @@ func main() {
...
-			zlog.Info(ctx).Msg("unregistered signal handler")
-		}()
-		zlog.Info(ctx).Msg("registered signal handler")
-		select {
-		case sig := <-c:
-			zlog.Info(ctx).
-				Stringer("signal", sig).
-				Msg("gracefully shutting down")
 			// Note that we're using the root context here, so that we get a
 			// full timeout if one errgroup goroutine returns uncleanly.
```
hdonnay: I think this comment is incorrect now

cmd/clair/main.go
```diff
@@ -117,44 +117,26 @@ func main() {
...
 			tctx, done := context.WithTimeout(ctx, 10*time.Second)
```
hdonnay:
```suggestion
			// Note that we're using a background context here, so that we get a
			// full timeout if the signal handler has fired.
			tctx, done := context.WithTimeout(context.Background(), 10*time.Second)
```

hdonnay (changes requested):

hdonnay (approved):

PR #1290, opened 2021-06-11
backport v4.1: config: omit Authorization header for empty claims

hdonnay:
This change makes the HTTP client configuration method accept a Claims
pointer, and if `nil` is passed, omits the automatic signing that would
happen if a PSK was configured.

Backports #1284.


crozzy (approved):
Nice

PR #1299, opened 2021-06-16
docker: reflect quay Dockerfile updates

crozzy:
The working directory for quay inside the container
was updated from `/` to `/quay-registry` so the config
needs to be volumed in there.


hdonnay (approved):
LGTM

PR #1313, opened 2021-07-14
Keymanager removal

hdonnay:
I could have sworn I did this before, but I must have lost the branch somewhere.

This set of changes removes the public key signing scheme that was never actually used.


crozzy (dismissed):
Feel like I have also read something like this PR before as well :man_shrugging: weird

hdonnay:
Ah, I did some of this on #1236 but never un-drafted that. I'll have to circle back and rebase that branch.

crozzy (approved):

PR #1318, opened 2021-07-20
introspection: use the response recorder

crozzy:
Previously the logs were showing only 200 responses, passing
the ResponseRecorder to the handler in place of the http.ResponseWriter
should ensure that when Write() and WriteHeader() are called, the
ResponseRecorder logs information about the http response.


hdonnay (approved):
:facepalm: LGTM

PR #1319, opened 2021-07-21
deployment: Fix microdnf install inconsistencies

lkrajice:
The microdnf generates `/etc/yum.repos.d/redhat.repo` file which
contains redhat rpm repositories. It seems that the content of the file
isn't always the same for every possible environment, which might cause
problems with the installation.

This commit guarantee that the `microdnf install` command is consistent
across all environments simply by restraining microdnf to selected ubi
repos.

Signed-off-by: Lukas Krajicek <lkrajice@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN go build\
   ./cmd/clairctl
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal AS final
-RUN microdnf install tar
+RUN microdnf install --disablerepo=* --enablerepo=ubi-8-baseos --enablerepo=ubi-8-appstream tar
 RUN curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 && chmod +x /usr/local/bin/dumb-init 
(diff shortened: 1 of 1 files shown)


lkrajice:
@jvulgan @hdonnay ptal

hdonnay:
I don't understand why this is needed. Why can't we trust however the ubi image is built?

lkrajice:
@hdonnay Basically, you can get into a situation where the microdnf install command fails.

From what I know, in case of the UBI, the package manager always creates `redhat.repo` and, by default, it is empty. Additionally, when the host, you do the container build on, has active RH subscription, the UBI inherits the subscription and base repos from the host. So when we do the `microdnf install` step while building, the microdnf also checks the `redhat.repo` repositories. This might cause problems, since the repositories from the node might require SSL verification for which the UBI image is not setup (this happens in our use-case).

The change in the PR does make sure that the install step is not dependent on your configuration and makes it consistent. Another option solving this would be to not use UBI image.

https://access.redhat.com/solutions/1443553

hdonnay (approved):
It's annoying that Red Hat insists on making its various projects harder to use when used together.

luze560715 (commented):
63936692

PR #1320, opened 2021-07-26
http: rate limit index report requests

crozzy:
Added a basic rate limiter that will keep concurrent connections
limited and return a 429 if they go above.

### DIFF
--- a/config/indexer.go
+++ b/config/indexer.go
@@ -26,6 +26,11 @@ type Indexer struct {
 	// Indexers will index a Manifest's layers concurrently.
 	// This value tunes the number of layers an Indexer will scan in parallel.
 	LayerScanConcurrency int `yaml:"layer_scan_concurrency" json:"layer_scan_concurrency"`
+	// Rate limits the number if index report creation requests.
+	//
+	// Any value below 1 is considered unlimited.
+	// The API will return a 429 status code if concurrency is exceeded.
+	IndexReportRequestConcurrency int `yaml:"index_report_request_concurrency" json:"index_report_request_concurrency"`
 	// A "true" or "false" value
 	//
 	// Whether Indexer nodes handle migrations to their database.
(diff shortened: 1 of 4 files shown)


hdonnay:
There's also the package https://pkg.go.dev/golang.org/x/time/rate, which is getting used in the client rate-limiter.

luze560715 (dismissed):

crozzy:
> There's also the package https://pkg.go.dev/golang.org/x/time/rate, which is getting used in the client rate-limiter.

Yeah, I thought about that, the implementation could look something like this:
```go
type RateLimitMiddleware struct {
	limiter *rate.Limiter
}

func NewRateLimitMiddleware(maxConcurrent int) *RateLimitMiddleware {
	rlm := &RateLimitMiddleware{
		limiter: rate.NewLimiter(rate.Limit(maxConcurrent), maxConcurrent),
	}
	return rlm
}

func (rlm *RateLimitMiddleware) Handler(endpoint string, next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if rlm.limiter == nil {
			next.ServeHTTP(w, r)
			return
		}

		if !rlm.limiter.Allow() {
			rateLimitedCounter.WithLabelValues(endpoint).Add(1)
			resp := &jsonerr.Response{
				Code:    "too-many-requests",
				Message: "server handling too many requests",
			}
			jsonerr.Error(w, resp, http.StatusTooManyRequests)
			return
		}

		next.ServeHTTP(w, r)
	})
}
```
The question is, do we want requests/sec to be limited, or concurrent requests to be limited. I'm not entirely sure.

hdonnay:
Ah, that's a good point; the request counting scheme means we could, in theory, serve more requests in the same amount of time.

crozzy:
@hdonnay I'm taking this out of draft, one thing that is still on the table is either going with the predefined number (like now), or derive the number from the size of the container and the config having an on/off switch instead of a number.

hdonnay (dismissed):
LGTM

Re: sizing: It'd be nice for the default to have some intelligent limit (that an operator can then override by doing >0 or disable entirely by going <0) but I don't have a good idea how, yet. Probably some heuristic with GOMAXPROCS as the input.

hdonnay (approved):

PR #1325, opened 2021-08-05
introspection: capture rate-limited requests

crozzy:
This is a bug, the introspection handler should
always be the outer-most handler or we lose insight
into how many requests were limited.


crozzy:
An example of how the introspection can inform the metrics of rate-limited indexing report requests.

image

hdonnay (approved):
LGTM

PR #1335, opened 2021-08-11
HTTP caching

hdonnay:
@crozzy and I discussed adding caching metadata to vulnerability report responses to reduce load and response churn.
This PR implements that as best as I can tell from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching), and some cleanups I noticed while doing it.

### DIFF
--- a/httptransport/vulnerabilityreporthandler.go
+++ b/httptransport/vulnerabilityreporthandler.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptrace"
 	"path"
+	"time"
 
 	"github.com/quay/claircore"
 	je "github.com/quay/claircore/pkg/jsonerr"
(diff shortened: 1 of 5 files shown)


httptransport/vulnerabilityreporthandler.go
```diff
@@ -16,91 +17,98 @@ import (
...
-				Message: "endpoint only allows GET",
-			}
-			je.Error(w, resp, http.StatusMethodNotAllowed)
-			return
-		}
-		ctx, done := context.WithCancel(r.Context())
-		defer done()
-		ctx = httptrace.WithClientTrace(ctx, oteltrace.NewClientTrace(ctx))
+// VulnerabilityReportHandler produces VulnerabilityReports.
+type VulnerabilityReportHandler struct {
```
crozzy, line 21: Nice, a lot cleaner

crozzy (approved):
This looks great

PR #1346, opened 2021-08-19
matcher: default garbage collection on

### LINKED ISSUE #1337: GC does not run unless "matcher.update_retention" is set explicitly

Backstory: The database of my Clair instance is growing much faster than it used to. Unfortunately, I don't have metrics going back that long, so I cannot pinpoint when this started, but I set it up in February on a 50 GiB volume, and it then hovered around 40 GiB usage for most of that time (after the initial updater runs had completed and all vulnerabilities were entered into the database). In July, I got disk-full alerts and extended from 50 GiB to 100 GiB. This week, just 6 weeks later, the 100 GiB volumes were full yet again and I had to upsize the volumes another time. So while I don't have exact numbers, there is definitely a stronger growth in the last two months than before then. I upgraded from 4.0.x to 4.1.0 at the beginning of June, so there is a timing correlation there, but I'm not able to provide strong evidence one way or the other.

### Expected Outcome

I would expect the DB size to grow slowly, but steadily (as new images get scanned and the updaters pull in new vulnerabilities).

### Actual Outcome

As described above, the database is growing much faster than it used to. I had a look around in psql to see what is using all that storage: The sizes of most tables are rounding errors (including `manifest` and `indexreport` and such, since I only have about 25000 images in the DB at the moment). The `vuln` table takes up ~25% of all storage, at just shy of 10 million entries (which is a lot, but I don't think it's that unusual since I have a lot of updaters enabled). The remaining ~75% are occupied by the `uo_vuln` table:

```
clair=# SELECT COUNT(*) FROM uo_vuln;
   count
-----------
 723471358
(1 row)
```

My understanding of this table is limited by what I can grasp from the database schema, but over 700 million entries seem excessive given that there are just 10 million `vuln` entries and about 30000 `update_operation` entries. There is apparently a ton of duplication going on here. For example, I grabbed one update operation's fingerprint at random and got this:

```
clair=# SELECT (SELECT COUNT(*) FROM uo_vuln WHERE uo = id) AS uo_vuln_row_count, * FROM update_operation WHERE fingerprint = '"206c95-5c39b36cc655d-gzip"';
 uo_vuln_row_count |  id   |                 ref                  |        updater        |         fingerprint         |             date              |     kind
-------------------+-------+--------------------------------------+-----------------------+-----------------------------+-------------------------------+---------------
               946 | 41537 | 243cab3d-397d-4367-9aab-b1058f884562 | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 10:44:59.386241+00 | vulnerability
               946 | 41547 | 1ccfad3e-c2d6-4afe-b309-88fdab6e245f | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 11:44:59.403178+00 | vulnerability
               946 | 41560 | 3dad7bda-c6d9-4993-a450-b8ce7b77ee0f | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 12:44:59.367305+00 | vulnerability
               946 | 41568 | 15d4307d-01ef-4851-9f14-639d217c31a1 | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 13:14:59.441119+00 | vulnerability
               946 | 41543 | 5aba8d2d-acb4-433d-9da0-941bf5ccc89d | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 11:15:28.283515+00 | vulnerability
               946 | 41553 | 3b6db3cf-6b2c-4975-82e3-f84de1d4a7b5 | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 12:14:59.472253+00 | vulnerability
               946 | 41576 | b5f68513-5366-44c5-afd8-917f62b1b2e5 | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 13:44:59.429989+00 | vulnerability
               946 | 41533 | 58362dcb-1f11-433b-9aeb-33b4165f6713 | debian-jessie-updater | "206c95-5c39b36cc655d-gzip" | 2021-05-31 10:14:28.178319+00 | vulnerability
(8 rows)
```

Those are 8 update operations with the same fingerprint, and for each of them, we have 946 entries in the `uo_vuln` table. That leads me to the following questions:

1. Is that kind of growth in the `update_operation` and `uo_vuln` tables expected by design, or the symptom of a bug?
2. Regardless of whether it's by design or a bug, what would be a safe method for pruning those tables of old entries?

### Environment

- Clair version/image: 4.1.1 (self-compiled from release tag on Alpine 3.13 with Go 1.16)
- Clair client name/version: N/A
- Host OS: Alpine in container, CoreOS on host
- Kernel (e.g. `uname -a`): `Linux $HOSTNAME 4.19.123-coreos #1 SMP Fri May 22 19:21:11 -00 2020 x86_64 Linux`
- Kubernetes version (use `kubectl version`): N/A
- Network/Firewall setup: N/A

crozzy:
Change the GC logic to reflect the docs and
also update the docs with more details.
```
< 0 == GC off
0 or unspecified == GC on w/ default retention
< 2 == GC on w/ default retention
>= 2 == GC on w/ specified value
```

### DIFF
--- a/config/matcher.go
+++ b/config/matcher.go
@@ -65,7 +65,12 @@ func (m *Matcher) Validate(combo bool) error {
 	if m.Period == 0 {
 		m.Period = DefaultPeriod
 	}
-	if m.UpdateRetention == 1 || m.UpdateRetention < 0 {
+	switch {
+	case m.UpdateRetention < 0:
+		// Less than 0 means GC is off.
+		m.UpdateRetention = 0
+	case m.UpdateRetention < 2:
+		// Anything less than 2 gets the default.
 		m.UpdateRetention = DefaultRetention
 	}
 	if m.CacheAge == 0 {
--- a/Documentation/reference/config.md
+++ b/Documentation/reference/config.md
@@ -216,7 +216,8 @@ constraints.
 
 Defaults to 10.
 
-If a value of 0 is provided, GC is disabled.
+If a value less than 0 is provided, GC is disabled. 2 is the minimum value to
+ensure updates can be compared for notifications. 
 
 ### `$.matchers`
 Matchers provides configuration for the in-tree Matchers and RemoteMatchers.
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -70,6 +70,102 @@ func Test_Config_Validate_Failure(t *testing.T) {
 	}
 }
 
+func TestConfigUpateRetention(t *testing.T) {
+	var table = []struct {
+		name              string
+		conf              config.Config
+		expectedRetention int
+	}{
+		{
+			name:              "Retention less than 0",
+			expectedRetention: 0,
+			conf: config.Config{
+				Mode:           config.ComboMode,
+				HTTPListenAddr: "localhost:8080",
+				Indexer: config.Indexer{
+					ConnString: "example@example/db",
+				},
+				Notifier: config.Notifier{
+					ConnString: "example@example/db",
+				},
+				Matcher: config.Matcher{
+					ConnString:      "example@example/db",
+					IndexerAddr:     "example@example/db",
+					UpdateRetention: -1,
+				},
+			},
+		},
+		{
+			name:              "Retention of 0",
+			expectedRetention: 10,
+			conf: config.Config{
+				Mode:           config.ComboMode,
+				HTTPListenAddr: "localhost:8080",
+				Indexer: config.Indexer{
+					ConnString: "example@example/db",
+				},
+				Notifier: config.Notifier{
+					ConnString: "example@example/db",
+				},
+				Matcher: config.Matcher{
+					ConnString:      "example@example/db",
+					IndexerAddr:     "example@example/db",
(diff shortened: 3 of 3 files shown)


config/matcher.go
```diff
@@ -65,7 +65,12 @@ func (m *Matcher) Validate(combo bool) error {
 	if m.Period == 0 {
 		m.Period = DefaultPeriod
 	}
-	if m.UpdateRetention == 1 || m.UpdateRetention < 0 {
+	switch {
```
crozzy, line 68: @hdonnay I know you love switches :D

hdonnay: A `switch` is just a generalized `if` 😎

hdonnay (approved):
LGTM

PR #1349, opened 2021-08-30
integrate ctxlock and additional changes from the livelock investigation

hdonnay:
This migrates off of distlock, which is removed in claircore.


hdonnay:
Note: this is going to fail until Claircore with ctxlock is merged and released.

notifier/service/service.go
```diff
@@ -237,20 +246,25 @@ func amqpDeliveries(ctx context.Context, opts Opts, lockPool *pgxpool.Pool, stor
...
+			delivery := notifier.NewDelivery(i, q, opts.DeliveryInterval, store, l)
 			ds = append(ds, delivery)
 		} else {
 			q, err := namqp.New(conf)
 			if err != nil {
 				return fmt.Errorf("failed to create AMQP deliverer: %v", err)
 			}
-			delivery := notifier.NewDelivery(i, q, opts.DeliveryInterval, store, distLock)
+			delivery := notifier.NewDelivery(i, q, opts.DeliveryInterval, store, l)
 			ds = append(ds, delivery)
```
crozzy, line 268: Not that it's directly related to this PR but do you know why it's appending to a slice and then iterating through it is preferable? I think trusting the `Deliver()` method to close the lock seems a bit fragile, and closing it directly in this function is a little nasty due to the double iteration (hence why I was wondering the initial thing)

crozzy (commented):
Just one comment and tangentially related to this PR

notifier/service/service.go
hdonnay: I'm not sure, I think it was to initialize everything, and then start doing work?

But this whole subsystem is a little under-loved and uses some non-idiomatic go -- like functions that immediately spawn goroutines and return -- so the answer to "why is this like this?" in the `notifier` directory is almost certainly "inertia."

crozzy: Makes sense, when giving it a little love, we might also want to discuss the location of the notifier logic that might also lead the merging repos chat (clair+claircore).

crozzy (approved):

hdonnay:
Arg, typo'd a command. This is merged, though.

PR #1350, opened 2021-08-31
Clairctl updates

hdonnay:
This is a raft of changes I made while debugging the claircore livelock issue.

### DIFF
--- a/cmd/clairctl/manifest.go
+++ b/cmd/clairctl/manifest.go
@@ -74,7 +74,7 @@ func manifestAction(c *cli.Context) error {
 }
 
 func Inspect(ctx context.Context, r string) (*claircore.Manifest, error) {
-	rt, err := rt(r)
+	rt, err := rt(ctx, r)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/clairctl/report.go
+++ b/cmd/clairctl/report.go
@@ -14,6 +14,8 @@ import (
 	"github.com/quay/claircore"
 	"github.com/quay/zlog"
 	"github.com/urfave/cli/v2"
+	"go.opentelemetry.io/otel/baggage"
+	"go.opentelemetry.io/otel/label"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/quay/clair/v4/internal/codec"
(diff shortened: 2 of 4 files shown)


cmd/clairctl/manifest.go
```diff
@@ -78,13 +81,38 @@ func Inspect(ctx context.Context, r string) (*claircore.Manifest, error) {
...
 	}
+	ct := 1 // Track retries, also used for retry backoff.
+	var temporary temporary
+	txErr := new(transport.Error)
 
 	ref, err := name.ParseReference(r)
 	if err != nil {
 		return nil, err
 	}
+Retry:
```
crozzy: https://github.com/google/go-containerregistry/blob/7e0ed51a7bb1930bfb6859f319975b578ed04de2/pkg/v1/remote/transport/retry.go#L57 Did you see this? Not sure if drops in

cmd/clairctl/report.go
```diff
@@ -150,21 +164,19 @@ func reportAction(c *cli.Context) error {
 
 	for i := 0; i < args.Len(); i++ {
 		ref := args.Get(i)
+		ctx := baggage.ContextWithValues(ctx, label.String("ref", ref))
 		zlog.Debug(ctx).
```
crozzy, line 169: Probably makes sense as a one-liner now

cmd/clairctl/report.go
```diff
@@ -150,21 +164,19 @@ func reportAction(c *cli.Context) error {
...
 			d, err := resolveRef(ref)
 			if err != nil {
 				zlog.Debug(ctx).
-					Str("ref", ref).
 					Err(err).
 					Send()
 				return err
 			}
+			ctx := baggage.ContextWithValues(ctx, label.String("digest", d.String()))
 			zlog.Debug(ctx).
```
crozzy, line 180: Probably makes sense as a one-liner now

cmd/clairctl/report.go
```diff
@@ -174,32 +186,55 @@ func reportAction(c *cli.Context) error {
...
+			}
+			zlog.Debug(ctx).
+				Int("attempt", ct).
+				Msg("requesting index_report")
 			err = cc.IndexReport(ctx, d, m)
 			switch {
 			case err == nil:
 			case errors.Is(err, errNeedManifest):
+				if c.Bool("novel") {
+					break
```
crozzy, line 205: Might be nice to debug explain why we're breaking here so the user has an idea that they affected the action

cmd/clairctl/report.go
```diff
@@ -174,32 +186,55 @@ func reportAction(c *cli.Context) error {
 			//
 			// If we need the manifest, populate the manifest and jump to Again.
 			var m *claircore.Manifest
+			ct := 1
 		Again:
+			if ct > 20 {
```
crozzy, line 192: What are you thoughts on the retry logic being encapsulated in the client? Feels like the client could eventually become it's own thing and the retry logic would be nice to have by default.

crozzy:
Overall, super guay :+1:

cmd/clairctl/report.go
hdonnay: I think this is a good eventual goal, but since the client is internal right now, I'm not too worried about the occasional layering leak like this one.

crozzy (approved):

PR #1360, opened 2021-09-09
cicd: add script to edit module for nightly

hdonnay:
This script should automate editing the clair module so that it builds
against claircore's main branch. The idea is to run this in a CI job,
build the resulting code, and then throw away the modifications.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM quay.io/projectquay/golang:1.16 AS build
+ARG GO_VERSION=1.17
+FROM quay.io/projectquay/golang:${GO_VERSION} AS build
 WORKDIR /build/
 ADD . /build/
 ARG CLAIR_VERSION=dev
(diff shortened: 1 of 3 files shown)


hdonnay:
You can see some successful runs [here](https://github.com/hdonnay/clair/actions/workflows/nightly.yml).

Dockerfile
```diff
@@ -24,7 +25,12 @@ RUN go build\
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal AS final
 RUN microdnf install --disablerepo=* --enablerepo=ubi-8-baseos --enablerepo=ubi-8-appstream tar
-RUN curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 && chmod +x /usr/local/bin/dumb-init 
+RUN case "$(uname -m)" in \
```
crozzy, line 28: this is a little bit of a shame, but can't think of a more graceful way

crozzy (approved):
awesome, updating my cluster images...

Dockerfile
hdonnay: Yeah, I don't know when the syntax/backend updates that allow the "buildx" extra variables will be usable.

crozzy: ...or rhel pick up / yelp publish rpms and microdnf would take care of the arch parsing

hdonnay: Yeah, I don't recall if there's an intent to package `dumb-init,` and I certainly don't want to be on the hook for an `rpm`.

The other options are:

1. write a little bit of code to catch `SIGCHLD` and ditch `dumb-init`
2. remove the exec calls that necessitate pid 1 handling (currently just `rpm`, I think)

PR #1366, opened 2021-09-14
cicd: use push flag with build

hdonnay:
This is how the other buildx usages are written.


hdonnay:
It'd be nice to be able to use the build cache inside the docker container, but I haven't figured that out yet.

We could also use the buildx caching (probably to ghcr) to speed up builds.

crozzy (approved):

PR #1369, opened 2021-09-16
CI updates

hdonnay:
These CI updates factor out some common actions, and add a lot of caching.


crozzy (approved):

hdonnay:
This is merged, github raced with itself.

PR #1385, opened 2021-09-29
Dockerfile: remove dumb-init, tar

hdonnay:
This change removes tar from the final image (no longer used by
claircore) and replaces dumb-init with a copy of catatonit from podman.

More info about catatonit can be found at
https://github.com/openSUSE/catatonit/.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,15 +23,11 @@ RUN go build \
 RUN go build\
   ./cmd/clairctl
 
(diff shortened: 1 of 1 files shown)


crozzy (approved):
:clap:

hdonnay:
Just double-checked that this package exists on `aarch64`, so :+1:

PR #1388, opened 2021-09-30
chore: build unstable tag for redhat-X.Y branches (PROJQUAY-2556)

flavianmissi:
quay-operator e2e workflow expects clair to provide a X.Y-unstable tag.
i.e 3.5-unstable

### DIFF
--- a/.github/workflows/build-and-publish.yaml
+++ b/.github/workflows/build-and-publish.yaml
@@ -1,0 +1,44 @@
+---
+name: Build and Publish
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


hdonnay:
I'm in the process of removing these branches; could the e2e tests use our nightly tag?

flavianmissi:
> I'm in the process of removing these branches; could the e2e tests use our nightly tag?

Alright - I wasn't aware of the deprecation of these branches. Will take your advice and use the night for now then, thx!

PR #1390, opened 2021-10-06
httptransport: simpler codegen for openapi JSON

hdonnay:
This produces the same output, in a much easier to understand way.


crozzy (approved):

hdonnay:
This is merged. I guess I need to add some sleeps to my script <_<

PR #1397, opened 2021-10-14
Dev environment changes

hdonnay:
While investigating PROJQUAY-2152, I made a lot of changes.

This is both the jig I built to verify the notifier, and the changes to the dev environment to be less dependent on Docker specifically and utilize `docker-compose`'s native profile support.


hdonnay:
TODO:

- [x] Verify the Quay changes (blocked on containers/podman#11966 being released)

hdonnay:
containers/podman#11966 is released in podman 3.4.1, so circling back to check on quay.

Documentation/howto/testing.md
```diff
@@ -29,72 +32,43 @@ See [Install Golang](https://golang.org/doc/install)
...
 git clone git@github.com:quay/clair.git
 cd clair
-make local-dev-up-with-quay
+make local-dev-quay
 ```
 
 After the local development environment successfully starts, the following infrastructure is available to you:
 
-```
-localhost:8080 --- Quay (single node, local storage)
```
crozzy, line 38: Definitely an improvement here, although can we find somewhere for the credentials to avoid browsing / execing around for the defaults?

Makefile
```diff
@@ -14,143 +14,43 @@
...
+local-dev: vendor
+	$(docker-compose) up -d
+
+compose_profiles = $(patsubst %,local-dev-%,quay notifier debug)
+.PHONY: $(compose_profiles)
+local-dev-%: vendor
+	$(docker-compose) --profile $* up -d
+
+local-dev-quay: local-dev/clair/quay.yaml vendor
+	CLAIR_CONFIG=$(<F) $(docker-compose) --profile quay up -d
```
crozzy, line 48: you think cleaning up the `local-dev/clair/quay.yaml` file here is a bit overreaching?

crozzy (dismissed):

Makefile
hdonnay: What do you mean "cleaning up?" It needs to exist while things are running, and needs to be slightly different from the standard config. It being a proper target and dependency at least makes sure it gets updated as the standard config changes.

crozzy (approved):

PR #1398, opened 2021-10-19
contrib: add grafana contribMap

crozzy:
This dashboard will be pulled and deployed by app-interface
to their grafana instances.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,15 @@ docs-build:
 	rsync --recursive --delete-after --exclude 'v4.*' --exclude .git\
 		./book/ $(DOCS_DIR)/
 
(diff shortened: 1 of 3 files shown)


hdonnay (approved):
This could be a real target instead of a phony one, but I don't really know what a good over-arching target to run would be.

crozzy:
> This could be a real target instead of a phony one, but I don't really know what a good over-arching target to run would be.

This is interesting, I have always just been under the impression that if the target isn't actually the name of a file you are creating then use .PHONY, I'm not sure if that file already exists if it'll recreate it. Reading more about .PHONY is seems there are more nuanced cases.

PR #1400, opened 2021-10-20
Feature/config

hdonnay:
This is a rework of the config package to have it split into a distinct module with minimal dependencies.
I added a linter because it stuck in my mind after I realized it could be done once all the types are in the same package.

This required pretty extensive surgery to the notifier's types, as the config referenced those directly.

Outstanding questions before the final version:

- Should we break the config package code API more? 
  Namely, rework the `Validate` method. `Lint` could become a replacement for `Validate` and return errors in same/similar situations. Perhaps unexport the `Mode` member and add setter/getters for it.
- How can we evolve defaults?
  Is that even possible/desirable when a user renders out a config and feeds it back?


cmd/clair/values.go
```diff
@@ -37,55 +37,33 @@ func (v *ConfValue) Set(s string) error {
...
-		*v = ModeCombo
-	case "index", "indexer":
-		*v = ModeIndexer
-	case "match", "matcher":
-		*v = ModeMatcher
-	case "notify", "notifier":
-		*v = ModeNotifier
-	default:
-		return fmt.Errorf("unknown mode argument %q", s)
+	case "", "dev", "pizza", "combination":
```
crozzy: you don't wanna keep combo?

cmd/clair/values.go
```diff
@@ -37,55 +37,33 @@ func (v *ConfValue) Set(s string) error {
...
+		// "Pizza", of course, being the best Combos flavor.
+		s = "combo"
+	case "index":
+		s = "indexer"
+	case "match":
+		s = "matcher"
+	case "notify":
+		s = "notifier"
```
crozzy: I kind of think we should just assign the mode here and this effectively becomes `ParseMode`, I get the case-insensitivity, but we're going to document one way, so adding another way might confuse things (I do enjoy pizza though).

config/tls.go
```diff
@@ -0,0 +1,86 @@
...
+	// The filesystem path where a root CA can be read.
+	//
+	// This can also be controlled by the SSL_CERT_FILE and SSL_CERT_DIR
+	// environment variables, or adding the relevant certs to the system trust
+	// store.
+	RootCA string `yaml:"root_ca" json:"root_ca"`
+	// The filesystem path where a tls certificate can be read.
+	Cert string `yaml:"cert" json:"cert"`
+	// The filesystem path where a tls private key can be read.
+	Key string `yaml:"key" json:"key"`
```
crozzy, line 26: I think maybe these should be suffixed with `_path` or `_file`

notifier/amqp/config.go
```diff
@@ -3,135 +3,80 @@ package amqp
...
+	AutoDelete bool
 }
 
-// Validate confirms configuration is valid and fills in private members
-// with parsed values on success.
-func (c *Config) Validate() (Config, error) {
-	conf := *c
-	if c.Exchange.Type == "" {
-		return conf, fmt.Errorf("AMQP config requires the exchange.type field")
+func loadTLSConfig(c *config.AMQP) (*tls.Config, error) {
```
crozzy: What are your thoughts on `loadTLSConfig` being generic for all messaging backends and the check to see whether to load TLS is done before calling this function. Just thinking we might want to extend this functionality to layer fetching and potentially vulnerability fetching. Just a thought

crozzy (commented):
Wow this got deep, I know it's in draft was just interested in taking a look. I'll review again when finished

config/tls.go
hdonnay: I think we have to leave it until a v2, as it'll change the on-disk representation.

cmd/clair/values.go
hdonnay: My preference, then, would be to just delegate to the `ParseMode` function.

cmd/clair/values.go
hdonnay: it's still in there, just implicitly via the ParseMode call

hdonnay:
For my questions:

- I broke the API a bit more. I think the better output around potential issues with the config is worth it.
- Punting on the defaults question. I think we have to accept that if you render out a config object, it has the defaults from that point in time.

cmd/clairctl/export.go
```diff
@@ -62,11 +63,18 @@ func exportAction(c *cli.Context) error {
 	}
 	cfgs := make(map[string]driver.ConfigUnmarshaler, len(cfg.Updaters.Config))
 	for name, node := range cfg.Updaters.Config {
-		cfgs[name] = node.Decode
+		node := node // Unsure if needed.
```
crozzy: Difficult one, I believe yes as when the function is evaluated `n` will already be at it's last value unless copied https://play.golang.org/p/74vXJ_Xx2qI

config/config.go
```diff
@@ -2,32 +2,19 @@ package config
...
-
+// Config is the configuration object for the commands in
+// github.com/quay/clair/v4/cmd/...
 type Config struct {
 	// One of the following strings
 	// Sets which mode the clair instances will run in
 	//
 	// "indexer": runs just the indexer node
 	// "matcher": runs just the matcher node
 	// "combo":	will run both indexer and matcher on the same node.
```
crozzy: Poor notifier, always the black sheep

config/defaults.go
```diff
@@ -0,0 +1,29 @@
...
+	// DefaultAddress is used if an "http_listen_addr" is not provided in the config.
+	DefaultAddress = ":6060"
+	// DefaultScanLockRetry is the default retry period for attempting locks
+	// during the indexing process. Its name is a historical accident.
+	DefaultScanLockRetry = 1
+	// DefaultMatcherPeriod is the default interval for running updaters.
+	DefaultMatcherPeriod = 30 * time.Minute
+	// DefaultMatcherRetention is the number of updates per vulnerability
+	// database to retain.
+	DefaultMatcherRetention = 10
```
crozzy: Thoughts on this name, I think `DefaultUpdateRetention` might be closer to the actual function.

notifier/amqp/config.go
```diff
@@ -1,137 +1,33 @@
...
-		conf.tls = &TLS
-	}
-
-	if !c.Direct {
-		callback, err := url.Parse(c.Callback)
-		if err != nil {
-			return conf, fmt.Errorf("failed to parse callback url")
-		}
-		conf.callback = *callback
+func exchangeFrom(c *config.AMQP) exchange {
```
crozzy: What is the point of doing this and not assigning directly?
```go
d.exchange = conf.Exchange
```

notifier/processor_create_test.go
```diff
@@ -79,7 +80,7 @@ func TestProcessCreate(t *testing.T) {
 // available
 func testProcessorStoreErr(t *testing.T) {
 	t.Parallel()
-	ctx := context.TODO()
+	ctx := zlog.Test(context.Background(), t)
```
crozzy, line 83: Kinda neat idea

notifier/stomp/config.go
```diff
@@ -3,95 +3,44 @@ package stomp
...
+)
 
-	if !c.Direct {
-		var u *url.URL
-		var err error
-		if u, err = url.Parse(c.Callback); err != nil {
-			return conf, fmt.Errorf("direct delivery is disabled but callback url could not be parsed.")
-		}
-		conf.callback = *u
+func loadTLSConfig(c *config.STOMP) (*tls.Config, error) {
```
crozzy: I can't tell where this is being called now

crozzy (changes requested):
Really great improvement, just a few comments

notifier/stomp/config.go
hdonnay: Might be left over, I'll double-check

notifier/amqp/config.go
hdonnay: This way, nothing's holding onto pointers back to the config object.

crozzy: Totally agree with the intent but as I see it `Exchange` in the config isn't passed by it's reference, so assigning it directly to a `Deliverer.Exchange` field would mean the values are copied over to a new object. Any mutation of `config.Exchange` or `Deliverer.Exchange` should not affect each other I believe.

crozzy: (leaving aside the whole "nothing is passed by reference" debate)

crozzy (approved):
LGTM!

PR #1427, opened 2021-11-11
docs: add history preprocessor directive

hdonnay:


hdonnay:
I was reading a post on [architecture decision records](https://andydote.co.uk/2019/06/29/architecture-decision-records/), and it pointed out that when processing documentation, one can leverage the vcs to add in historical information.

This adds a tool that slip-streams the history of a file into the rendered output.

PR #1436, opened 2021-11-18
local-dev: deterministic postgres port

crozzy:
It's nice if you're using database tool to have a
predetermined port to avoid having to change
connection configs.


crozzy:
Also open to another port if contention is an issue

PR #1442, opened 2021-11-22
contrib: add initContainer to wipe VPC on startup

crozzy:
In unexpected situations the indexer can crash and
not clean up it's layers. When using a statefulSet this
means those layers are always hanging around. This adds
an initContainer to flush the VPC before starting the
indexer pod.

### DIFF
--- a/contrib/openshift/manifests/manifests.yaml
+++ b/contrib/openshift/manifests/manifests.yaml
@@ -81,6 +81,14 @@ objects:
                   mountPath: /etc/clair
                 - name: indexer-layer-storage
                   mountPath: /tmp
+          initContainers:
+            - name: init-wipe-vpc
+              image: ${{UBI_IMAGE}}:${{UBI_IMAGE_TAG}}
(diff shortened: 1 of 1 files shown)


contrib/openshift/manifests/manifests.yaml
```diff
@@ -81,6 +81,14 @@ objects:
                   mountPath: /etc/clair
                 - name: indexer-layer-storage
                   mountPath: /tmp
+          initContainers:
+            - name: init-wipe-vpc
+              image: ${{UBI_IMAGE}}:${{UBI_IMAGE_TAG}}
+              command: ['sh', '-c', "rm -rf /tmp/*"]
```
crozzy, line 87: I was between this and adding the layer prefix. Thoughts?

hdonnay: I'd maybe mount it somewhere else for the delete if you're not sure about the glob. Seems fine to me, though.

hdonnay (approved):

PR #1449, opened 2021-12-01
httptransport: fix content negotiation

### LINKED ISSUE #1441: Discovery handler does not handle accept header correctly
I'm trying to load /openapi/v1 in swagger ui, but this fails with an error message of `endpoint only allows application/json or application/vnd.oai.openapi+json`.

Swagger uses a GET request with an Accept header of `application/json, */*` which should be valid.
Manually trying it with `application/json` or `*/*` works.
Looking at the code it seems that it tries to match the full header instead of treating it as a comma separated list.

### Expected Outcome

It should return the openapi spec with a status code of 200

### Actual Outcome

An error message of `endpoint only allows application/json or application/vnd.oai.openapi+json` with a status code of 400.

### Environment

- Clair version/image: 4.3.5
- Clair client name/version: n/a
- Host OS: n/a
- Kernel (e.g. `uname -a`): n/a
- Network/Firewall setup: n/a

hdonnay:
This fixes the handling of the "accept" header to process lists and "q"
parameters properly, and moves it to a common helper function.

Closes: #1441

### DIFF
--- a/httptransport/common.go
+++ b/httptransport/common.go
@@ -1,0 +1,96 @@
+package httptransport
+
+import (
+	"errors"
+	"fmt"
+	"mime"
+	"net/http"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// PickContentType sets the response's "Content-Type" header.
+//
+// If "Accept" headers are not present in the request, the first element of the
+// "allow" slice is used.
+//
+// If "Accept" headers are present, the first (ordered by "q" value) media type
+// in the "allow" slice is chosen. If there are no common media types, "415
+// Unsupported Media Type" is written and ErrMediaType is reported.
+func pickContentType(w http.ResponseWriter, r *http.Request, allow []string) error {
+	// There's no canonical algorithm for this, it's all server-dependent
+	// behavior. Our algorithm is:
+	//
+	//	- Parse the Accept header(s) as MIME media types joined by commas.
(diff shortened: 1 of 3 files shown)


httptransport/common.go
```diff
@@ -0,0 +1,96 @@
...
+	//	- Pick the first match.
+	//
+	// BUG(hank) Content type negotiation does an O(n*m) comparison driven on
+	// user input, which may be a DoS issue.
+	as, ok := r.Header["Accept"]
+	if !ok {
+		w.Header().Set("content-type", allow[0])
+		return nil
+	}
+	var client []accept
```
crozzy: For me this var name was a little confusing

crozzy (changes requested):
LGTM just one comment

httptransport/common.go
```diff
@@ -0,0 +1,96 @@
...
+	//	- Pick the first match.
+	//
+	// BUG(hank) Content type negotiation does an O(n*m) comparison driven on
+	// user input, which may be a DoS issue.
+	as, ok := r.Header["Accept"]
+	if !ok {
+		w.Header().Set("content-type", allow[0])
+		return nil
+	}
+	var acceptable []accept
```
crozzy, line 37: :smile:

crozzy (approved):

PR #1461, opened 2021-12-17
auto: fall back to root cgroup

hdonnay:
It seems that some container runtimes mount things such that paths
reported from procfs don't line up in sysfs. This falls back to
examining the root cgroup if a specific one doesn't exist; as noted in
the code, if this isn't actually the process' cgroup, it'll be the root
cgroup and end up matching the default GOMAXPROCS guess.


crozzy (approved):
Seems good according to what the node guys are suggesting :+1:

PR #1471, opened 2022-01-12
Manifest Delete

hdonnay:
These changes specify and add the API for deleting manifests, along with some helpers picked up along the way.

Needs a claircore release containing the relevant API.


httptransport/indexer_v1.go
```diff
@@ -0,0 +1,344 @@
...
+		defer writerError(w, &err)()
+		w.WriteHeader(http.StatusOK)
+		enc := codec.GetEncoder(w)
+		defer codec.PutEncoder(enc)
+		err = enc.Encode(ds)
+	}
+}
+
+const (
+	linkIndex  = `<%s>; rel="https://projectquay.io/clair/v1/index_report"`
```
crozzy, line 150: This is a nice addition but should the `rel` be `next` or something?

httptransport/indexer_v1.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+
+		w.Header().Set("etag", validator)
+		w.Header().Set("location", next)
+		defer writerError(w, &err)()
+		w.WriteHeader(http.StatusCreated)
+		enc := codec.GetEncoder(w)
+		defer codec.PutEncoder(enc)
+		err = enc.Encode(report)
+	case http.MethodDelete:
```
crozzy, line 127: What do you think about the batch delete being on a dedicated endpoint instead of co-locating it with the creation of one indexreport? For me these batch deletes always fall out of "simple design", but the two different request payloads might be confusing.

httptransport/indexer_v1.go
```diff
@@ -0,0 +1,344 @@
...
+	defer codec.PutEncoder(enc)
+	err = enc.Encode(affected)
+}
+
+func init() {
+	indexerv1wrapper.init()
+}
+
+var indexerv1wrapper = &wrapper{
+	RequestCount: prometheus.NewCounterVec(
```
crozzy, line 301: We should update the grafana dashboards, I think these metric names are different,

httptransport/indexer_v1.go
hdonnay: No, it's the relation type: https://www.rfc-editor.org/rfc/rfc5988#section-5.3

httptransport/indexer_v1.go
hdonnay: Yeah, I changed them to be in-line with what I could tell the best practices are.

httptransport/indexer_v1.go
hdonnay: I don't think that a different method having different arguments is that surprising. For better or worse, the HTTP layer is REST-y and having an imperative-named endpoint would be a significant divergence.

httptransport/indexer_v1.go
crozzy: Oh I interpreted that as if you use the `link-extension` then it had to be an absolute URI otherwise it should describe the relationship between the current resource and the link in the header, I was looking at this: https://datatracker.ietf.org/doc/html/rfc2068#section-19.6.2.4

httptransport/indexer_v1.go
crozzy: It looks like the 429s aren't caught with this metric as the code will return before getting into these handlers, is the idea to track them with `concurrencylimited_total`, if so I'll add a TODO to add some more charts and and SOP for excessive 429 rates

httptransport/server.go
```diff
@@ -161,26 +161,28 @@ func (t *Server) configureComboMode(ctx context.Context) error {
...
-		intromw.InstrumentedHandler(IndexReportAPIPath+"GET", t.traceOpt, IndexReportHandler(t.indexer)))
-
-	t.Handle(IndexStateAPIPath,
-		intromw.InstrumentedHandler(IndexStateAPIPath, t.traceOpt, IndexStateHandler(t.indexer)))
-
+	v1, err := NewIndexerV1(ctx, prefix, t.indexer, t.traceOpt)
+	if err != nil {
+		return fmt.Errorf("indexer configuration: %w", err)
+	}
+	sem := semaphore.NewWeighted(int64(t.conf.Indexer.IndexReportRequestConcurrency))
```
crozzy: I think if the user leaves `IndexReportRequestConcurrency` undefined in the config every request will be 429ed as opposed to none of them, if I'm reading correctly

crozzy (commented):
I tested locally and added some more comments

httptransport/indexer_v1.go
hdonnay: I guess we can embed a semaphore in the handler. Whichever way you'd rather it work.

crozzy: I'm fine as it is, just making sure I understand as 429s can be used to trigger scaling up the indexers in prod environments

crozzy (approved):

PR #1472, opened 2022-01-14
webhook: clone headers out of Config struct

hdonnay:
A re-org in main fixed this misuse of the API, this implements an
equivalent change.

Closes: PROJQUAY-3044


crozzy (dismissed):

hdonnay:
merging via superpowers

PR #1475, opened 2022-01-17
config: auto-size concurrent request capacity

hdonnay:
This just uses a multiple of available cores.

Closes: PROJQUAY-2330

### DIFF
--- a/config/indexer.go
+++ b/config/indexer.go
@@ -1,5 +1,7 @@
 package config
 
+import "runtime"
+
 // Indexer provides Clair Indexer node configuration
 type Indexer struct {
 	// Scanner allows for passing configuration options to layer scanners.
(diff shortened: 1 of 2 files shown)


config/indexer.go
```diff
@@ -75,7 +87,7 @@ func (i *Indexer) lint() (ws []Warning, err error) {
 			msg:  `large values may exceed resource quotas`,
 		})
 	}
-	if i.IndexReportRequestConcurrency < 1 {
+	if i.IndexReportRequestConcurrency < 0 {
 		// Remove this lint if we come up with a heuristic instead of just
```
crozzy: We might want to update this comment as people will be actively requesting unlimited

config/indexer.go
```diff
@@ -24,7 +26,10 @@ type Indexer struct {
 	LayerScanConcurrency int `yaml:"layer_scan_concurrency" json:"layer_scan_concurrency"`
 	// Rate limits the number if index report creation requests.
 	//
-	// Any value below 1 is considered unlimited.
+	// Setting this to 0 will attempt to auto-size this value. Setting a
```
crozzy, line 29: Do you think we should set this value to -1 in the config created by the operator or adopt a new default behaviour? I'm in favor of the 2nd but we might need to flag it to support to avoid cases being created.

hdonnay: I think we want the default to be auto-sized, but let's ask the operator team on their thoughts.

hdonnay: @jonathankingfc said on IM that "auto" is probably what we want.

crozzy (approved):

PR #1479, opened 2022-01-25
Connection pooling control

hdonnay:
This, when claircore is updated, should bring all the connection strings in line so they're handled the same way: via the connection string.

### DIFF
--- a/Documentation/reference/config.md
+++ b/Documentation/reference/config.md
@@ -62,7 +62,6 @@ indexer:
     airgap: false
 matcher:
     connstring: ""
-    max_conn_pool: 0
     indexer_addr: ""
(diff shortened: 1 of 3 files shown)


crozzy (approved):

ricardomaraschini (approved):
Thanks for this @hdonnay ! It looks good. I understand that this is in a per deployment basis but are you able to provide us with some decent default values for the pools ?

hdonnay:
Looks like it's based on the number of cores the program can see: https://github.com/jackc/pgx/blob/v4.14.1/pgxpool/pool.go#L284-L287

That probably means the maximum pool size is way too big in k8s environments.

PR #1481, opened 2022-01-27
Dockerfile: add plugin support files

hdonnay:
This updates the Dockerfile to use the "-trimpath" flag to facilitate
more repeatable builds, with an eye to supporting plugin loading.
Another "builder" Dockerfile is added to use as a "builder"-style base.


hdonnay:
Not going to do this.

PR #1484, opened 2022-02-03
Notifier database refactor

hdonnay:
In the course of investigating a notifier database problem, I ended up refactoring it.


hdonnay:
This patchset grew a bit.

It changes a handful of interrelated things:

1. It refactors the database interface to use common code where appropriate.
   - It adds some metrics in the process
   - It adds a raft of helper functions in the process
2. It removes the single notification synchronous delete method (which was not used anywhere) with a general garbage collection method meant to be called periodically.
   - Cascading restraints are added to various foreign keys as a part of this
     - The SQL migrations are moved to flat-files and `embed`, since they're being touched.
   - A hard-coded limit for how long a notification hangs around is added.
3. The various notifier processes are switched to using caller-managed concurrency so that they can be managed in one place.
   - Some layering violations with the database are fixed because of this.
4. The various callsites are updated to use new API.

hdonnay:
Requesting review from @kwestpharedhat specifically for if the new migration should be broken apart.

notifier/postgres/notifications.go
```diff
@@ -19,102 +24,353 @@ var (
...
+		defer timer.ObserveDuration()
+		var rows pgx.Rows
+		rows, err = c.Query(ctx, pagedQuery, id, page.Next, limit)
+		notificationsCounter.WithLabelValues(`pagedQuery`, errStr(err)).Add(1)
 		for rows.Next() {
-			var n notifier.Notification
-			err := rows.Scan(&n.ID, &n)
-			if err != nil {
-				return nil, p, clairerror.ErrBadNotification{id, err}
+			ns = append(ns, notifier.Notification{})
```
crozzy, line 165: I'm sure there's a good reason for this, avoiding two allocations?

notifier/postgres/notifications.go
```diff
@@ -19,102 +24,353 @@ var (
...
+	return nil
+}
+
+func (r *notificationSource) Columns() []string {
+	return []string{"id", "notification_id", "body"}
+}
+
+// CollectNotifications garbage collects all notifications.
+//
+// Normally Receipter.SetDeleted will be issues first, however application logic
```
crozzy: spelling nit `issued`?

notifier/postgres/notifications.go
```diff
@@ -19,102 +24,353 @@ var (
...
+// may decide to gc notifications which have not been set deleted after some
+// period of time, thus this condition should not be checked.
+func (s *Store) CollectNotifications(ctx context.Context) error {
+	ctx = baggage.ContextWithValues(ctx,
+		label.String("component", "notifier/postgres/Store.CollectNotifications"),
+	)
+	const (
+		tryLock            = `SELECT pg_try_advisory_lock($1, $2);`
```
crozzy: Is this a good use-case for `pg_try_advisory_xact_lock`?

notifier/postgres/notifications.go
```diff
@@ -19,102 +24,353 @@ var (
...
+	metrics := statusMetrics{
+		dur:      gcNotificationDuration,
+		counter:  gcNotificationCounter,
+		affected: gcNotificationAffected,
 	}
 
-	return notifications, outPage, nil
+	err := s.pool.BeginTxFunc(ctx, txOpt, func(tx pgx.Tx) error {
+		var ok bool
+		if err := tx.QueryRow(ctx, tryLock, adminKeyspace, gcLock).Scan(&ok); err != nil {
```
crozzy, line 351: Does this get released implicitly somehow?

notifier/postgres/store.go
```diff
@@ -2,93 +2,89 @@ package postgres
...
-	return deleted(ctx, s.pool)
-}
+	_ int32 = iota
+	gcLock
+)
 
-// SetDelivered marks the provided notification id as delivered
-func (s *Store) SetDelivered(ctx context.Context, id uuid.UUID) error {
-	return setDelivered(ctx, s.pool, id)
+func errStr(e error) string {
```
crozzy: This function name is a little confusing to me, maybe something like `IsErrLabel`?

notifier/postgres/store.go
```diff
@@ -2,93 +2,89 @@ package postgres
...
 }
 
-// SetDeleted marks the provided notification id as deleted
-func (s *Store) SetDeleted(ctx context.Context, id uuid.UUID) error {
-	return setDeleted(ctx, s.pool, id)
+func txExec(ctx context.Context, m statusMetrics, tx pgx.Tx, name, query string, args []interface{}) error {
+	var err error
+	var tag pgconn.CommandTag
+	timer := prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
+		m.dur.WithLabelValues(name, errStr(err)).Observe(v)
```
crozzy: This could panic based on the status metrics called with this function. We could use the `GetMetricWithLabelValues` method and check errors (but then the nice one-liner is ruined)? Or I'm fine commenting that it could happen

notifier/processor.go
```diff
@@ -39,49 +33,38 @@ type Processor struct {
 	matcher matcher.Service
 	// a store instance to persist notifications
 	store Store
-	// a integer id used for logging
-	id int
+
+	// NoSummary controls whether per-manifest vulnerability summarization
+	// should happen.
+	NoSummary bool
```
crozzy, line 39: Maybe stick this comment :point_down: above the definition?

notifier/service/notifier.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, ErrNoDelivery
+	}
+	srv.del = notifier.NewDelivery(store, locks, del, opts.DeliveryInterval)
+
+	return &srv, nil
+}
+
+// TestModeInit will inject a mock Indexer and Matcher into opts
+// to be used in testing mode.
+func testModeInit(ctx context.Context, opts *Opts) error {
```
crozzy, line 156: This might be nicer if it returned `(*Opts, error)` what do you think?

crozzy (commented):
It's a little difficult to get a picture of the entire logic working together so I intend to poke around a little more running locally but I added some general comments and nits on the code.

notifier/postgres/notifications.go
hdonnay: good catch

notifier/postgres/notifications.go
hdonnay: Yeah, avoiding the copy by letting the `Scan` work on the struct directly. The append is effectively an allocation and then everything is pretty much just modified in place.

notifier/service/notifier.go
hdonnay: :man_shrugging: I think this way of triggering the test behavior is a giant wart that needs to get fixed anyway.

crozzy (dismissed):
LGTM

crozzy (approved):

PR #1489, opened 2022-02-10
contrib: Use the readyz endpoint in startup probes

### LINKED ISSUE #1488: startupProbe and readinessProbe should use the readyz endpoint

Since the `/readyz` endpoint was [introduced](https://github.com/quay/clair/pull/1211) I would expect any checks on startup and readiness to use them but the openshift manifest does not.

### Expected Outcome

The `readinessProbe` and `startupProbe` to use the `/readyz` health endpoint.

### Actual Outcome

They use the `healthz` endpoint.

### Environment

- Clair version/image: main
- Clair client name/version: main
- Host OS: n/a
- Kernel (e.g. `uname -a`): n/a
- Kubernetes version (use `kubectl version`): all
- Network/Firewall setup: n/a

iainduncani:
Fixes #1488

### DIFF
--- a/contrib/openshift/manifests/manifests.yaml
+++ b/contrib/openshift/manifests/manifests.yaml
@@ -70,12 +70,14 @@ objects:
                   port: ${{HEALTH_PORT}}
               readinessProbe:
                 httpGet:
-                  path: ${{HEALTH_PATH}}
+                  path: ${{READY_PATH}}
                   port: ${{HEALTH_PORT}}
               startupProbe:
                 httpGet:
-                  path: ${{HEALTH_PATH}}
+                  path: ${{READY_PATH}}
                   port: ${{HEALTH_PORT}}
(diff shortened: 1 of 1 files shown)


crozzy:
This is a great catch, one other thing to add here is that the startup probe should cover longer running DB migrations (possibly up to 30 mins) so we should set the `failureThreshold` and `periodSeconds` to cover that, something like:

```
  failureThreshold: 200
  periodSeconds: 10
```
?

iainduncani:
@crozzy good point, I would probably make it an hour though as we've seen that length of time with some of the migrations so have updated the commit accordingly.

crozzy (dismissed):

crozzy:
@iainduncani can you please rebase this and we get it merged?

iainduncani:
@crozzy done!

crozzy (approved):

PR #1501, opened 2022-02-25
local-dev: Remove updaters to ensure all run

crozzy:
Currently we only run 4 (types of) updaters, this change
removes an explit list meaning all will run.


crozzy:
@hdonnay was there a reason to exclude some here?

hdonnay:
it's slow to run more of them

PR #1505, opened 2022-03-01
metrics: matcher add status code label to matcher latency metric

crozzy:
This is something that is needed as a hotfix in order to
honor the production SLO.


crozzy:
I'm also working on a PR to move the Matcher service over to the same structure as `IndexerV1` which will be more involved as it will mean updating all metrics we currently specify in alerts / SLOs.

hdonnay (approved):

PR #1533, opened 2022-03-21
go version

hdonnay:


hdonnay:
Changed this to only update the main module. The config module is largely just data structures, so we can update that module when it needs something.

crozzy (dismissed):

go.mod
```diff
@@ -37,4 +36,54 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
 
+require (
```
crozzy, line 39: Had to look this one up, didn't realize this was valid in 1.17

crozzy (approved):

PR #1538, opened 2022-03-23
all: remove Quay keyserver support

hdonnay:
This support was never actually used and only tested locally.


crozzy (dismissed):
LGTM

crozzy (dismissed):

PR #1548, opened 2022-04-11
logger: add request id to log context for traceability

### LINKED ISSUE #1547: Unable to trace requests though logs

Trying to debug problems using the logs is hard as there is no way to connect logs to specific requests. It would be great to have the request id (and possibly more info such as the request URI) included in the logger context so that requests are traceable.

### Environment

- Clair version/image: v4.4.0

paulaldridge:
this allows requests to be traced through the logs

Fixes #1547


crozzy:
let's wait for https://github.com/quay/clair/pull/1514 to be merged so we can do the same for the matcher

hdonnay:
I think this is done, as noted in the linked issue.

PR #1567, opened 2022-05-18
services: update initialization

crozzy:
Claircore refactoring necessitates clair to be more
verbose in its service instanciation, all the dependecies
are now created explicitly in the initialization step and
passed to the constructors via the respective Options structs.


crozzy:
Blocked on https://github.com/quay/claircore/pull/614 and https://github.com/quay/claircore/tree/restructure-dev being merged

hdonnay:
Updated to use published versions. This version references a pseudo-version, but we can change that when the next upstream claircore release is published.

hdonnay (approved):

PR #1592, opened 2022-06-29
contrib: Changes to grafana dashboard

crozzy:
We now have prometheus metrics derived from pgxpool.Stat()
values for the indexer and matcher DBs, this PR adds two
charts to the dashboard describing max, idle, total and acquired.

Right now the metrics are being thrown off by the fact that 500s will
return very fast. This will ensure only latency for successful requests
is considered.


crozzy:
image

crozzy:
image

hdonnay (approved):

PR #1594, opened 2022-06-29
contrib: Charts should take their datasource from the dashboard variable

crozzy:
In most cases outside of local-dev the datasource won't be
called "Prometheus" (as null usually defaults to).


crozzy:
Converted to draft to add some more grafana updates noticed in prod

hdonnay (approved):

PR #1599, opened 2022-07-07
httptransport: notification content type, etc

### LINKED ISSUE #1523: [notifier] api/v1/notification returns contentType: text/plain when application/json is expected

The api call `/notifier/api/v1/notification/{id}` returns with Content-Type: text/plain where application/json is expected. Even if we specify the application/json in the accept header, we still get a plain text response. Would be great if this could be fixed in the upcoming 4.4 release.

### Expected Outcome

Content-Type: application/json

### Actual Outcome

Content-Type: text/plain

### Environment

N/A

- Clair version/image: 4.3.5
- Clair client name/version: 
- Host OS: Docker Container on AWS ECS

hdonnay:
Updated/fixed the notification handler, then updated the discovery endpoint for good measure.


crozzy (dismissed):

hdonnay:
clean rebase and tests

PR #1600, opened 2022-07-08
contrib: First wipe anything that might be left before starting clair indexers

crozzy:
We have observed that the postStart command is not consistently
wiping the /tmp dir as it appears like there is a race
condition as the PVC is not guaranteed to be mounted when it is
run.

### DIFF
--- a/contrib/openshift/manifests/manifests.yaml
+++ b/contrib/openshift/manifests/manifests.yaml
@@ -53,11 +53,10 @@ objects:
                 requests:
                   cpu: ${{INDEXER_CPU_REQS}}
                   memory: ${{INDEXER_MEM_REQS}}
-              lifecycle:
-                postStart:
-                  exec:
-                    command: ['sh', '-c', 'rm -rf /tmp/sha* /tmp/fetch.*']
-              command: ['clair']
+              command:
+                - sh
+                - '-c'
+                - rm -rf /tmp/sha* /tmp/fetch.* && exec clair
               env:
                 - name: CLAIR_CONF
                   value: '/etc/clair/config.yaml'


contrib/openshift/manifests/manifests.yaml
```diff
@@ -53,11 +53,10 @@ objects:
...
                   memory: ${{INDEXER_MEM_REQS}}
-              lifecycle:
-                postStart:
-                  exec:
-                    command: ['sh', '-c', 'rm -rf /tmp/sha* /tmp/fetch.*']
-              command: ['clair']
+              command:
+                - sh
+                - '-c'
+                - rm -rf /tmp/sha* /tmp/fetch.* && clair
```
hdonnay:
```suggestion
                - rm -rf /tmp/sha* /tmp/fetch.* && exec clair
```
Adding `exec` means `clair` will take over the shell process and not have it hanging around.

hdonnay (changes requested):

contrib/openshift/manifests/manifests.yaml
```diff
@@ -53,11 +53,10 @@ objects:
...
-                  exec:
-                    command: ['sh', '-c', 'rm -rf /tmp/sha* /tmp/fetch.*']
-              command: ['clair']
+              command:
+                - sh
+                - '-c'
+                - rm -rf /tmp/sha* /tmp/fetch.* && exec clair
               env:
                 - name: CLAIR_CONF
                   value: '/etc/clair/config.yaml'
```
hdonnay, line 62: @prr approve

hdonnay (dismissed):
LGTM

crozzy:
merging with superpowers as conflictless rebase

PR #1606, opened 2022-07-22
clair: better argument error messages

### LINKED ISSUE #1605: "must provide a -conf flag or set "CLAIR_CONF" in the environment" instead of "File not found"
In this example my config is actually `config.yml` so when `config.yaml` is called in the podman command the file will be "not found" but Clair says I am missing an env var or flag instead.

```
sudo podman run --rm -it --pod clair --name clair_scanner -e "CLAIR_CONF=/clair/config.yaml" -e "CLAIR_MODE=combo" -v ./clair/:/clair:Z 'quay.io/projectquay/clair:4.4.4'

2022/07/21 17:43:30 must provide a -conf flag or set "CLAIR_CONF" in the environment
```

hdonnay:
This replaces the custom flag values with just visiting the flags and
processing the string argument at that time.

Closes: #1605


cmd/clair/main.go
```diff
@@ -35,32 +35,51 @@ const (
...
+		if fv == "" && ok {
+			fv = v
+		}
+		switch f.Name {
+		case "conf":
+			cf, err := os.Open(fv)
+			if err != nil {
+				golog.Fatalf("failed to open config file: %v", err)
+			}
+			defer cf.Close()
```
crozzy, line 64: Maybe w/o the defer and a few lines down, but not sure it matters

crozzy (approved):

cmd/clair/main.go
hdonnay: I don't think it matters

PR #1612, opened 2022-08-15
indexer: Return 4XX status code when Index() returns tarfs.ErrBadFormat

crozzy:
Currently when there is any error indexing a manifest we return InternalServerError.
As Clair can be sent layer files that are not valid, we should return
a client error. This will also help unmask real unexpected application
failures.


hdonnay (approved):
LGTM

PR #1620, opened 2022-09-14
Fix build with newer ubi8/ubi-minimal image; Strip binaries to reduce size

SuperSandro2000:


crozzy:
For context, this is what now happens when using the "old" repo names...
```
[root@140462909e0b /]# microdnf install --disablerepo=* --enablerepo=ubi-8-baseos --enablerepo=ubi-8-appstream
error: repo ubi-8-baseos not found
```

crozzy:
Thanks for the PR, really useful. Can you please `--amend` the commit message to fit the [format](https://github.com/quay/clair/blob/main/Documentation/contribution/commit_style.md) (it helps with auto-compiling the change log (https://github.com/quay/clair/actions/runs/3053778154/jobs/4990285296).

crozzy:
Maybe we should use a stable tag of UBI, advantage being that things like this don't break things (until we make a manual step), disadvantage is that the security updates don't get auto-rolled into newly created images.

SuperSandro2000:
Sigh, done.

> disadvantage is that the security updates don't get auto-rolled into newly created images.

I encountered that problem because we rebuild the image to "fix" CVEs found by clair for itself.

crozzy (approved):

crozzy:
> Sigh, done.
> 
> > disadvantage is that the security updates don't get auto-rolled into newly created images.
> 
> I encountered that problem because we rebuild the image to "fix" CVEs found by clair for itself.

image

crozzy:
thanks @SuperSandro2000

SuperSandro2000:
You're welcome

PR #1648, opened 2022-11-14
build: version of clairctl and clair with git --tags

leiqi96:
When building clair repo with make container-build，the version is hard code in Dockerfile. This patch set the version with git --tags.

Signed-off-by: Anquan Wu <leiqi96@hotmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,13 @@ vendor: vendor/modules.txt
 vendor/modules.txt: go.mod
 	go mod vendor
 
+LAST_GIT_TAG ?= $(shell git describe --tags --match 'v*' 2>/dev/null)
+CLAIR_VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(LAST_GIT_TAG))
 .PHONY: container-build
 container-build:
-	$(docker) build -t clair-local:latest .
+	$(docker) build \
+		--build-arg CLAIR_VERSION=${CLAIR_VERSION} \
+		-t clair-local:latest .
 
 DOCS_DIR ?= ../clair-doc
 .PHONY: docs-build


leiqi96:
> Can one of the admins verify this patch?

no. What should I do

hdonnay:
Is the goal here to have a make target that tries to automatically set the version build argument?

leiqi96:
> Is the goal here to have a make target that tries to automatically set the version build argument?
yes，In addition to automatically setting the version, RELEASE_TAG allows you to customize the version.

This idea is connected with the [issue 1649](https://github.com/quay/clair/issues/1649).  
When I went looking for the cause of the problem, I found the version dev is hard code in  Dockerfile.

leiqi96:
> Is the goal here to have a make target that tries to automatically set the version build argument?

yes，In addition to automatically setting the version, RELEASE_TAG allows you to customize the version.
This idea is connected with the https://github.com/quay/clair/issues/1649.
When I went looking for the cause of the problem, I found the version dev is hard code in Dockerfile.

leiqi96:
@crozzy
can you review this request?

crozzy:
I think it makes sense to add the version in the container when the git context doesn't exist (although it should probably be  the same as cmd/build.go tries to do for consistency i.e a simple `git describe`. I'm less onboard with `RELEASE_TAG` as the version it primarily used as a debugging starting point and letting people set it arbitrarily (or at a stretch accidentally) seems like  it might come back to bite us. Can you explain a situation when someone might use it?

leiqi96:
If the output of command  "git describe --tags --match 'v*' 2>/dev/null"  is null,  A custom version may be useful.

CUSTOME_TAG  would be a more appropriate name.

hdonnay:
I think #1660 does this in a more comprehensive (if also more involved) way. I'd appreciate it if you could try that branch and see if it gives desirable results.

hdonnay:
Closing in favor of #1660.

PR #1659, opened 2022-12-05
database: add mapping for newer Ubuntu versions

s-kotte:
Add mapping for Ubuntu Groovy, Hirsute, Impish, Jammy and Kinetic.
This is needed to support newer Ubuntu versions.


crozzy:
Hi @s-kotte, thanks for the PR, unfortunately we no longer maintain Clair V2 so there will be no further changes accepted.  All of these distros are available in V4 and are now auto-discoverable .

Thanks

PR #1660, opened 2022-12-05
cmd: version injection

hdonnay:
These changes use git to inject the `git describe` information into the common version information. This is what our release process and the go module proxy use for their archives. In addition, the `Version` setting function gained some logic to detect if the underlying `claircore` has been swapped out and report on its version as well, if so.

The second commit then cleans up various build spots and the Dockerfile.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+cmd/build.go export-subst
+*.go diff=golang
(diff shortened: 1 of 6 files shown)


hdonnay:
The `claircore` special casing may seem odd, but it will help signpost nightly builds, as those have `replace` stanzas injected.

crozzy (approved):
this LGTM, I took a look at the midstream and I don't see anything obvious that needs to be updated, might be missing something

hdonnay:
The Quay midstream builds use the `build-arg`, so we'll need to modify the Dockerfile to not have the forced suffix, if it still uses a version of our Dockerfile.

PR #1661, opened 2022-12-06
httptransport: add a `request_id` to logs

### LINKED ISSUE #1547: Unable to trace requests though logs

hdonnay:
This re-uses any trace ID that OpenTelemetry supports, or generates a weakly random one based on some request features.

Closes: #1547

### DIFF
--- a/httptransport/common.go
+++ b/httptransport/common.go
@@ -1,14 +1,21 @@
 package httptransport
 
 import (
+	crand "crypto/rand"
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"math/rand"
 	"mime"
 	"net/http"
 	"path"
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
+
+	"github.com/quay/zlog"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/quay/claircore"
 )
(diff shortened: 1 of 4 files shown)


hdonnay:
cc @paulaldridge

httptransport/common.go
```diff
@@ -106,3 +114,24 @@ func (a *accept) Match(mt string) bool {
...
+	const key = `request_id`
+	ctx := r.Context()
+	sctx := trace.SpanContextFromContext(ctx)
+	if sctx.HasTraceID() {
+		ctx = zlog.ContextWithValues(ctx, key, sctx.TraceID().String())
+	} else {
+		h := idPool.Get().(hash.Hash64)
+		h.Reset()
+		io.WriteString(h, r.RequestURI)
+		io.WriteString(h, r.RemoteAddr)
```
crozzy: Is this going to render the same hash for the same `RequestURI` `RemoteAddr` pair? Are you thinking the port will add enough entropy?

crozzy (dismissed):
LGTM just one comment

httptransport/common.go
hdonnay: That's my thinking. Hopefully a client isn't requesting the same URI on the same connection. Although, as I type that, perhaps  that's not good enough and we should just use a `rand` source or similar.

crozzy: I think the majority of the time it'll work :ok_hand: I just wonder if there are certain proxy situations where it's always 127.0.0.1:same_port

hdonnay: Swapped it for a pool of `rand.Rand`s. That should just work in all cases.

crozzy (approved):

PR #1667, opened 2023-01-06
Strict client

hdonnay:
This PR reworks the HTTP client framework inside of Clair.

It removes the `http.RoundTripper` that modified passing requests (a source of some previous data races and recommended against in the documentation) in favor of some helper functions.

- The signing is pulled out into a dedicated object that modifies requests, meaning the caller needs to have one and use it.
- The user-agent setting is pulled into a wrapper around `http.NewRequestWithContext`
  This also means that if we want to enforce additional defaults on Requests, it can be done here without needing to hunt down call sites.
- The client constructor now only cares about whether it should use a dialer that allows public networks or not.

Then all the packages are updated to use this new API surface as needed.

### DIFF
--- a/Documentation/reference/config.md
+++ b/Documentation/reference/config.md
@@ -159,9 +159,9 @@ A key file for the TLS certificate. Encryption is not supported on the key.
 Indexer provides Clair Indexer node configuration.
 
 #### `$.indexer.airgap`
-Boolean.
-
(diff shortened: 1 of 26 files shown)


crozzy (dismissed):
LGTM

hdonnay:
Clean rebase, waiting for CI

PR #1680, opened 2023-02-07
chore: Update go version in Dockerfile

crozzy:
Making the default go version 1.20 in the Dockerfile.


crozzy:
Ha, that's cool that this fails

hdonnay (dismissed):
LGTM, but I'm unsure why that one check is failing.

hdonnay (approved):

PR #1688, opened 2023-02-10
contrib: Set the version when building the service image

crozzy:
Pipe the output of the git archive command to docker build to include
the version context. It seemed better to stop using the makefile
rather than add this specific use-case.

### DIFF
--- a/contrib/openshift/build_and_deploy.sh
+++ b/contrib/openshift/build_and_deploy.sh
@@ -4,13 +4,13 @@ set -exv
 REPOSITORY="quay.io/app-sre"
 IMAGE="${REPOSITORY}/clair"
 
-make container-build
+git archive HEAD|
+docker build -t clair-service:latest -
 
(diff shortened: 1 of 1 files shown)


contrib/openshift/build_and_deploy.sh
```diff
@@ -4,13 +4,14 @@ set -exv
 REPOSITORY="quay.io/app-sre"
 IMAGE="${REPOSITORY}/clair"
 
-make container-build
-
 GIT_HASH=`git rev-parse --short=7 HEAD`
+
+docker build --build-arg CLAIR_VERSION=${GIT_HASH} -t clair-service:latest .
```
hdonnay:
```suggestion
git archive HEAD|
docker build -t clair-service:latest .
```

How does something like this sit with you?

hdonnay (dismissed):
Wow great idea👍💯

crozzy:
Going to merge as it was approved before and in fact the reviewers suggestion

PR #1700, opened 2023-03-16
airgap: Remove libindex Airgap option

crozzy:
This option no longer exists in claircore and is handled by clair when creating the client that is passed to claircore.


crozzy:
related: https://github.com/quay/claircore/pull/863

hdonnay (approved):
I don't think this requires any documentation changes.

PR #1704, opened 2023-03-27
chore: Bump Claircore to v1.4.20

crozzy:
Bump Claircore to the latest tag.


RTann (dismissed):
LGTM though not sure what's going on with CI

hdonnay:
I'll try to reproduce tomorrow

hdonnay (approved):
Hm, looks like `pgx` was working around a bug and no longer is. Definitely don't want to be passing a `**Notifier`.

PR #1711, opened 2023-04-12
chore: Update go version in go.mod

crozzy:
Some of the go detection logic was prone to this bug: https://github.com/golang/go/issues/33121 in the go debug library. As 1.20 includes the fix we deem it required to avoid panics originating in claircore.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/quay/clair/v4
 
-go 1.18
+go 1.20
 
 require (
 	github.com/go-stomp/stomp v2.0.8+incompatible


hdonnay (approved):

hdonnay:
Possible `NeedsBackport`? Although I don't think changing the required version in a patch release is a great idea.

crozzy:
> Possible `NeedsBackport`? Although I don't think changing the required version in a patch release is a great idea.

I *think* we don't backport this change and it gets auto-rolled into 4.7.0, if any bugs are reported for the issue we can instruct them to update their go version.

PR #1715, opened 2023-04-17
Reduce confusion by using one path for the notification endpoint

SuperSandro2000:
Based on https://github.com/quay/clair/blob/main/httptransport/notification_v1.go#L50 I think singular should be the "correct" one

If this is wrong, all mentions of ``notifier/api/v1/notification`` should get the s appended.


crozzy:
Thanks for the contribution, this seems like a positive change, there are just a couple of issues that the CI is highlighting ([commit style](https://github.com/quay/clair/blob/main/Documentation/contribution/commit_style.md), etc) but apart from that it lgtm

hdonnay (approved):

PR #1732, opened 2023-05-04
chore: move emulator tests to a nightly run

hdonnay:
These tests are annoyingly slow.
If anyone is concerned about tests not running on every commit, please get in touch -- we'd like to have hardware runners.


crozzy (approved):
:timer_clock: :100: :champagne:

PR #1743, opened 2023-05-09
stomp: use rabbitmq

hdonnay:
This updates the `stomp` package and the tests to use only RabbitMQ instead of both RabbitMQ and ActiveMQ.


hdonnay:
@RTann, I'm very sorry for forgetting to put this in draft :(

local-dev/traefik/config/rabbitmq.yaml
```diff
@@ -0,0 +1,26 @@
+---
+http:
+  routers:
+    activemq:
```
crozzy: should this be `rabbitmq`?

crozzy (commented):
Looks good, just one comment

local-dev/traefik/config/rabbitmq.yaml
hdonnay: yes

crozzy (approved):

PR #1747, opened 2023-05-15
updater: Extend default updater time to 6 hours

crozzy:
Given the expense of updating and the lack of value to such a high frequency this change extends the default update period from 30 mins to 6 hours.


RTann (approved):
LGTM

RTann:
ACS used to do it every 1 hour, but now it's 3. 6 makes sense. Usually I only see updates 1x/day

hdonnay (approved):
+1

PR #1759, opened 2023-05-17
config: support drop-ins

hdonnay:
While debugging something, I got annoyed that SOP is to put the clair config into a k8s Secret because password auth is alarmingly common for database connections. This is a POC for systemd-style drop-in configs, with hopefully sane merge semantics.

If people think this is a good idea, there's two PRs here: one to update the `config` module, then another updating the config module and adding the new code.


hdonnay:
done via #1781 and #1782

PR #1782, opened 2023-06-15
dropins: [2/2]: dropins

hdonnay:
Part 2 of dropins. Needs new config module tagged, so blocked on #1781.


hdonnay:
Probably warrants a callout in the prose documentation.

cmd/config.go
```diff
@@ -0,0 +1,187 @@
...
+//
+// For example, given the paths:
+//
+//	config.yaml
+//	config.yaml.d/
+//	config.yaml.d/secrets.yaml
+//	config.yaml.d/override.yaml-patch
+//	config.yaml.d/unloved.json-patch
+//
+// "Config.yaml" will be the base config,
```
crozzy, line 32: nit: case

cmd/clair/main.go
```diff
@@ -55,13 +54,8 @@ func main() {
...
 		switch f.Name {
 		case "conf":
-			cf, err := os.Open(fv)
-			if err != nil {
-				golog.Fatalf("failed to open config file: %v", err)
-			}
-			defer cf.Close()
-			if err := yaml.NewDecoder(cf).Decode(&conf); err != nil {
-				golog.Fatalf("failed to decode yaml config: %v", err)
+			if err := cmd.LoadConfig(&conf, fv, true); err != nil {
```
crozzy, line 57: Thinking aloud; do we want strict to be false here? Like, is it more useful to know all the errors at once or try repeatedly?

crozzy (commented):
Looks great and not too complex, just a clarifying comment

cmd/config.go
hdonnay: It's the start of a sentence.

cmd/clair/main.go
hdonnay: My thinking was if it fails, you can use the new `clairctl` subcommand to check it. And also to dry-run the config before committing it.

crozzy: Yeah the new command will be really useful. I guess it's more complicated as some errors will likely lead to other errors so it's not checking the actual config for correctness (i.e. strange values for certain fields where the multi errors would be really useful) but checking the config structure for correctness i.e.:
* get an error from `b, err := loadAsJSON(name)`
* will cause an error from `if len(b) == 0`
* will cause an error from `dec.Decode(cfg)`

As I write that I'm wondering if the non-strict mode makes sense

cmd/config.go
crozzy: My thought is `Config.yaml` != `config.yaml`, no matter, it's a nit, I'm confident no one will actually get confused by this.

cmd/clair/main.go
hdonnay: I think permissive for a "check" tool and "strict" for actually using the config works the best once it's a valid Config document, but yeah it'll throw odd errors before that.

crozzy (approved):

PR #1784, opened 2023-06-15
cicd: fix nightly-ci error

hdonnay:
This fixes the nightly-ci workflow and adds a means to lint and validate the workflow files.

### DIFF
--- a/.github/workflows/Makefile
+++ b/.github/workflows/Makefile
@@ -1,0 +1,19 @@
+check: github-workflow.json-schema yajsv yq
+	for f in *.yml; do ./yq -o json "$$f" > "$${f%yml}json"; done
+	./yajsv -s $< *.json
+	rm *.json
+	command -v yamllint >/dev/null 2>&1 && yamllint .
+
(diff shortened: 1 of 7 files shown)


.github/workflows/Makefile
```diff
@@ -0,0 +1,19 @@
+check: github-workflow.json-schema yajsv yq
+	for f in *.yml; do ./yq -o json "$$f" > "$${f%yml}json"; done
```
crozzy, line 2: super nit, but now that we'll potentially have a lot of `yaml` (but no `yml`) files in the repo with the config changes should we just decide on one supported extension? I suppose we can say configs will be `yaml` and GH manifests will be `yml`

crozzy (approved):
I'm fine to merge this, just offered a comment

.github/workflows/Makefile
hdonnay: Yeah, this is just for the github actions, as it seems like it all prefers `yml`.

PR #1795, opened 2023-06-21
cicd: fix nightly connection strings

hdonnay:
This fixes the nightly CI through some changes to the stomp package and moving the actual test job into a common workflow. The common workflow didn't exist or was new enough that I didn't know about it when the nightly CI was added.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,127 @@
+---
+name: Tests
+
+on:
+  workflow_call:
+    inputs:
+      package_expr:
+        required: true
+        type: string
+        description: 'Package expression(s) passed to `go test`'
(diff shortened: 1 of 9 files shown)


hdonnay:
This does change things such that builds are always run in a container again, but the common workflow does try to avoid some extra work and does make sure to use arch-specific caches.

hdonnay:
image
The nightly CI also gained superpowers to be able to specify the package expression, go version, and platform when triggering a run manually. This should help when hunting issues by making the execution time much shorter.

.github/workflows/tests.yml
crozzy, line 1: cool!

crozzy (approved):

PR #1800, opened 2023-06-26
chore: Add buildvsc=false flag when building

crozzy:
There appears to have been some change in the default behaviour between 1.18 and 1.20 where 1.20 will by default try to stamp the created binary with VSC information, for us the .git/ dir is no present so it's failing. See https://github.com/golang/go/issues/51748.

### DIFF
--- a/.github/workflows/cut-release.yml
+++ b/.github/workflows/cut-release.yml
@@ -139,7 +139,7 @@ jobs:
         run: |
           tar -xz -f ${{steps.download.outputs.download-path}}/clair-${{ needs.config.outputs.version }}.tar.gz --strip-components=1
           go build\
-            -trimpath -ldflags="-s -w"\
+            -trimpath -ldflags="-s -w" -buildvcs=false\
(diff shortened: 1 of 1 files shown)


hdonnay:
I don't think this is spelled correctly

.github/workflows/cut-release.yml
```diff
@@ -139,7 +139,7 @@ jobs:
         run: |
           tar -xz -f ${{steps.download.outputs.download-path}}/clair-${{ needs.config.outputs.version }}.tar.gz --strip-components=1
           go build\
-            -trimpath -ldflags="-s -w"\
+            -trimpath -ldflags="-s -w" -buildvsc=false\
```
hdonnay:
```suggestion
            -trimpath -ldflags="-s -w" -buildvcs=false\
```

hdonnay (changes requested):

.github/workflows/cut-release.yml
crozzy: ffs

PR #1801, opened 2023-06-26
cicd: don't checkout source on clairctl builds

hdonnay:
The presence of a `.git` directory messes with the new go toolchain vcs tooling, and we don't actually want a source checkout, anyway.


crozzy (approved):
spelling `vcs` the traditional way I see

hdonnay:
American spelling

PR #1803, opened 2023-06-29
contrib: Add manifest for a Job to run DB jobs

crozzy:
We need to be able to deploy Jobs into the production cluster that are tasked with running maintenance jobs against the DBs.


hdonnay (approved):
LGTM

PR #1804, opened 2023-06-29
chore: Remove dogstatsd variable and references

crozzy:
This came up as we have this field in the documentation but clairctl will error out if it's used when invoking any command.


hdonnay (approved):
LGTM

PR #1805, opened 2023-06-29
Clairctl: Logging changes

crozzy:
* Remove DSN logging
* Add line signifying done


hdonnay (approved):
Ah, yeah, probably not great to log the string that may have a password in it.

PR #1807, opened 2023-07-03
clairctl: Scan the pointer to the pointer of the bool

crozzy:
I believe that even though ok is a *bool type, we still need to pass it's address to Scan() otherwise it's akin to ok := bool; Scan(&ok), which I believe only supports binary values and not nullable bools.


crozzy:
not really sure why the function returns on the `Scan()` though, I would have thought it'd just push that info into the error

hdonnay (approved):
Ah, oops

hdonnay:
The devshift.net build looks like a spurious failure. I don't know how to re-trigger it, though.

PR #1816, opened 2023-07-07
docs: Add description of debugging services available during local-dev

crozzy:
As pyroscope was added I realized that these services are not documented but could be really useful to contributors.


crozzy:
closes https://github.com/quay/clair/issues/1815

hdonnay (approved):
lgtm

PR #1826, opened 2023-07-24
chore: Bump claircore to v1.5.12

crozzy:
Update claircore dependency to latest tag.


RTann (approved):

RTann:
might be worth looking into making a github workflow to trigger the claircore bump upon creating the claircore tag so you can just come here and approve some bot's PR instead of requiring another person's approval

PR #1833, opened 2023-07-27
cmd: version for old gits

hdonnay:
This also adds a workaround for actions/checkout#882 where the new tag get blasted away with a new, lightweight tag for the duration of the run.


crozzy (approved):
LGTM

PR #1858, opened 2023-09-15
contrib: Add annotations to clair deployment manifest template

syed:
This can be used to restart clair in prod without doing a dummy image build

### DIFF
--- a/contrib/openshift/manifests/manifests.yaml
+++ b/contrib/openshift/manifests/manifests.yaml
@@ -38,6 +38,8 @@ objects:
           labels:
             service: indexer
             app: clair
+          annotations:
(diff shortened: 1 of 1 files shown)


hdonnay:
Wouldn't this require new commits to this repository? Why can't something like `kubectl rollout restart` be used?

crozzy:
Nice, this is definitely something we'd want. Is there not a way to do this directly in app-interface? 

The only thing I'd say about the PR itseld is; is there an advantage to being about to override the annotation key? That seems like something that we'd never update. Thanks!

hdonnay (commented):
@crozzy what do you want to do here?

PR #1859, opened 2023-09-18
http updates

hdonnay:
This patchset ends up making a few interrelated changes:

- Update the internal middleware to understand the `http.ResponseController` contract.
- Make use of the `ResponseController` pattern.
- Change the `apiError` helper to be impossible to use incorrectly from an `http.Handler`.
- Update the DiscoveryHandler to use the new style.
- Simplify the `httptransport.New` usage by not returning a custom type.

Supersedes and closes #1719.


middleware/compress/handler.go
```diff
@@ -152,80 +153,77 @@ func (c *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	return c.zwriter.Write(b)
+}
+func (c *compressWriter) FlushError() error {
+	e0 := c.zwriter.Flush()
+	e1 := http.NewResponseController(c.ResponseWriter).Flush()
+	if errors.Is(e1, http.ErrNotSupported) {
+		e1 = nil
 	}
-	c.next.ServeHTTP(nw, r)
+	return errors.Join(e0, e1)
```
crozzy: Thoughts on calling these something more descriptive?

httptransport/error.go
```diff
@@ -52,7 +56,14 @@ func apiError(ctx context.Context, w http.ResponseWriter, code int, f string, v
...
+	switch err := http.NewResponseController(w).Flush(); {
+	case errors.Is(err, nil):
+	case errors.Is(err, http.ErrNotSupported):
+		// Skip
+	default:
+		zlog.Warn(ctx).
+			Err(err).
+			Msg("unable to flush http response")
 	}
+	runtime.Goexit()
```
crozzy, line 89: I'm a little on the fence about this, on the one hand I like the fact that it's terminal and it is always the last function called before the return anyway so why not incorporate the "return", however, the `_v1.go` files seem slightly harder to follow

```go
		report, ok, err := h.srv.IndexReport(ctx, d)
		if !ok {
			apiError(ctx, w, http.StatusNotFound, "index report not found")
		}
		if err != nil {
			apiError(ctx, w, http.StatusInternalServerError, "could not retrieve index report: %v", err)
		}
                // It appears like we'd get to here even if !ok and err != nil, but that isn't the case
		w.Header().Add("etag", validator)
```

httptransport/server.go
```diff
@@ -40,204 +40,96 @@ const (
...
 
-	var e error
+	// Note that the arms have a conditional at the end to make them work like a
+	// C switch.
 	switch conf.Mode {
-	case config.ComboMode:
-		e = t.configureComboMode(ctx)
-		if e != nil {
-			return nil, e
+	case config.IndexerMode, config.ComboMode:
```
crozzy: So much cleaner but maybe we don't want a switch here and it'd be clearer with common or garden `if`s

```go
	if conf.Mode == config.IndexerMode || conf.Mode == config.ComboMode {
		if indexer == nil {
			return nil, fmt.Errorf("mode %q requires an indexer service", conf.Mode)
		}
		prefix := indexerRoot + apiRoot

		v1, err := NewIndexerV1(ctx, prefix, indexer, traceOpt)
		if err != nil {
			return nil, fmt.Errorf("indexer configuration: %w", err)
		}
		var sem *semaphore.Weighted
		if ct := conf.Indexer.IndexReportRequestConcurrency; ct > 0 {
			sem = semaphore.NewWeighted(int64(ct))
		}
		rl := &limitHandler{
			Check: func(r *http.Request) (*semaphore.Weighted, string) {
				if r.Method != http.MethodPost && r.URL.Path != IndexAPIPath {
					return nil, ""
				}
				// Nil if the relevant config option isn't set.
				return sem, IndexAPIPath
			},
			Next: v1,
		}
		mux.Handle(prefix, rl)
	}
	if conf.Mode == config.MatcherMode || conf.Mode == config.ComboMode {
		if indexer == nil || matcher == nil {
			return nil, fmt.Errorf("mode %q requires both an indexer service and a matcher service", conf.Mode)
		}
		prefix := matcherRoot + apiRoot
		v1 := NewMatcherV1(ctx, prefix, matcher, indexer, time.Duration(conf.Matcher.CacheAge), traceOpt)
		mux.Handle(prefix, v1)
	}
	if conf.Mode == config.NotifierMode || conf.Mode == config.ComboMode {
		if conf.Mode == config.ComboMode && notifier == nil {
			zlog.Debug(ctx).Msg("skipping unconfigured notifier")
			break // Skip
		}
		if notifier == nil {
			return nil, fmt.Errorf("mode %q requires a notifier service", conf.Mode)
		}
		prefix := notifierRoot + apiRoot
		v1, err := NewNotificationV1(ctx, prefix, notifier, traceOpt)
		if err != nil {
			return nil, fmt.Errorf("notifier configuration: %w", err)
		}
		mux.Handle(prefix, v1)
	}
```
Up to you

crozzy (commented):
The meat of this looks great, the couple of comments are pretty much style nits

httptransport/error.go
hdonnay: I suppose, but it's documented pretty well, IMO? I can add an additional callout in the docs if that would help.

crozzy (approved):
LGTM

PR #1870, opened 2023-09-20
config: add OTLP configuration types

hdonnay:
Closes #1857.

PR #1895, opened 2023-10-18
config: add Sentry

hdonnay:
Depends on #1870

### DIFF
--- a/config/introspection.go
+++ b/config/introspection.go
@@ -10,12 +10,14 @@ type Trace struct {
 	Probability *float64  `yaml:"probability,omitempty" json:"probability,omitempty"`
 	Jaeger      Jaeger    `yaml:"jaeger,omitempty" json:"jaeger,omitempty"`
 	OTLP        TraceOTLP `yaml:"otlp,omitempty" json:"otlp,omitempty"`
+	Sentry      Sentry    `yaml:"sentry,omitempty" json:"sentry,omitempty"`
 }
 
 func (t *Trace) lint() ([]Warning, error) {
 	switch t.Name {
 	case "":
 	case "otlp":
+	case "sentry":
 	case "jaeger":
 		return []Warning{{
 			path: ".name",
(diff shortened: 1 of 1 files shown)


hdonnay:
Pretty sure these are the only parts of the [Sentry SDK] we want to expose directly. There's already a sampling knob, and we should be able to cook up a way to get sample-only-if-marked working.

[Sentry SDK]: https://pkg.go.dev/github.com/getsentry/sentry-go#ClientOptions

config/introspection.go
```diff
@@ -50,6 +52,17 @@ type Jaeger struct {
...
 
+// Sentry is the [Sentry] specific tracing configuration.
+//
+// [Sentry]: https://sentry.io
+type Sentry struct {
+	// DSN to be passed to [github.com/getsentry/sentry-go.ClientOptions.Dsn].
+	DSN string `yaml:"dsn" json:"dsn"`
+	// Environment to be passed to
+	// [github.com/getsentry/sentry-go.ClientOptions.Environment].
+	Environment string `yaml:"environment,omitempty" json:"environment,omitempty"`
```
crozzy, line 63: Presumably this will be `stage`, `prod` etc?

crozzy: Regarding `Debug`, do we want to allow users to toggle that or could we eventually key it off of `clair.config.log_level`.

crozzy: I don't think anything else makes sense, `dist` I'd imagine is something we want to describe the Clair version (and all it's facets).

hdonnay: I think the `Debug` spews stuff out to a configured writer, so we'd have to cook something up to repackage that into our logs. I think I can/will add this if it ends up being needed.

hdonnay: That is to say: yes, we should control that via the main logging tunable, but it'll complicate the integration a little bit.

crozzy (approved):
LGTM

PR #1905, opened 2023-10-31
all: annoying scan warnings

hdonnay:
Someone decided to run a bunch of static analysis tools and make these could-be-a-problem-if-you-squint issues our problem.


crozzy (approved):
LGTM

PR #1915, opened 2023-11-16
admin: add command to update go packages with norm_version

crozzy:
There are a number of Go packages in the `package` table that do not have the needed norm_version which is required for Go matching.

This looks up the potential Go packages from the `matcher.vuln` table, the difficultly doing this is entirely in the indexer DB is how to identify that these packages came from the gobin scanner. Trying to use the `package_scanartifact` (joining to the `scanners` table is prohibitively expensive as the query path is not accounted for in the index (i.e. we're always expecting the first clause to the the layer_id).

### DIFF
--- a/cmd/clairctl/admin.go
+++ b/cmd/clairctl/admin.go
@@ -297,9 +297,12 @@ func adminPre473(c *cli.Context) error {
 
 func updateGoPackages(c *cli.Context) error {
 	const (
-		getPackageNames = "SELECT DISTINCT package_name FROM vuln WHERE updater = 'osv/go'"
-		getPackages     = "SELECT id, version FROM package WHERE name = $1 and norm_version IS NULL"
-		updatePackages  = "UPDATE package SET norm_version=$1::int[], norm_kind=$2 WHERE id = $3 and norm_version IS NULL"
+		// TODO (crozzy): This could describe something more interesting like >6 or 6-10 but
+		// at the moment that seems like overkill.
+		compatibleMigrationVersion = 7
+		getPackageNames            = "SELECT DISTINCT package_name FROM vuln WHERE updater = 'osv/go'"
+		getPackages                = "SELECT id, version FROM package WHERE name = $1 and norm_version IS NULL"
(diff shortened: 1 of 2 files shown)


crozzy:
Local output:
```
 ► go build -o . ./cmd/clairctl && ./clairctl -D -c local-dev/clair/config.yaml admin oneoff update-golang-packages
2023-11-16T10:00:32-08:00 INF using discovered connection params database=matcher host=localhost port=32775 user=clair
2023-11-16T10:00:32-08:00 DBG resizing pool pool size=2
2023-11-16T10:00:32-08:00 INF using discovered connection params database=indexer host=localhost port=32775 user=clair
2023-11-16T10:00:32-08:00 DBG resizing pool pool size=2
2023-11-16T10:00:32-08:00 DBG working on version package_name=k8s.io/apiserver version=v0.27.4
2023-11-16T10:00:32-08:00 INF successfully updated package row package_id=3818 package_name=k8s.io/apiserver rows affected=1 version=v0.27.4
2023-11-16T10:00:32-08:00 DBG working on version package_name=stdlib version="go1.20.10 X:strictfipsruntime"
2023-11-16T10:00:32-08:00 WRN error parsing semver error="Invalid Semantic Version" package_name=stdlib version="go1.20.10 X:strictfipsruntime"
```

hdonnay:
Do we want this to be `go` specific, or couch it as a general "fix-up vulnerability table" command?
I don't feel strongly, other than my initial instinct was to do the latter.

crozzy:
> Do we want this to be `go` specific, or couch it as a general "fix-up vulnerability table" command? I don't feel strongly, other than my initial instinct was to do the latter.

I think the issue with it not being go-specific is; is the version parsing logic and the `norm_kind` value applicable to any other ecosystems? Most matchers aren't relying on that value, are you thinking that we could make it general purpose and convert some more matching logic over to do `func (matcher *Matcher) VersionAuthoritative() bool { return true }`?

hdonnay:
I meant for the command itself -- are we ever going to need another `update-someother-package` command?

hdonnay (dismissed):
LGTM

hdonnay (approved):

PR #1945, opened 2024-01-04
config: update minimum TLS version for server

crozzy:
The config is used both when setting up the HTTP API and for the notifier's Deliverer.


hdonnay (approved):
LGTM

PR #1954, opened 2024-01-18
clair: restructure shutdown

### LINKED ISSUE #1946: Graceful shutdown ... isn't

`cmd/clair/main.go` installs a signal handler (SIGINT only) to perform a graceful shutdown.
I don't think this is working as desired

### Expected Outcome
I would have expected HTTP servers to stop accepting new connections, then a [short grace period](https://github.com/quay/clair/blob/6de0d807a5d63d5b10d63c274c297fcf095a308b/cmd/clair/main.go#L163) for in-flight requests to complete (exiting as soon as no requests are in flight), as a last resort terminating connections then exiting if any are still running when the grace period expires.

### Actual Outcome
HTTP servers are shutdown immediately, with any in-flight HTTP requests being severed. 

**Recreate:**

1. start a local Postgres
1. run a clair indexer locally:
   `go run ./cmd/clair/ -conf ./local-dev/clair/config-frostmar-local.yaml -mode indexer`
1. start a long-running HTTP request, eg an indexing operation
   `curl -H "Content-Type: application/json" http://localhost:6060/indexer/api/v1/index_report -d ...<snipped>...`
1. send SIGINT - [CTRL+C] the `go run` command, or `kill -INT <pid>`
   => logs output `gracefully shutting down  component=main signal=interrupt`
   => process _immediately_ exits
   => curl request immediately gives error `curl: (52) Empty reply from server`

I'd hope to see:
=> logs to output `gracefully shutting down  component=main signal=interrupt`
=> 10 sec grace-period timeout
=> process exits only after the grace period
=> logs to output `unregistered signal handler`
=> curl to fail with an error response

### Environment

Running locally on a Mac

- Clair version/image: git tag `v4.7.2`
- Clair client name/version: curl
- Host OS: MacOS
- Kernel (e.g. `uname -a`): 
- Network/Firewall setup: n/a

hdonnay:
This changes the incorrect-but-at-least-stops implementation with a
little tree of goroutines.

The full fix requires `context.WithoutCancel`, which we can't quite use
yet.

Closes: #1946


hdonnay:
Current dependencies on/for this PR:
* `main`
  * **PR #1954** <a href="https://app.graphite.dev/github/pr/quay/clair/1954?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1955** <a href="https://app.graphite.dev/github/pr/quay/clair/1955?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/quay/clair/1954?utm_source=stack-comment).

frostmar:
Could you add a SIGTERM too while refactoring here?
It'd be especially convenient when running under Kubernetes, which sends SIGTERM when it's time for a container to shutdown

hdonnay:
This'll probably have to wait until go1.22 releases, so we can make use of new go1.21 API.

frostmar:
FYI: go 1.22 released 6 February 2024 
https://go.dev/blog/go1.22

PR #1963, opened 2024-01-29
httptransport: GET vuln report returns 404 when indexing in-progress

frostmar:
For issue https://github.com/quay/clair/issues/1962

`GET /matcher/api/v1/vulnerability_report/<digest>` returns HTTP404 when the index report is not complete, instead of a partial vulnerability report based on whatever is present in the index

This doesn't distinguish between in-progress states (a later request will likely be successful) and the final error state `IndexError` (a later request will likely still be unsuccessful). I'm not sure what would be preferred - HTTP404? HTTP500 with some error body?

### DIFF
--- a/httptransport/matcher_v1.go
+++ b/httptransport/matcher_v1.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/quay/claircore"
+	indexerController "github.com/quay/claircore/indexer/controller"
 	"github.com/quay/claircore/libvuln/driver"
 	"github.com/quay/zlog"
 	oteltrace "go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace"
@@ -125,9 +126,10 @@ func (h *MatcherV1) vulnerabilityReport(w http.ResponseWriter, r *http.Request) 
 	if err != nil {
 		apiError(ctx, w, http.StatusInternalServerError, "experienced a server side error: %v", err)
 	}
-	// now check bool only after confirming no err
-	if !ok {
+	// now check present and finished only after confirming no err
+	if !ok || indexReport.State != indexerController.IndexFinished.String() {
 		apiError(ctx, w, http.StatusNotFound, "index report for manifest %q not found", manifest.String())
+		return
 	}
 
 	vulnReport, err := h.srv.Scan(ctx, indexReport)


frostmar:
@crozzy sorry for the bump - just checking if this PR has been overlooked?

httptransport/matcher_v1.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/quay/claircore"
+	indexerController "github.com/quay/claircore/indexer/controller"
```
crozzy, line 15: I wouldn't bother with the alias here.

httptransport/matcher_v1.go
```diff
@@ -125,9 +126,10 @@ func (h *MatcherV1) vulnerabilityReport(w http.ResponseWriter, r *http.Request)
 	if err != nil {
 		apiError(ctx, w, http.StatusInternalServerError, "experienced a server side error: %v", err)
 	}
-	// now check bool only after confirming no err
-	if !ok {
+	// now check present and finished only after confirming no err
+	if !ok || indexReport.State != indexerController.IndexFinished.String() {
```
crozzy, line 130: nit but it might be nicer to compare the state types as opposed to the strings:
```go
var state controller.State
state.FromString(indexReport.State)
if !ok || state != controller.IndexFinished {
...
```

crozzy (commented):
I think this is a good idea. Thinking through re-indexing, will there be a situation where the report is 200OK then is being re-indexed so will be 404Not Found the 200OK again?

crozzy:
> @crozzy sorry for the bump - just checking if this PR has been overlooked?

Sorry @frostmar it had indeed fallen through the cracks

frostmar:
>Thinking through re-indexing, will there be a situation where the report is 200OK then is being re-indexed so will be 404Not Found the 200OK again?

I think that will happen, from the user's point of view the vuln report seems to be unavailable for a brief period.

That's pretty much true, however. While (re-)indexing is in progress, there is no valid index report to base a vuln report on, so better to say it's unavailable than give an incorrect one? 

As a future extension, there could be a new response instead of HTTP404, to communicate "come back shortly", used when we know we _ought_ to have a completed vuln report available shortly (as indexing is in progress, but hasn't completed) .... but this seems like an edge case that few will benefit from.

crozzy:
Thanks for the updates, I think it seems like an improvement, @hdonnay is there a situation with clients where you could see this being a problem?

@frostmar looks like you checked in a .vscode file

frostmar:
>... checked in a .vscode file

Sorry, sorted now

frostmar:
@crozzy / @hdonnay I'm hoping this can make the next release of Clair, if it looks OK to you now?

frostmar:
@crozzy / @hdonnay bump - can I do anything to progress this please?

crozzy (dismissed):

crozzy (approved):
Thanks!

PR #1985, opened 2024-02-19
httptransort: fix http metrics

hdonnay:
The `runtime.Goexit` usage messes with the `prometheus` metrics, so this inserts a handler to catch unwinds using the documented `http.ErrAbortHandler`.


crozzy (approved):
lgtm

PR #1999, opened 2024-03-05
httputil: work around NPE in signer.Add()

glebsts:
If `auth` section is empty in config file, instance of `Signer` is initialized with default field values:
https://github.com/quay/clair/blob/v4.7.3/internal/httputil/signer.go#L29

Later e.g. in `report` command processing if configuration is used, method `Add` is used as _"marks the authority in uri as one that expects signed requests"_:
https://github.com/quay/clair/blob/v4.7.3/cmd/clairctl/report.go#L145

In this method an entry is being added to `use` field of `signer` receiver no matter what. But if `auth` section was missing in config, `use` has default value of nil and produces `assignment to entry in nil map`.

This patch fixes internal logic, but I decided to keep URI parsing before actually checking for nil to give faster feedback to user if uri is e.g. malformed. Feel free to suggest change in such a flow.

Method `Add` is called from multiple commands, so I changed it inside the method, not pre-checking the calls.

Other `Signer` fields seem to be safe (ranging over `nil` map is ok).

There is no test file for signer, ~so I did not create a new test case~  I added one


hdonnay (changes requested):
Ah, good catch.

I'd prefer if the conditional were the other way, to keep it consistent with the other checks in that function. With that change, LGTM

glebsts:
@hdonnay Hello, could you please clarify about "conditionals being other way"?
Do you mean smth like 
```
if s.use == nil {
  return errors.New("authority map not initialized, perhaps missing auth section in config")
}
```
?

internal/httputil/signer_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func TestNewSigner(t *testing.T) {
+	ctx := zlog.Test(context.Background(), t)
+	cfg := config.Config{}
+	signer, err := NewSigner(ctx, &cfg, jwt.Claims{})
+	if err != nil {
+		t.Error("signer initialization with empty config should succeed")
+	}
+	if signer.use != nil {
+		t.Error("signed request authority map should be non-initialized")
```
hdonnay, line 19:
```suggestion
		t.Error("signed request authority map should be initialized")
```

hdonnay (changes requested):
Sorry for letting this languish.

I think the test's error message is backwards.

With that changed, this just needs a rebase and sign-off (`git commit -s`).
<details>
<summary>rebase steps</summary>

Just adding a note here in case you're unfamiliar with a rebase workflow.
Conceptually, you're just taking your commits and applying the diff to the current `main` instead of where they were originally committed.
Assuming `origin` is the `quay/claircore` repo, and `fork` is your fork:

1. `git checkout BRANCH_NAME`
1. `git fetch origin`
1. `git rebase origin/main`
1. `git push --force-with-lease fork`

If you get a merge conflict in the `rebase` step, just replace the conflict block with the code you think should be there. If something's wrong, we'll catch it. If it's for an autogenerated file like `go.sum`, just recreate it.

If we asked you to change a commit message, add the `-i` flag to the `rebase` command and use the `reword` command.
</details>

PR #2015, opened 2024-03-22
clair: orderly shutdown

### LINKED ISSUE #1946: Graceful shutdown ... isn't

hdonnay:
Closes #1946


hdonnay:
~~Needs CI updates; working on those.~~ Done

hdonnay:
Not sure if there's a good way to cook up a test for this?

hdonnay:
ping @crozzy

introspection/server.go
```diff
@@ -36,7 +36,7 @@ const (
 // exposing Clair metrics and traces
 type Server struct {
 	// configuration provided when starting Clair
-	conf config.Config
+	conf *config.Config
```
crozzy, line 39: What is the reason for this change? Just saving the allocation when passing?

cmd/clair/main.go
```diff
@@ -98,88 +99,105 @@ func main() {
...
+			return srv.Shutdown(ctx)
+		})
+		return eg.Wait()
+	}
+}
+
+func serveIntrospection(ctx context.Context, cfg *config.Config) func() error {
+	return func() error {
+		zlog.Info(ctx).Msg("launching introspection server")
+		srv, err := introspection.New(ctx, cfg, nil)
```
crozzy, line 178: What is the reason for having the healthcheck arg when we only pass nil, just placeholding for future?

crozzy (commented):
Looks good, just a couple of questions

introspection/server.go
hdonnay: Yeah, I noticed keeping the same signature would mean either two copies (`clair.serveAPI` not being passed a pointer, then `introspection.New` not getting a pointer) or using the explicit dereference syntax for no apparent gain.

cmd/clair/main.go
hdonnay: Yeah, it's a placeholder for some real healthcheck scheme (#1974, perhaps?).

crozzy (approved):

PR #2016, opened 2024-03-22
cicd: update go version

hdonnay:
This rebuilds some of the testing workflow based on the `claircore` workflows (consolidating these is an ongoing process 🫠) and then updates the go version for modules in this repo.

Along the way, I added another lint to check version skew on the `Dockerfile` and `docker-compose.yaml`.

### DIFF
--- a/.github/workflows/config-ci.yml
+++ b/.github/workflows/config-ci.yml
@@ -16,52 +16,50 @@ on:
       - release-4.*
 
 jobs:
-  config:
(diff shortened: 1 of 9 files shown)


hdonnay:
Also switch the versions (and thus names) to `stable` and `oldstable`, which will automatically track the go upstream's supported versions.

crozzy (approved):
LGTM not sure why the tests are stalled

PR #2018, opened 2024-03-25
cicd: don't upload workspace on failure

hdonnay:
This was largely unused, and the checkout action embeds short-lived tokens in the workspace which end up in the artifact.


crozzy (approved):
LGTM

PR #2021, opened 2024-04-05
json2: experiment with `go-json-experiment`

hdonnay:


hdonnay:
Added a benchmark to the simple usage added in `notifier/webhook`, which seems promising:

```
goos: linux
goarch: amd64
pkg: github.com/quay/clair/v4/notifier/webhook
cpu: 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
BenchmarkEncodingJSON-16                  896776              1246 ns/op             657 B/op         11 allocs/op
BenchmarkCodecJSON-16                    1399909               850.9 ns/op           656 B/op         11 allocs/op
BenchmarkExperimentalJSON-16             2512387               470.7 ns/op            80 B/op          3 allocs/op
PASS
ok      github.com/quay/clair/v4/notifier/webhook       4.866s
```

PR #2028, opened 2024-04-09
chore: update production manifest with new tmp dir

crozzy:
Since https://github.com/quay/clair/pull/2019 clair will now use /var/tmp instead of /tmp by default to store the manifest layers. This PR updates where the PVC is mounted in order to concide with the new default.

### DIFF
--- a/contrib/openshift/manifests/manifests.yaml
+++ b/contrib/openshift/manifests/manifests.yaml
@@ -56,7 +56,7 @@ objects:
               command:
                 - sh
                 - '-c'
-                - rm -rf /tmp/sha* /tmp/fetch.* /tmp/rpmscanner.* && exec clair
+                - rm -rf /var/tmp/sha* /var/tmp/fetch.* /var/tmp/rpmscanner.* && exec clair
               env:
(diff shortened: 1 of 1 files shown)


contrib/openshift/manifests/manifests.yaml
```diff
@@ -56,7 +56,7 @@ objects:
               command:
                 - sh
                 - '-c'
-                - rm -rf /tmp/sha* /tmp/fetch.* /tmp/rpmscanner.* && exec clair
+                - rm -rf /var/tmp/sha* /var/tmp/fetch.* /var/tmp/rpmscanner.* && exec clair
```
hdonnay: The `rm`s shouldn't be needed, now.

Is there some way to double-check that?

hdonnay (changes requested):

contrib/openshift/manifests/manifests.yaml
crozzy: We can remove them and check the PVC metrics once it's deployed

hdonnay (approved):
LGTM w/rebase

PR #2030, opened 2024-04-09
httptransport: fix test flake

hdonnay:
Saw this test flake in CI and managed to reproduce locally. I haven't been able to reproduce with this patch, but don't have a great explanation as to why it's not consistent.

### DIFF
--- a/httptransport/instrumentation_test.go
+++ b/httptransport/instrumentation_test.go
@@ -45,19 +45,26 @@ clair_http_test_request_total{code="500",handler="err",method="get"} 1
 
 	srv := httptest.NewUnstartedServer(m)
 	srv.Config.BaseContext = func(_ net.Listener) context.Context { return ctx }
-	srv.Start()
-	defer srv.Close()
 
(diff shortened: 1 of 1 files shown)


httptransport/instrumentation_test.go
```diff
@@ -45,19 +45,26 @@ clair_http_test_request_total{code="500",handler="err",method="get"} 1
...
-	defer srv.Close()
 
-	c := srv.Client()
-	for _, p := range []string{"ok", "err", "err?panic=1"} {
-		u := srv.URL + "/" + p
-		t.Logf("making request: %q", u)
-		res, err := c.Get(u)
-		if err != nil {
-			t.Error(err)
+	// Create a scope for do the actual requests.
```
crozzy: nit: comment
```suggestion
	// Create a scope for doing the actual requests.
```

crozzy (commented):
Looks good, one nit

crozzy (approved):

PR #2031, opened 2024-04-10
clairctl: warn when range requests are not honored

### LINKED ISSUE #2029: clairctl: warn when server is not honoring range requests
> Do you think it would be possible to warn the user in case the range requests is not supported?

_Originally posted by @freedge in https://github.com/quay/clair/issues/2022#issuecomment-2041919759_

freedge:
As discussed in https://github.com/quay/clair/issues/2029 the manifest generation is efficient only when HTTP "Range: bytes=0-0" header is well supported by the server. If not, the server might return the full layer, making the manifest generation very expensive.

We warn the user in case their HTTP server does not support HTTP range requests.

### DIFF
--- a/cmd/clairctl/manifest.go
+++ b/cmd/clairctl/manifest.go
@@ -141,12 +141,22 @@ func Inspect(ctx context.Context, r string) (*claircore.Manifest, error) {
 		if err != nil {
 			return nil, err
 		}
+		// The request is needed to follow any redirection chain that the server sends to a client,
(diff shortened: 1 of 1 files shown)


hdonnay:
LGTM, will rebase and merge.

hdonnay (approved):
LGTM

/fast-forward

hdonnay:
Weird that there's a permissions issue on the approval comment fast-forward...

/fast-forward

PR #2038, opened 2024-04-16
contrib: update build script to use podman

crozzy:
There is an issue in the app-sre runners that mean skopeo is no longer working due to the docker version being update and requiring a minimum client version that the skopeo version in the runner doesn't have.

### DIFF
--- a/contrib/openshift/build_and_deploy.sh
+++ b/contrib/openshift/build_and_deploy.sh
@@ -3,17 +3,15 @@ set -exuo pipefail
 
 : "${QUAY_USER:?Missing QUAY_USER variable.}"
 : "${QUAY_TOKEN:?Missing QUAY_TOKEN variable.}"
-: "${REPOSITORY:=quay.io/app-sre}"
+: "${REGISTRY:=quay.io}"
+: "${REPOSITORY:=${REGISTRY}/app-sre}"
 : "${IMAGE:=${REPOSITORY}/clair}"
 GIT_HASH="$(git rev-parse --short=7 HEAD)"
 
 git archive HEAD |
-	docker build -t clair-service:latest -
+	podman build -t clair-service:latest -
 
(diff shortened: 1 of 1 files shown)


contrib/openshift/build_and_deploy.sh
```diff
@@ -3,17 +3,15 @@ set -exuo pipefail
...
 : "${QUAY_USER:?Missing QUAY_USER variable.}"
 : "${QUAY_TOKEN:?Missing QUAY_TOKEN variable.}"
-: "${REPOSITORY:=quay.io/app-sre}"
+: "${REGISTRY:=quay.io}"
+: "${REPOSITORY:=${REGISTRY}/app-sre}"
 : "${IMAGE:=${REPOSITORY}/clair}"
 GIT_HASH="$(git rev-parse --short=7 HEAD)"
 
 git archive HEAD |
 	docker build -t clair-service:latest -
```
hdonnay: Is there a good reason to keep using docker for the build instead of podman? I think when the first version of this script was written podman wasn't really usable in the runner.

contrib/openshift/build_and_deploy.sh
```diff
@@ -3,17 +3,15 @@ set -exuo pipefail
...
 
-skopeo copy --dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
-    "docker-daemon:clair-service:latest" \
-    "docker://${IMAGE}:latest"
+podman login -u="${QUAY_USER}" -p="${QUAY_TOKEN}" "$(REGISTRY)"
```
hdonnay: This seems wrong, it will try to execute a command named `REGISTRY`.

hdonnay (changes requested):

contrib/openshift/build_and_deploy.sh
crozzy: That is a good point, I'll update that

hdonnay (approved):

PR #2039, opened 2024-04-17
contrib: account for different container engine clients

crozzy:
Apparently the runner might not support podman, this change allows for either. Unfortunately, with docker you need to pre-tag images and can't do it all with one command like podman.


crozzy:
The classic contrib 2 step

crozzy:
I thought about adding a make command, but decided it's weird to have something in an upstream makefile that is only ever executed by internal process.

hdonnay:
yeah, adding a `Makefile` would be weird

hdonnay (approved):
seems fine

Extremely dumb we have to do it this way!

PR #2040, opened 2024-04-18
cicd: fix nightly build

hdonnay:
This needed some tweaks after the go version change.


hdonnay:
See https://github.com/quay/clair/actions/runs/8739907508 and https://github.com/quay/clair/actions/runs/8741486409

crozzy (approved):

PR #2054, opened 2024-05-01
chore: Add merge step when creating release binaries

crozzy:
The default behaviour of the upload-artifact action changed so that now any subsequent pushes to the same named artifact will result in an error as opposed to merging all results. This change adds an explicit merge step to corral all the release binaries together in an artifact.


hdonnay (changes requested):
Needs to get retrieved and used in the `release` job

hdonnay (approved):

PR #2058, opened 2024-05-01
chore: use the merge-multiple directive when downloading binaries

crozzy:
From the docs: When multiple artifacts are matched, this changes the behavior of the destination directories. If true, the downloaded artifacts will be in the same directory specified by path. If false, the downloaded artifacts will be extracted into individual named directories within the specified path.

So essentially, the artifacts were being downloaded into individual dirs and hence the release failure.

https://github.com/quay/clair/actions/runs/8912316665

### DIFF
--- a/.github/workflows/cut-release.yml
+++ b/.github/workflows/cut-release.yml
@@ -250,6 +250,7 @@ jobs:
         id: download
         with:
           pattern: clairctl-*
+          merge-multiple: true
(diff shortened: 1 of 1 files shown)


hdonnay (approved):

hdonnay:
Is this still needed?

crozzy:
Yes, I believe so, had to do it manually in the last release IIRC, going to rebase

PR #2071, opened 2024-06-05
Dockerfile: add `GOTOOLCHAIN`

hdonnay:
Changes in go1.21 means the build recipe here was doing the wrong thing. This should restore the expected behavior of using the toolchain present in the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+ARG GOTOOLCHAIN=local
 ARG GO_VERSION=1.21
 FROM quay.io/projectquay/golang:${GO_VERSION} AS build
 WORKDIR /build


hdonnay:
Cc: @bcaton85

bcaton85:
I still get the same `(go: no modules specified (see 'go help mod download'))`

hdonnay:
Using what command line?

bcaton85:
this is ran in a bash shell, if i'm understanding your question correctly

hdonnay:
No, what's the command you're using to build the container?

bcaton85:
just `podman build .`

hdonnay:
Can't reproduce, works on Fedora 40

bcaton85:
Yeah i'm on mac, maybe it's a mac issue. TBF i don't think this is a big deal if it doesn't effect development

hdonnay:
This works with "normal" `podman` but not `podman-remote` due to SELinux shenanigans:
The `go: no modules specified` error is reported when there's not a module in the specified directory, but also ends up getting reported when SELinux denies any access to the `go.mod` file and its metadata.
Adding the [`relabel` flag](https://github.com/containers/common/pull/2044) would make this work uniformly under `podman` and `podman-remote`, but is not accepted (AFAICT) by Docker/Moby/BuildKit.

@crozzy This is probably good to go.

crozzy (approved):

PR #2079, opened 2024-06-25
chore: allow podman to relabel files objects

crozzy:
Currently there is an issue when attempting to bind mount certain files during the container build process. This appears to be a combination of podman and SELinux conspiring to lead to the cryptic error: 
```
go: no modules specified (see 'go help mod download')
```
Which logically leads to something going wrong mounting the go.mod / go.sum files.

https://ci.ext.devshift.net/job/quay-clair-gh-build-main/460/console

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,8 @@ FROM quay.io/projectquay/golang:${GO_VERSION} AS build
 WORKDIR /build
 RUN --mount=type=cache,target=/root/.cache/go-build \
 	--mount=type=cache,target=/go/pkg/mod \
-	--mount=type=bind,source=go.mod,target=go.mod \
-	--mount=type=bind,source=go.sum,target=go.sum \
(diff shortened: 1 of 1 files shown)


crozzy:
This problem appeared after switching the builder (back) to podman from docker

hdonnay:
Does this work for docker? The `z` option isn't documented there.

I really hate this `docker`/`podman` incompatibility. Maybe we should only support using `buildkit`, so we can control the frontend that's used.

crozzy:
No, doesn't work for docker it turns out, damn.

Looking at the `buildkit` docs (https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypebind) I don't see an option for this either (which makes sense I suppose given they aren't supported by docker). Maybe it's worth removing those bind mounts and COPYing the go.{mod,sum}?

hdonnay:
superseded by #2083

PR #2081, opened 2024-06-26
change clairctl default logout

MK-BK:
When using k8s remotecommand to execute the command clairctl import-updaters in the Clair container, data cannot be read from stdout. The default output of the source code for clairctl is os.Stderr. Should it be changed to os.Stdout?


hdonnay:
No, the logs go to stderr and requested output goes to stdout.
Without doing this, it's impossible to redirect output while not having it mangled with logs.

PR #2082, opened 2024-06-28
chore: openshift ci

hdonnay:
This adds "dry run" to the `build_and_deploy.sh` script, then makes the `pr_check.sh` script use it.


hdonnay:
Melding into a bigger buildsystem change.

PR #2083, opened 2024-06-28
cicd: openshift CI

hdonnay:
This retools the Makefile and Dockerfile to use a buildkit backend directly in an attempt to stop needing to write a Dockerfile and scripts that only use the intersection of `podman` and `docker` features.

TODO:
- [x] Documentation updates
- [x] ~~Port the GitHub Actions CI to use the Makefile where appropriate.~~ (Will do in additional PRs)


hdonnay:
Currently failing due to extremely annoying AppSRE footguns, will pick up later.

hdonnay:
cc @Marcusk19 @bcaton85

There might be some Quay build concerns, so FYI.

etc/config.mk
```diff
@@ -0,0 +1,74 @@
...
+# The package used (via `go run`) to format go files.
+goimports ?= golang.org/x/tools/cmd/goimports@latest
+
+# These are arguments added to `go test` invocations.
+testargs =
+
+# VERSION is the version used when guessing at the current version to be used
+# in dist artifacts.
+#
+# This should be able to be interpreted accorting to gitrevisions(7) and a
```
crozzy: nit: spelling

etc/config.mk
```diff
@@ -0,0 +1,74 @@
+# `docker` and `docker-compose` control the commands invoked for a container
+# engine and a "compose file" handler, respectively.
+docker ?= $(notdir $(shell command -v podman 2>/dev/null || command -v docker 2>/dev/null))
+ifndef docker
+$(error 'docker' must not be defined as empty)
```
crozzy: Do these errors get surfaced to the user? If so should we mention that 'podman' will suffice too?

etc/container.mk
```diff
@@ -0,0 +1,81 @@
...
+ifdef DEBUG
+buildctl += --progress plain
+endif
+buildctl += --opt build-arg:BUILDKIT_MULTI_PLATFORM=1
+buildctl += $(strip $(foreach v,$(buildkit_passthru),$(if $($v),--opt build-arg:$v=$($v),)))
+buildctl += --local context=$$src
+buildctl += --local dockerfile=$$src
+buildctl += --output $(call splice,$(output_args))
+ifdef GITHUB_ACTIONS
+buildctl += --export-cache type=gha,mode=max
```
crozzy, line 17: huh, special mode for gh actions, nice

crozzy (changes requested):

etc/config.mk
hdonnay: They do, I'll make that change

hdonnay:
I guess ignore the devshift failure, despite it being the motivator of this whole exercise?

hdonnay:
[test]

crozzy (dismissed):
SGTM

crozzy (approved):

PR #2093, opened 2024-07-12
auto: improve log messages

hdonnay:
Adds some helpers and improves logging around missing cgroup interface files.

See-also: #2092

### DIFF
--- a/initialize/auto/cpu_linux.go
+++ b/initialize/auto/cpu_linux.go
@@ -18,9 +18,7 @@ import (
 // the current process's cgroup.
 func CPU() {
 	if os.Getenv("GOMAXPROCS") != "" {
-		msgs = append(msgs, func(ctx context.Context) {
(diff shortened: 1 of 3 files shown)


initialize/auto/cpu_linux.go
```diff
@@ -43,9 +41,16 @@ func CPU() {
 }
 
 func cgLookup(r fs.FS) (int, error) {
+	usingDefault := func() { infoLog("no CPU quota set, using default") }
```
crozzy: Nit: What is the idea for not having these in-line? For me the pros of one string are outweighed by the cons of more indirection.

crozzy (dismissed):
SGTM, one comment, feel free to merge as is

initialize/auto/cpu_linux.go
hdonnay: yeah, I'll just rewrite this to use a constant string :+1:

crozzy (approved):
LGTM

PR #2127, opened 2024-08-30
admin: add pre v4.8.0 admin command to delete OVAL vulns

crozzy:
The migration to delete OVAL vulnerabilities can take a while (10 mins plus depending on resources). This patch adds a command to pre-delete the vulnerabilities after adding the migration version entry.

### DIFF
--- a/cmd/clairctl/admin.go
+++ b/cmd/clairctl/admin.go
@@ -47,6 +47,13 @@ var AdminCmd = &cli.Command{
 					Usage:  "create `idx_manifest_layer_layer_id` index in the `indexer` database",
 					Action: adminPre473,
 				},
+				{
+					Name:        "v4.8.0",
+					Aliases:     []string{"4.8.0"},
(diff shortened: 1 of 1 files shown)


hdonnay:
I think this should be a pre job to make an entry in the migrations table, and a post job to do the deletions.

As is, this leaves a (larger) period of time where there will be incorrect results for RHEL.

crozzy:
> I think this should be a pre job to make an entry in the migrations table, and a post job to do the deletions.
> 
> As is, this leaves a (larger) period of time where there will be incorrect results for RHEL.

We could do it like that, the worry I'd have is that there would be a period where we'll be pulling a bunch of vulnerabilities back as the query will be looser and we'll be pulling the OVAL and VEX vulns. We could try and deploy everything together ( admin task and version switch) to reduce that period.

hdonnay:
I'd rather pull "extra" results during the manual migration window than be guaranteed to produce false negatives.

crozzy:
Split into 2

hdonnay (approved):

PR #2143, opened 2024-09-23
chore: update references to rhel updater to rhel-vex updater

crozzy:
Claircore no longer has the 'rhel' updater and in its place sits the rhel-vex updater, this change updates clair to reference the new updater.


RTann (commented):
seems like it's just a documentation change and local dev change, so LGTM

RTann (approved):

PR #2147, opened 2024-09-30
services: add rhcc enricher

crozzy:
Claircore has a new enricher that identifies RHCC layers. This patch enables it during service initialization.


crozzy:
Awaiting the Claircore side https://github.com/quay/claircore/pull/1057

PR #2154, opened 2024-10-14
build: fixing references in Dockerfile (PROJQUAY-8087)

bcaton85:
Variables within the heredoc need to be escaped in order to be assigned correctly.

Signed-off-by: bcaton <bcaton@redhat.com>


hdonnay (changes requested):
This changes the Dockerfile to execute a bash command (with broken escaping I think) instead of a script. That is, it executes `sh -c 'bash <<. ...'` instead of `sh -c '...'`. In the latter, the heredoc is Dockerfile syntax whereas in the former it's shell syntax.

I think any changes should only be made if there's a problem running this via buildkit, which is the only upstream-supported container builder.

PR #2178, opened 2024-11-25
clairctl: command to add the relevant config options for disconnected

crozzy:
Adding a command that can create a new drop-in config with the relevant disconnected settings in-place. You can also specify where the red hat indexing files are located.


hdonnay:
I don't think this is actually very useful: it's very Red Hat-specific and can be done just as easily by using a drop-in.

PR #2190, opened 2025-01-10
ci: nightly flakes

hdonnay:
This is the error that trips up slow, weird architectures in our testing. It shouldn't affect anything's correctness (it's harness code) to catch it and not error out.

### DIFF
--- a/notifier/amqp/doc.go
+++ b/notifier/amqp/doc.go
@@ -1,0 +1,5 @@
+// Package amqp implements a [Deliverer] over the AMQP protocol.
+//
+// Deprecated: This package will be removed in a future version. Users should
+// write a webhook to AMQP transducer.
+package amqp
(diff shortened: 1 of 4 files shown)


notifier/stomp/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package stomp implements a [Devliverer] over the STOMP protocol.
```
crozzy: nit: Deliverer

notifier/amqp/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package amqp implements a [Devliverer] over the AMQP protocol.
```
crozzy: nit: Deliverer

crozzy (changes requested):

crozzy (approved):

PR #2191, opened 2025-01-10
introspection: otlp integration

hdonnay:
This has been languishing for a while, but we should land it.

Our current `jaeger` tracing support has been unsupported by upstream for some time, and this will let us transition to OTLP.


crozzy (approved):
LGTM, I think I already reviewed this before IIRC, I guess I never actually approved, my bad

PR #2226, opened 2025-03-19
ci: Allow go test to access un-vendored dependencies

crozzy:
There is a situation where claircore has a dependency but because the code using the dependency isn't accessible in clair, it is not added to the go.mod/go.sum files. As this CI job not only tests clair it also tests claircore (and others), these missing dependencies cause the tests to fail. This flag allows go test to modify the go.sum file to add any missing dependencies.


crozzy:
https://github.com/quay/clair/actions/runs/13950298345

hdonnay (approved):

PR #2239, opened 2025-04-25
notifier: increase default durations to be more reasonable

crozzy:
This change increases the notifier.poll_interval and the notifier.delivery_interval default durations to be less aggressive. It doesn't make sense for them to so low given the default matcher.period is 6 hours and nothing can change in the vulnerability data within that period.


hdonnay (approved):
LGTM.

This will need a new release of the `github.com/quay/clair/config` module, then that version pulled into clair.

crozzy:
> LGTM.
> 
> This will need a new release of the `github.com/quay/clair/config` module, then that version pulled into clair.

Ahh right, thanks for the reminder

PR #2306, opened 2025-09-25
config: add ability to disable enrichment

crozzy:
This option should be used to disable claircore attempting to enrich vulnerability results.


hdonnay (approved):

hdonnay:
There's no real release process for the config module, just tag it correctly and it should be fine. I think the tag should be `config/v1.5.0` to release v1.5.0, what do you think?

crozzy:
> There's no real release process for the config module, just tag it correctly and it should be fine. I think the tag should be `config/v1.5.0` to release v1.5.0, what do you think?

Yep I _think_ I've done it couple of times before, just doing some final testing before merging

PR #2321, opened 2025-10-31
PROJQUAY-9434 updated go version to 1.25.3 for CVE-2025-47907

rhdmalone:


rhdmalone:
@hdonnay Since we're removing oldstable Go version support in this PR (due to CVE-2025-47907 requiring Go 1.25.3), could you please update the branch protection rules to remove `Tests / Tests (oldstable, amd64)` from the required status checks? The workflow now only tests with `stable` Go version.

PR #2322, opened 2025-11-03
PROJQUAY-9434: updated go version to 1.24.9 for CVE-2025-47907

rhdmalone:


hdonnay (changes requested):
In the commit message, please use a descriptive prefix like `deps:` or `go.mod:` instead of `PROJQUAY-9434:`.

The change in `config/go.mod` is unneeded and should be removed.

PR #2325, opened 2025-11-05
chore(manifests): add anti-affinity rules

cubismod:
Resolve DVO alerts by using anti-affinity rules to try to schedule clair replicas on different nodes.
The scheduler will try to schedule replicas on different nodes but schedule replicas on the same one if there aren't enough nodes available. 

[PROJQUAY-9586](https://issues.redhat.com/browse/PROJQUAY-9586)


crozzy (approved):
LGTM

PR #2329, opened 2025-11-12
clair: enable h2c

hdonnay:
Go1.24 added stdlib support for `h2c` without upgrade, so this is "just" enabling that.

### DIFF
--- a/cmd/clair/main.go
+++ b/cmd/clair/main.go
@@ -115,6 +115,12 @@ func serveAPI(ctx context.Context, cfg *config.Config) func() error {
 			BaseContext: func(_ net.Listener) context.Context {
 				return context.WithoutCancel(ctx)
 			},
+			Protocols: func() *http.Protocols {
+				var p http.Protocols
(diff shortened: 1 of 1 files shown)


BradLugo (commented):
What's the benefit here? Do you know of any clients that speak http/2? Should we put this behind a config/flag?

hdonnay:
Load balancers like [HAProxy](https://www.haproxy.com/documentation/haproxy-configuration-tutorials/protocol-support/http/#http%2F2-over-http-(h2c)-to-the-server) can use h2c to multiplex client requests onto fewer TCP connections.

BradLugo:
That makes sense. I _feel_ like it would be worth it to at least allow folks to opt-out, but I can't think of a good reason off the top of my head why someone would

BradLugo (approved):

hdonnay:
I'm confident in just turning it on because it requires prior knowledge from the client, just by the nature of the `h2c` protocol.

PR #2331, opened 2025-11-17
chore: update go version to 1.24.9 (excluding config) (PROJQUAY-9434)

rhdmalone:
Update go version to 1.24.9 (excluding config)
Fix for CVE-2025-47907


rhdmalone:
From crozzy:
Hey! I spoke with Hank about this and I think the consensus was that limiting all users of the upstream project to 1.24.9 was not ideal, when the various RH product that use clair can specify the go version they want to build the product with (correct me if I'm wrong)

PR #2365, opened 2026-01-28
feature: update logging

hdonnay:
This patch set switches the whole module over to using `log/slog` instead of `github.com/quay/zlog`.

### DIFF
--- a/cmd/clairctl/report.go
+++ b/cmd/clairctl/report.go
@@ -7,12 +7,12 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"log/slog"
 	"os"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/quay/claircore"
-	"github.com/quay/zlog"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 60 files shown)


cmd/clairctl/report.go
```diff
@@ -196,39 +192,34 @@ func reportAction(c *cli.Context) error {
...
-						Err(err).
-						Msg("manifest error")
+					log.DebugContext(ctx, "manifest error",
+						"reason", err)
 					if keepgoing {
-						zlog.Info(ctx).
-							Err(err).
-							Msg("ignoring manifest error")
+						log.InfoContext(ctx, "ignoring manifest error",
+							"error", err)
```
crozzy: should this be `"reason"` like the rest of the code?

initialize/logging.go
```diff
@@ -3,42 +3,37 @@ package initialize
...
+// Logging configures slog according to the provided configuration.
 func Logging(ctx context.Context, cfg *config.Config) error {
-	l := zerolog.New(os.Stderr)
 	switch cfg.LogLevel {
 	case config.DebugColorLog:
-		// set logger to use ConsoleWriter for colorized output
-		l = l.Level(zerolog.DebugLevel).
-			Output(zerolog.ConsoleWriter{Out: os.Stderr})
+		opts := logging.DefaultOptions()
+		opts.ProseFormat = true
```
crozzy, line 18: Should we `opts.forceANSI` (or `opts.ForceANSI`) here if you client has explicitly requested colour?

crozzy (commented):
Just a couple of comments

initialize/logging.go
hdonnay: There's no exposed knob for forcing color, just for affecting prose/JSON. It depends on whether logs are going to a TTY or not.

crozzy: Yeah, I guess I'm asking if there should be an exposed knob (`akin to opts.ForceANSI`) so if the user chooses colourized logs and is running the application with `docker compose` they will get colour like before?

hdonnay: I just a hater and don't like colorized output, so didn't add an explicit toggle. We could add it to `zlog/v2` if desired, then change this bit when updating to that future version.

crozzy: ok, fair dos, I'll make a note to open a PR in zlog and make a case for all the people who like bright coloured things there

crozzy (approved):

PR #2417, opened 2026-03-10
chore: downgrade pgx due to goroutine leak

crozzy:
Pin pgx to v5.5.3 to avoid goroutine leak in ctxwatch.ContextWatcher introduced in v5.5.4. See https://github.com/jackc/pgx/issues/1999


crozzy:
This should be reverted when https://github.com/jackc/pgx/pull/2515 is merged and available

PR #2433, opened 2026-04-07
clair: fix profile flags

hdonnay:
Does what it says on the tin.


crozzy (dismissed):

crozzy:
Ran and works

hdonnay:
I want to shove a test in here, but can't find a good way that doesn't make it much more complicated.

crozzy (approved):

PR #2445, opened 2026-04-24
clair: print flag parse errors

BradLugo:
Previously, unrecognized or invalid flags caused a silent failure exit.
Print the error to stderr so users can see what went wrong.

### DIFF
--- a/cmd/clair/main.go
+++ b/cmd/clair/main.go
@@ -46,6 +46,7 @@ func main() {
 	case errors.Is(err, flag.ErrHelp):
 		os.Exit(0)
 	default:
+		fmt.Fprintln(os.Stderr, err)
 		os.Exit(2)
 	}
 	cpu, err := flags.SetupCPUProfile()


cmd/clair/main.go
```diff
@@ -46,7 +46,8 @@ func main() {
 	case errors.Is(err, flag.ErrHelp):
 		os.Exit(0)
 	default:
-		os.Exit(2)
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
```
hdonnay: This should be `os.Exit(2)`, which is the "normal" behavior: https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/flag/flag.go;l=1153

hdonnay (changes requested):

cmd/clair/main.go
BradLugo: TIL. Thank you!

hdonnay (approved):

PR #2448, opened 2026-04-30
Add Enrichment latency panel to sample dashboard

bainsy88:
Add enrichments latency panel in the same style as the other panels using similar mertrics.

We have found this particularly useful when debug performance issue with enrichments enabled. 

Screen grab of the row with this added 

image


crozzy:
Thanks for contributing back

crozzy (approved):

PR #2454, opened 2026-05-09
httptransport: fix negotiated error responses

immanuwell:
Small cleanup for content negotiation error path.

Before this, `pickContentType` wrote `415` itself, then callers tried to send `apiError`. Too late: headers were already committed, so `/openapi/v1` with `Accept: application/xml` got a `415` with no JSON content-type. tiny paper cut, but annoying for clients.

Repro:
- before: `go test ./httptransport -run 'TestDiscovery/Failure' -count=1 -v` fails with `got content-type: "", want: "application/json"`
- after: same test passes

Validation:
- `go test ./httptransport -run 'TestDiscovery' -count=1`
- `go test ./...`

Related: #1441, #1523-ish

### DIFF
--- a/httptransport/common.go
+++ b/httptransport/common.go
@@ -35,8 +35,8 @@ func getDigest(_ http.ResponseWriter, r *http.Request) (d claircore.Digest, err 
 // "allow" slice is used.
 //
 // If "Accept" headers are present, the first (ordered by "q" value) media type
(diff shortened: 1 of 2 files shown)


crozzy (approved):

crozzy:
Thanks @immanuwell for the contribution

PR #2464, opened 2026-05-31
fix: align update operation delete api

immanuwell:
this fixes a small matcher internal API mismatch.

docs/spec said `DELETE /matcher/api/v1/internal/update_operation/{digest}` and showed `200`, but the handler actually does `uuid.Parse` on that path segment. so a client following the docs gets `400`, which is kinda rough.

repro:
1. call `DELETE /matcher/api/v1/internal/update_operation/sha256:0000000000000000000000000000000000000000000000000000000000000000`
2. before this patch it returns `400` because the handler expects a UUID update operation id, not a digest
3. after this patch the contract is `{id}` and successful deletes return `204`

tests: `go test ./...`


crozzy (dismissed):

httptransport/matcher_v1.go
```diff
@@ -239,15 +239,15 @@ func (h *MatcherV1) updateOperationHandlerDelete(w http.ResponseWriter, r *http.
...
+		slog.WarnContext(ctx, "could not parse update operation id", "reason", err)
+		apiError(ctx, w, http.StatusBadRequest, "could not parse update operation id: %v", err)
 	}
 
 	_, err = h.srv.DeleteUpdateOperations(ctx, uuid)
 	if err != nil {
 		apiError(ctx, w, http.StatusInternalServerError, "could not get update operations: %v", err)
 	}
-	// TODO(hank) This should return HTTP 204.
+	w.WriteHeader(http.StatusNoContent)
```
crozzy, line 250: I _think_ this might cause a panic if there `err` is not nil and apiError already writes `http.StatusInternalServerError`.

